PR #121, opened 2021-10-02
Use ccache ci

blagoev:
This enables ccache on our CI build for iOS and brings down the build time from 25 min to 7min as of now. The caching is done with a key <OS>-<realm-core-commit-hash>. Whenever realm-core submodule changes this will create a new store for the ccache. The reasoning is that realm-core takes the most time to build so we should cache that build to speed up our own. 
Later on we can enable the ccache on the Android build and on the builds frn Linux as well.


nielsenko (approved):
Looks good to me.. with my limited knowledge :-)

PR #123, opened 2021-10-05
use macos 10.15 for gh actions

blagoev:
Seems like macos-latest switched to 11 and Android CI started to fail sporadically. Previously on Mac 10.15 it was running deterministically so use that one instead

### DIFF
--- a/.github/workflows/realm-flutter-ios.yml
+++ b/.github/workflows/realm-flutter-ios.yml
@@ -24,7 +24,9 @@ jobs:
           submodules: 'recursive'
 
       - name: get-ccache-id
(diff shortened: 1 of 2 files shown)


.github/workflows/realm-flutter-ios.yml
```diff
@@ -24,10 +24,10 @@ jobs:
           submodules: 'recursive'
 
       - name: get-ccache-id
-        run: echo "CCACHE_ID=${{ runner.OS }} $(git submodule status src/realm-core)" >> $GITHUB_ENV
+        run: echo "CCACHE_ID=os:${{ runner.OS }} build:ios realm-core commit hash $(git submodule status src/realm-core | cut -c 1-41)" >> $GITHUB_ENV
```
nielsenko, line 27: Should we use the target, instead of the runner os for the cache id tag?

nielsenko (approved):

PR #125, opened 2021-10-06
update ccache to 1.0.5

blagoev:


nielsenko:
I thought `@v1`would automatically pick the highest non-breaking version (in the SemVer sense), but I cannot easily find the docs right now?

nielsenko (approved):

PR #135, opened 2021-11-01
Move android flutter ci to ubuntu-20.04

nielsenko:


blagoev:
we need to use the macos host for the faster android emulators.

PR #138, opened 2021-11-10
Realm config models and open

blagoev:
added ffigen project to generate binding code
refactored configuration
refactored Realm class and added Realm.open support
added RealmPropertyType and refactored RealmProperty
added RealmCore which abstracts away the realm-core library and the ffigen code usage from the rest of the dart project
added realm_dart_scheduler.cpp - a Realm Scheduler implementation for Dart
added Configuration tests
aded Realm.open tests
initialize Dart*_DL functions on initRealm


ffigen/README.md
```diff
@@ -0,0 +1,14 @@
+This package configures the ffigen to generate bindings to realm-core library.
+The is executed manually as needed and this package is never published.
+
+Usage: 
+
+On Windows: dart run ffigen --config windows.yaml
```
nielsenko, line 6: Why do we have these 3 configs. The only change to the generated `realm-dart/ffigen/../lib/src/native/realm_bindings.dart` is a comment telling us what config was used, ie:
```
-/// A Windows config for ffigen Usage: dart run ffigen --config windows.yaml
+/// A Linux config for ffigen Usage: dart run ffigen --config linux.yaml
```

nielsenko:
I noticed that examples don't compile

ffigen/README.md
blagoev: these are host platform configurations and not target platform configs. They currently produce the same output yes.

blagoev:
@nielsenko thanks for the heads up about the examples. Currently this requires manually generated schema so examples are left untouched. The tests are supposed to work so this can pass CI, which I am currently fixing.

nielsenko:
Have made a branch with suggested changes [here](https://github.com/realm/realm-dart/tree/pr-suggestions-for-realm-config-models-and-open)

lib/src/configuration.dart
```diff
@@ -216,24 +47,40 @@ class Configuration {
...
+  int get schemaVersion => realmCore.getSchemaVersion(this);
+  set schemaVersion(int value) => realmCore.setSchemaVersion(this, value);
+
+  String get path => realmCore.getConfigPath(this);
+  set path(String value) => realmCore.setConfigPath(this, value);
+}
+
+class SchemaObject {
+  Type type;
+  String get name => type.toString();
```
nirinchev, line 59: We don't have to support it initially, but we'd eventually want to enable users to map dart types to different Realm tables. In C# this is achieved using an attribute:

```csharp
[MapTo("bar")]
class Foo : RealmObject { }
```

Since we have the `@RealmModel` annotation in dart, we can allow users to use it to override the name of the model - e.g.

```dart
@RealmModel("bar")
class _Foo { }
```

which will generate something like:

```dart
class Foo extends _Foo with RealmObject {
   @override
   static SchemaObject get schema => SchemaObject(Foo, name: "bar")..properties = [ ];
 }
```

Where `SchemaObject` is defined like:

```dart
class SchemaObject { 
  Type type;
  String name;

  SchemaObject(Type type, { String? name = null }) {
    this.type = type;
    this.name = name ?? type.toString();
  }
}
```

lib/src/configuration.dart
```diff
@@ -216,24 +47,40 @@ class Configuration {
...
 
-  /// Specifies if this Realm should be deleted if a migration is needed.
-  bool? deleteRealmIfMigrationNeeded;
+  List<SchemaProperty> properties = [];
+
+  SchemaObject(this.type);
+}
+
+class RealmSchema extends Iterable<SchemaObject> {
+  late final SchemaHandle handle;
```
nirinchev, line 67: I'm new to dart, but should this (and other such handles) be public? Sounds like users should never need to use them.

lib/src/realm_class.dart
```diff
@@ -19,231 +19,79 @@
...
+class Realm {
+  /// The [Configuration] object of this [Realm]
+  Configuration get config => throw RealmException("not implemented");
+  late RealmHandle _realm;
+  late final _Scheduler _scheduler;
+
+  /// Opens a Realm using the default or a custom [Configuration] object
+  Realm(Configuration config) {
+    if (!config.schema.isEmpty) {
+      realmCore.validateSchema(config.schema);
```
nirinchev, line 52: Why do we need this? I imagine we'll validate the schema when we try to open the Realm anyway.

lib/src/realm_class.dart
```diff
@@ -19,231 +19,79 @@
...
-  void removeListener(Event event, ListenerCallback callback) {
-    _removeListener(event.toString().split(".")[1], callback);
-  }
-  void _removeListener(String name, ListenerCallback callback) native 'Realm_removeListener';
-
-  /// Closes this [Realm] so it may be re-opened with a newer schema version. 
-  /// All objects and collections from this Realm are no longer valid after calling this method.
-  /// This method will not throw an exception if called multiple times.
-  close() native 'Realm_close';
+  static String get version => realmCore.libraryVersion;
```
nirinchev, line 59: Should this return the version of the dart library or the version of Core?

lib/src/realm_object.dart
```diff
@@ -23,133 +23,145 @@
...
-
-    Object result = _native_get(name);
-    if (result is RealmList) {
-      throw new Exception("Invalid RealmObject. RealmLists should be retrieved using super_get method");
-    }
-
-    return result;
+class RealmObject {
+  String _value = "";
+  setString(String anme, String value) {
```
nirinchev, line 39:
```suggestion
  setString(String name, String value) {
```

lib/src/configuration.dart
blagoev: Yes we could do that. We have already discussed the support for MapTo attribute. This could be applied to properties and whole classes.

lib/src/configuration.dart
blagoev: yes handles need to be public but they are kept with empty signatures (except toString()) on purpose so nobody have any useful thing to do with them.

lib/src/realm_class.dart
blagoev: yes this is something we discussed already internally . I will be removing this validation

lib/src/configuration.dart
blagoev: Appreciate the code samples, but I don't think we need to change anything now. Keep in mind one thing in Dart, unlike .NET, changing a field to a getter is not a breaking change :)

lib/src/realm_class.dart
blagoev: None of these. I will be removing it. Good to point it out

lib/src/realm_class.dart
blagoev: done

lib/src/realm_class.dart
blagoev: done

lib/src/configuration.dart
nirinchev: Why do they need to be public though? Is it an ffi limitation or something that the generator needs to use?

blagoev: In Dart private members are private to the library not the package. What issue do you see with these being public except that these are one extra field on certain Realm classes? They do not provide anything useful and couldn't be used in any manner whatsoever.

blagoev:
@nielsenko I have incorporated `late` removal where possible. For Realm class I didn't since follow-up PR will need to change back again.

nielsenko:
👍 I rebased the [pr-suggestions-for-realm-config-models-and-open ](https://github.com/realm/realm-dart/tree/pr-suggestions-for-realm-config-models-and-open) branch. Only minor nitpickings left

nielsenko (approved):

ffigen/README.md
```diff
@@ -0,0 +1,14 @@
...
+On MacOS: dart run ffigen --config macos.yaml
+On Linux: dart run ffigen --config linux.yaml
+
+Note: Dart ffigen tool generated platform dependent bindings which are tailored for the host platform the ffigen is executed. Currently all of platforms generate similar output with the configuration workaround here: https://github.com/dart-lang/ffigen/pull/119.
```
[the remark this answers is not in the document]
blagoev: it is by chance right now that mac and linux can use same config. Linux is can't. These are configurations for the host platform and not the target platform so I suggest we leave them in. If we don't need them in the future we could remove them. These are temporary anyways since we are waiting for the platform independent ffi generator

lib/main.dart
```diff
@@ -18,118 +18,51 @@
...
-  return DynamicLibrary.open(fullPath);
-}
+  print("Current PID ${pid}");
 
-void main() {
-  if (IsFlutterPlatform) {
-    final testLibrary = dlopenPlatformSpecific("realm_flutter");
+  //can read the default config
+  {
+    var config = new Configuration([Car.schema, Person.schema]);
```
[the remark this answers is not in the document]
blagoev: Yes I know. Thanks. I am already getting used to it. Personally I think this really shouldn't matter but... we have this changing language under the hood and the linter is complaining so lets follow the "guidance".

lib/src/native/realm_core.dart
```diff
@@ -0,0 +1,187 @@
...
+  _RealmCore._();
+
+  factory _RealmCore() {
+    return _instance ??= _RealmCore._();
+  }
+  //
+
+  String get libraryVersion => _realmLib.realm_get_library_version().cast<Utf8>().toDartString();
+
+  //TODO: Use Finalizers, when available, instead of native WeakHandles https://github.com/dart-lang/language/issues/1847
```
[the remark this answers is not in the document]
blagoev: Would loved this to be a code suggestion :) I will mention it sure, but imo the issue link should be enough most of the time
```suggestion
  //TODO: Use Finalizers, when available, instead of native WeakHandles https://github.com/dart-lang/language/issues/1847
  //The proposal: https://github.com/dart-lang/language/blob/master/working/1847%20-%20FinalizationRegistry/proposal.md
```

lib/main.dart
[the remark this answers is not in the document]
blagoev:
```suggestion
    var config = Configuration([Car.schema, Person.schema]);
```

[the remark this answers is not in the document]
blagoev: I will not commit this since this is already gone in another PR. to skip a CI cycle for this.

lib/src/realm_class.dart
```diff
@@ -20,230 +20,50 @@ import 'dart:async';
...
-      throw new RealmException("Could not delete the Realm at $path error: $e");
+class Realm {
+  /// The [Configuration] object of this [Realm]
+  Configuration get config => throw RealmException("not implemented");
+  late RealmHandle _realm;
+
+  /// Opens a Realm using the default or a custom [Configuration] object
+  Realm(Configuration config) {
+    if (!config.schema.isEmpty) {
+      realmCore.validateSchema(config.schema);
```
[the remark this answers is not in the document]
blagoev: this is outdated. will resolve

lib/src/configuration.dart
blagoev: resolving this one

lib/src/configuration.dart
blagoev: @nirinchev Can I resolve this comment

lib/src/realm_object.dart
```diff
@@ -25,133 +25,156 @@
...
-    return schema;
+  String toString() {
+    return "RealmException: $message";
   }
+}
 
-  Object isValid() native "RealmObject_isValid";
+  // Map<String, Object>? _unmanagedProperties;
+
+  // /**
```
nielsenko, line 65: Should we remove this out-commented code?

nielsenko (approved):
A few minor comments, nothing major

.github/workflows/realm-flutter-ios.yml
```diff
@@ -56,5 +56,5 @@ jobs:
           os_version: '>= 14.0'
 
       - name: Run tests on iOS Simulator
-        run: flutter drive --target=test_driver/app.dart --dart-define=testName="Realm version"
+        run: flutter drive --target=test_driver/app.dart --dart-define=testName=""
```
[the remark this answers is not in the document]
blagoev: this is just a reminder how to pass the filter. so next time one needs to do it its right there ready for consumption.

lib/src/realm_object.dart
blagoev: yes. lets remove that when read object is ready

.github/workflows/realm-flutter-ios.yml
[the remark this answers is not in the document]
blagoev: maybe leaving it as a trailing comment will be better

lib/src/realm_object.dart
blagoev: resolving this one. will remember to clear the code in the following PR

lib/src/configuration.dart
blagoev: resolving

blagoev: @nirinchev we have moved these handles to private here https://github.com/realm/realm-dart/blob/master/lib/src/configuration.dart#L32

PR #140, opened 2021-11-23
Realm add

blagoev:
implement adding objects to realm


ffigen/linux.yaml
```diff
@@ -10,12 +10,15 @@ headers:
     - 'realm.h'
     - 'realm_dart.h'
     - 'realm_dart_scheduler.h'
+    - 'realm_android_platform.h'
```
nielsenko, line 13: bit surprised to see `android` here.. and similar for macos and windows, but I assume it is intended (just checking)

lib/realm.dart
```diff
@@ -58,7 +58,7 @@ void initRealm() {
       return true;
     }());
     
-    if (!path.isEmpty && path.endsWith("/")) {
+    if (path.isNotEmpty && path.endsWith("/")) {
```
nielsenko, line 61: Consider using [path](https://pub.dev/packages/path) package for path manipulation.

lib/src/configuration.dart
```diff
@@ -27,19 +28,38 @@ import 'helpers.dart';
...
+    schemaVersion = 0;
+    path = defaultPath;
+    realmCore.setSchema(this);
   }
 
+  /// The platform dependent path to the default realm file
+  static String get defaultPath {
+    var path = "default.realm";
+    if (Platform.isAndroid || Platform.isIOS) {
+      path = "${realmCore.getFilesPath()}${Platform.pathSeparator}$path";
```
nielsenko, line 48: I suggest using [path](https://pub.dev/packages/path) here as well

lib/src/native/realm_core.dart
```diff
@@ -1,30 +1,49 @@
...
 
 final _RealmCore realmCore = _RealmCore();
 
 class _RealmCore {
-   
   //From realm.h. Currently not exported from the shared library
   // ignore: constant_identifier_names
   static const int RLM_INVALID_CLASS_KEY = 0x7FFFFFFF;
-  // ignore: constant_identifier_names
+  // ignore: constant_identifier_names, unused_field
```
nielsenko, line 44: consider using ignore_for_file to avoid repetition

lib/src/native/realm_core.dart
```diff
@@ -39,63 +58,78 @@ class _RealmCore {
...
+      final schemaProperties = arena<Pointer<realm_property_info_t>>(classCount);
 
       for (var i = 0; i < classCount; i++) {
         final schemaObject = schema.elementAt(i);
         final classInfo = schemaClasses.elementAt(i).ref;
 
-        classInfo.name = schemaObject.name.toNativeUtf8().cast();
-        classInfo.primary_key = "".toNativeUtf8().cast();
+        classInfo.name = schemaObject.name.toUtf8Ptr(arena);
+        classInfo.primary_key = "".toUtf8Ptr(arena);
```
nielsenko, line 96: I assume the empty string is intended here as a temporary, until we are further (just checking)

lib/src/native/realm_core.dart
```diff
@@ -39,63 +58,78 @@ class _RealmCore {
...
 
         for (var j = 0; j < propertiesCount; j++) {
           final schemaProperty = schemaObject.properties[j];
           final propInfo = properties.elementAt(j).ref;
-          propInfo.name = schemaProperty.name.toNativeUtf8().cast();
-          propInfo.public_name = "".toNativeUtf8().cast();
-          propInfo.link_target = "".toNativeUtf8().cast();
-          propInfo.link_origin_property_name = "".toNativeUtf8().cast();
+          propInfo.name = schemaProperty.name.toUtf8Ptr(arena);
+          propInfo.public_name = "".toUtf8Ptr(arena);
```
nielsenko, line 109: similar observation..

lib/src/native/realm_core.dart
```diff
@@ -116,26 +150,146 @@ class _RealmCore {
...
+    return RealmObjectHandle._(realmPtr);
+  }
+
+  Object readProperty(RealmObject object, int propertyKey, RealmPropertyType propertyType) {
+    return using((Arena arena) {
+      Pointer<realm_value_t> value = arena<realm_value_t>();
+      _realmLib.invokeGetBool(() => _realmLib.realm_get_value(object.handle._pointer, propertyKey, value));
+
+      switch (propertyType) {
+        case RealmPropertyType.int:
```
nielsenko, line 248: Reminds of a thing I intended to discuss.. Should we just call it `RealmType` instead of `RealmPropertyType`?

lib/src/native/realm_core.dart
```diff
@@ -164,12 +315,48 @@ class ConfigHandle extends Handle<realm_config> {
...
+  }
+}
+
+class RealmObjectHandle extends Handle<realm_object> {
+  RealmObjectHandle._(Pointer<realm_object> pointer) : super(pointer) {
+    _realmLib.realm_attach_finalizer(this, _pointer.cast(), 112);
+  }
+}
+
+extension _StringEx on String {
```
nielsenko, line 334: You can drop `_StringEx` and just do `extension on String` since it is private anyway

lib/src/native/realm_core.dart
```diff
@@ -164,12 +315,48 @@ class ConfigHandle extends Handle<realm_config> {
...
+    final units = utf8.encode(this);
+    final Pointer<Uint8> result = allocator<Uint8>(units.length + 1);
+    final Uint8List nativeString = result.asTypedList(units.length + 1);
+    nativeString.setAll(0, units);
+    nativeString[units.length] = 0;
+    return result.cast();
+  }
+}
+
+extension _RealmLibraryEx on RealmLibrary {
```
nielsenko, line 345: similar observation

lib/src/realm_object.dart
```diff
@@ -16,37 +16,107 @@
...
- */
+abstract class RealmAccessor {
+  T get<T extends Object>(RealmObject object, String name);
+  void set<T extends Object>(String name, T value);
+}
 
-import 'list.dart';
-import 'realm_property.dart';
+class RealmValuesAccessor implements RealmAccessor {
+  final Map<String, Object> _values = <String, Object>{};
```
nielsenko, line 28: Shouldn't this be `Map<String, Object?>` to allow for null values?

lib/src/realm_object.dart
```diff
@@ -16,37 +16,107 @@
...
 
-// ignore_for_file: native_function_body_in_non_sdk_code
+import 'native/realm_core.dart';
+import 'realm_property.dart';
 
-/**
- *  The file is intentionaly not follwoing the dart naming guidelines. The name is used from native code by convention
- */
+abstract class RealmAccessor {
+  T get<T extends Object>(RealmObject object, String name);
```
nielsenko, line 23: Why not just `T get<T>(`.. anything extends Object? This is just one example.

lib/src/realm_property.dart
```diff
@@ -46,7 +46,6 @@ enum RealmPropertyType {
   uuid,
 }
 
-/// A annotation class used to define Realm data model classes and their properties
 class RealmProperty {
```
nielsenko, line 49: Do we need this anymore?

ffigen/linux.yaml
blagoev: yes this is needed. this is the linux host config. In other words  if we want to run the ffigen on linux this is the config to use. It should have these headers to generate correct bindings

lib/realm.dart
blagoev: thanks this is old code when we did not depend on path package. Will try to update

lib/src/configuration.dart
blagoev: 👍

lib/src/native/realm_core.dart
blagoev: I intentionally did not do it. since it will ignore everything in the file from now on.

lib/src/native/realm_core.dart
blagoev: yes we need to update it when we do the primary key support. Perhaps I should add a TODO there, to get analyzer help

lib/src/native/realm_core.dart
blagoev: yes we could do that, I just thought at the time that this is more descriptive

lib/src/native/realm_core.dart
blagoev: 👍

lib/src/native/realm_core.dart
blagoev: 👍

lib/src/realm_object.dart
blagoev: good point will fix

lib/src/realm_property.dart
blagoev: SchemaProperty is RealmProperty. I can merge the two classes or I could change this to a private base class.

lib/realm.dart
blagoev: actually since this is split by platform I don't see the gain of using path here. path pkg is useful when dealing with paths on generic code. but here we are explicit for every single platform.

lib/src/native/realm_core.dart
blagoev: also this is how core c-api names these values. I suggest we leave it as RealmPropertyType. And then there will be RealmCollectionType

lib/src/realm_object.dart
blagoev: I did not plan on supporting null hence its written like that. But I guess I was mistaken since we need to support null values here as well.

nielsenko (approved):

PR #145, opened 2021-12-06
Default values

blagoev:
Handle default values explicitly and pass the correct argument to core when creating objects.


lib/src/realm_object.dart
```diff
@@ -21,23 +21,64 @@ import 'realm_property.dart';
 
 abstract class RealmAccessor {
   T get<T>(RealmObject object, String name);
-  void set<T>(RealmObject object, String name, T value);
+  void set<T>(RealmObject object, String name, T value, [bool isDefault = false]);
+
+  static final Map<Type, Map<String, Object?>> _defaultValues = <Type, Map<String, Object?>>{};
```
nielsenko, line 26: Why not let the factories registered on RealmObject handle default values?

blagoev: We discussed this on the round table, so resolving it

nielsenko (approved):

PR #149, opened 2021-12-10
Lists

blagoev:
Initial support for Lists


desistefanova (approved):

lib/src/list.dart
```diff
@@ -16,13 +16,77 @@
...
+  late final Realm _realm;
+
+  RealmList._(this._handle, this._realm);
+
+  @override
+  int get length => realmCore.getListSize(this);
+
+  //settng lenght is no operation
+  @override
+  set length(int length) {}
```
nielsenko, line 39: I vaguely remember that set length is used to truncate lists and that some of the methods inherited from ListBase<T> expect this to be true... So I'm not sure it is kosher to make it a no-op. 

But maybe I'm wrong. Just something to check up on

blagoev: will overwrite add instead

nielsenko (approved):

PR #153, opened 2021-12-29
realm delete many

### LINKED ISSUE #37: Support Results for..in
Results should be iterable. Consider exposing asList which returns a List<T>

### LINKED ISSUE #60: Implement deleteMany
Takes a List of RealmObjects and removes them all

desistefanova:
realm.deleteMany implemented "Fixes #60".
Support RealmResults for..in "Fixes #37" without snapshot


nielsenko (changes requested):

lib/src/realm_class.dart
```diff
@@ -87,6 +87,22 @@ class Realm {
     realmCore.removeRealmObject(object);
   }
 
+  void removeMany<T extends RealmObject>(Iterable<T> list) {
```
nielsenko, line 90: Nitpicking.. but maybe call list something else (fx. items)

lib/src/results.dart
```diff
@@ -95,7 +95,7 @@ class _ResultsList<T extends RealmObject> extends collection.ListBase<T> {
 /// Instances of this class are typically live collections returned by [Realm.objects]
 /// that will update as new objects are either added to or deleted from the Realm
 /// that match the underlying query.
-class RealmResults<T extends RealmObject> {
+class RealmResults<T extends RealmObject> extends collection.IterableBase<T> {
```
nielsenko, line 98: Great for now. We might choose to implement a fatter interface than Iterable<T> later

nielsenko (approved):

lib/src/native/realm_core.dart
```diff
@@ -384,6 +384,14 @@ class _RealmCore {
...
   }
 
+  void realmListRemoveAll(RealmList list) {
+    _realmLib.invokeGetBool(
+        () => _realmLib.realm_list_remove_all(list.handle._pointer));
+  }
+
+  void realmResultsDeleteAll(RealmResults results) {
+    _realmLib.invokeGetBool(() => _realmLib.realm_results_delete_all(results.handle._pointer));
+  }
```
nirinchev, line 394: Are these doing the same thing? I.e. deleting the object from the Realm? If so, should we name them consistently?

lib/src/realm_class.dart
```diff
@@ -87,6 +87,22 @@ class Realm {
...
+    } else if (list is RealmList<T>) {
+      realmCore.realmListRemoveAll(list);
+    } else {
+      for (T realmObject in list) {
+        realmCore.removeRealmObject(realmObject);
+      }
+      }
+    } catch (e) {
+      throw RealmException("Error deleting objects from databse. Error: $e");
+    }
```
nirinchev, line 103: Indentation seems off here.

lib/src/results.dart
```diff
@@ -207,3 +210,34 @@ extension RealmResultsInternal on RealmResults {
     return RealmResults<T>._(handle, realm);
   }
 }
+
+class RealmResultsIterator<T extends RealmObject> implements Iterator<T> {
+  final Iterable<T> _iterable;
+  final int _length;
+  int _index;
+  T? _current;
```
nirinchev, line 218: [q] What happens if `T` is already nullable? I.e. if I have a `RealmResults<Person?>`, will this work?

nirinchev (commented):
Mostly questions and nits from me - I was just curious about the PR.

lib/src/results.dart
```diff
@@ -207,3 +210,34 @@ extension RealmResultsInternal on RealmResults {
...
+  bool moveNext() {
+    int length = _iterable.length;
+    if (_length != length) {
+      throw ConcurrentModificationError(_iterable);
+    }
+    if (_index >= length) {
+      _current = null;
+      return false;
+    }
+    _current = _iterable.elementAt(_index);
```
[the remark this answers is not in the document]
desistefanova: I see your point. Yes, it could be changed.

test/realm_test.dart
```diff
@@ -552,5 +552,135 @@ Future<void> main([List<String>? args]) async {
...
+      final teamTwo = Team()..name = "Maserati";
+      realm.write(() => {realm.add(teamOne), realm.add(teamTwo)});
+
+      //Ensule teams exist in DB
+      var teams = realm.all<Team>();
+      expect(teams.length, 2);
+
+      //Remove all objects in RealmResults from DB
+      realm.write(() => realm.removeMany(teams));
+      teams = realm.all<Team>();
```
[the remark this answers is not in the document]
desistefanova: Since it is a test I wanted to read them all again from DB. Just wanted to be sure that they are really deleted from DB, not only from the collection. Is there better way to do it? Do you think it is useless?

lib/src/native/realm_core.dart
desistefanova: Yes, you are right. I named them similar to the API, but we will discuss these names, maybe they have to be consistent.

lib/src/results.dart
desistefanova: In Dart nullable types are a little bit different than .Net. Since generic type extends RealmObject (not RealmObject?), it is not possible to pass nullable type. So I think it is OK to be like that.

lib/src/realm_class.dart
```diff
@@ -87,6 +87,16 @@ class Realm {
     realmCore.removeRealmObject(object);
   }
 
+  void removeMany<T extends RealmObject>(List<T> list) {
```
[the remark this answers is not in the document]
desistefanova: Done

test/realm_test.dart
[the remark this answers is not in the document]
nirinchev: I think there's value in both tests - i.e. ensuring that the collection has automatically updated as well as that the value in the database has been updated. Something like:

```dart
       //Remove all objects in RealmResults from DB
       realm.write(() => realm.removeMany(teams));
       
       // Ensure teams is now empty
       expect(teams.length, 0);

       // Just in case, try to re-read the table from the database and confirm its empty
       teams = realm.all<Team>();
       expect(teams.length, 0);
```

lib/src/results.dart
nirinchev: Ugh... I think this will be a problem down the line as `RealmResults` should be able to contain `null` objects. This is not the case for regular queries like `realm.all<Foo>`, but when you add support for dictionaries, the keys and the values are represented as `RealmResults` and the values collection can contain `null` objects. But it doesn't have to be fixed in this PR - it's more of a deeper reaching design change for `RealmResults` in general.

lib/src/results.dart
```diff
@@ -142,16 +142,19 @@ class RealmResults<T extends RealmObject> {
...
 
   /// Returns `true` if the Results collection is empty
-  bool isEmpty() {
-    return length == 0;
-  }
+  @override
+  bool get isEmpty => length == 0;
+
+  @override
+  Iterator<T> get iterator => RealmResultsIterator(this);
```
nielsenko, line 149: Do we need a full class here? Could we not just use a `sync*` getter here?

nielsenko: Scratch that .. your are implementing IterableBase.. my mistake

lib/src/results.dart
```diff
@@ -207,3 +210,30 @@ extension RealmResultsInternal on RealmResults {
...
+  final int _length;
+  int _index;
+
+  RealmResultsIterator(Iterable<T> iterable)
+      : _iterable = iterable,
+        _length = iterable.length,
+        _index = 0;
+
+  @override
+  T get current => (_index >= 0 && _index < _length) ? _iterable.elementAt(_index) : null as T;
```
nielsenko, line 225: `null as T` will throw as T is not nullable. Maybe better to throw a `RangeError` in that case

nielsenko: Shouldn't happen though.. so why have the check at all?

lib/src/results.dart
```diff
@@ -207,3 +210,30 @@ extension RealmResultsInternal on RealmResults {
     return RealmResults<T>._(handle, realm);
   }
 }
+
+class RealmResultsIterator<T extends RealmObject> implements Iterator<T> {
+  final Iterable<T> _iterable;
+  final int _length;
+  int _index;
+
+  RealmResultsIterator(Iterable<T> iterable)
```
nielsenko, line 219: This ctor should take a `RealmResults<T>` - it will not work for an arbitrary `Iterable<T>`. Also, consider making it a private class - users should not be instantiating these by them selves.

bmunkholm:
Please link to the issues this resolves with "Fixes #xyz".

desistefanova:
> Please link to the issues this resolves with "Fixes #xyz".

OK, thanks. Done.

lib/src/results.dart
desistefanova: Right, fixed in https://github.com/realm/realm-dart/pull/153/commits/86e68144180eff8d4aa10ca61940ee4dc50c4b8b

lib/src/results.dart
desistefanova: Fixed in https://github.com/realm/realm-dart/pull/153/commits/86e68144180eff8d4aa10ca61940ee4dc50c4b8b
and https://github.com/realm/realm-dart/pull/153/commits/9c67ce3f8dd5edd209001a2f6f07e784962bc593

lib/src/results.dart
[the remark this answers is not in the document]
nielsenko: Come to think of it, I prefer the eager evaluation. It ensures that `calling` current will never throw, despite changes to underlying collection. Instead only `moveNext` will throw. The optimisation for the case where only some of the iterated items are actually accessed is less important in my opinion, so I would keep the `_current` field.

lib/src/results.dart
```diff
@@ -207,3 +210,33 @@ extension RealmResultsInternal on RealmResults {
...
+        _length = results.length,
+        _index = 0;
+
+  @override
+  T get current => _current as T;
+
+  @override
+  bool moveNext() {
+    int length = _results.length;
+    if (_length != length) {
```
nielsenko, line 231: I would consider dropping `_length field` and just check against `_index`, ie. `_index >= length`. This would allow more modifications to the result set, even if the produced sequence will be odd. The current check doesn't catch all modifications anyway, but maybe check some of the other SDKs?

lib/src/results.dart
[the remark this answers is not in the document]
desistefanova: Yes, I already returned _current back, also because it has to be evaluated before the index is incremented.

lib/src/results.dart
```diff
@@ -207,3 +210,33 @@ extension RealmResultsInternal on RealmResults {
...
+  int _index;
+  T? _current;
+
+  _RealmResultsIterator(RealmResults<T> results)
+      : _results = results,
+        _length = results.length,
+        _index = 0;
+
+  @override
+  T get current => _current as T;
```
nielsenko, line 226: How do you access the first element? Calling `moveNext` when `_index == 0` will set `_current = _results[1]`. Prior to that `_current == 0`.

nielsenko: Scratch that :-) .. 

> bool moveNext()
> Advances the iterator to the next element of the iteration.
> 
> Should be called before reading current. If the call to moveNext returns true, then current will contain the next element of the iteration until moveNext is called again. If the call returns false, there are no further elements and current should not be used any more.
> 
> It is safe to call moveNext after it has already returned false, but it must keep returning false and not have any other effect.
> 
> A call to moveNext may throw for various reasons, including a concurrent change to an underlying collection. If that happens, the iterator may be in an inconsistent state, and any further behavior of the iterator is unspecified, including the effect of reading current.

desistefanova: On the first call of moveNext index == 0 and  _current  _results[index]. After that index is incremented. It seems to be correct according to me.

lib/src/results.dart
nielsenko: Ie. somthing like:
```
  @override
  bool moveNext() {
    int length = _results.length;
    if (_index == length) return false;
    if (_index > length) {
      throw ConcurrentModificationError(_results);
    }
    _current = _results[_index];
    _index++;
    return true;
  }

```

desistefanova: I think it is better to have some notification when something has changed in the result concurrently while it is iterated. Otherwise we could break some of the logic in the apps. Let's discuss it and decide how to do it.

nielsenko: Could even consider just returning `false` on `_index >= length`?

nielsenko: We won't see all changes anyway. If the i'th index in the result set is changed, the current check won't catch that. I think I prefer to be as robust as possible here. Sync may change a realm result, and I don't think users will like to guard all iterations against `ConcurrentModificationError`.. but I agree - it is not optimal

nielsenko: Follow up on zoom discussion.. to match .net would be:@override
```
  bool moveNext() {
    int length = _results.length;
    if (_index >= length) {
      _current = null;
      return false;
    }
    _current = _results[_index];
    _index++;
    return true;
  }
```
.. but also snapshot (if possible) the collection before iteration

nirinchev: For some background on the .NET code - while this is a departure from the BCL, we felt it's warranted because of the nature of Realm. Getting `ConcurrentModificationError` due to sync/some background thread modifying the collection while you're iterating over it is going to be super unintuitive for users and as Kasper mentions, most users will not think of adding error handling to each foreach (which is understandable considering how annoying it will be), meaning throwing here will likely result in a hard crash in production. So my recommendation would be not to throw and add snapshotting if possible.

nielsenko: I think we should go with the above, and look into snapshotting later

lib/src/results.dart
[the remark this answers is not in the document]
nirinchev: I don't have strong opinions on the subject and don't know how the dart runtime does iteration, but if it's anything like the .NET runtime, `moveNext` and `current` will always be called together, so which one throws is kind of irrelevant for the developer. The only minor performance difference is when the enumerator is used to check if the collection is empty, such as `isEmpty = !results.getEnumerator().moveNext()`, but I don't suppose we need to optimize for this.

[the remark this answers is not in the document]
nielsenko: We decided to drop the throw even in `moveNext` (ie. just return false, `if _index >=` current length), so correct iteration should never throw, but it may produce an odd sequence if the underlying collection was changed during the iteration (and the realm result was not snapshotted)

lib/src/results.dart
```diff
@@ -207,3 +210,33 @@ extension RealmResultsInternal on RealmResults {
...
+    int length = _results.length;
+    if (_length != length) {
+      throw ConcurrentModificationError(_results);
+    }
+    if (_index >= length) {
+      _current = null;
+      return false;
+    }
+    _current = _results[_index];
+    _index++;
```
nirinchev, line 239: I don't like that `_index` and `_current` are pointing to different elements in the collection - it's not super intuitive and took me a while to convince myself that the code is correct. Even if we do keep the eager evaluation of `_current`, I would strongly suggest that we modify the code to ensure `_index` and `_current` are consistent.

nielsenko: We could start `_index` on -1, and increment before setting `_current`

nielsenko: This would allow us to implement `T get current => _results[_index]`, but I still tend to prefer an explicit `_current` that is set in `moveNext`.. just for the odd chance that collection is changed after `moveNext`, but before calling `current`

nirinchev: It should not change assuming both calls happen on the same thread without yielding the thread between them, but as I said, I don't have strong opinions on the subject since the assumption is that `moveNext` and `current` will always be called as a pair. The reason we don't do eager evaluation in .NET is because it makes the code more concise and we're lazy 😝

desistefanova: It is my opinion, but let's leave this PR for removeMany like it is. The implementation of the iterator will be changed for sure later. We have other issues also that we need to take in consideration: allowing nullable objects in results, editing or not objects while iterate, snapshot or not, concurrently changing objects order for sorted results and changing results collection for filtered results. We probably will implement common iterator for RealmList and RealmResults. I can prepare a list of test cases. It is important to have a decision how apps developers would expect it to behave. Let's discuss it with Lubo first.

lib/src/results.dart
desistefanova: Btw, in .Net you will also receive exception at runtime if you are trying to change collection while iterate. I think it is expected. Otherwise we may introduce unpredictable and hard to detect problems. Let me think a little bit.

nielsenko: Yes for the BCL, but not for `RealmCollectionBase`. As @nirinchev state, that is a deliberate choice.

desistefanova: Yes, it is acceptable not to throw exceptions but only in case we have a snapshot, otherwise would be a problem.

nielsenko (approved):

blagoev:
also we should snapshot before iteration

lib/src/results.dart
blagoev: I tend to agree with @nirinchev. And we need to shapshot the collection.

nirinchev:
Note that snapshotting is only possible for object collections -  I don't think dart supports results of primitives yet, but when you do, you won't be able to snapshot those.

lib/src/results.dart
blagoev: I created a task for that for the alpha, we need to support this

test/realm_test.dart
```diff
@@ -552,5 +552,183 @@ Future<void> main([List<String>? args]) async {
...
+      int itemsCount = teams.length;
+      for (var i = 0; i < itemsCount; i++) {
+        list.add(teams[i]);
+      }
+
+      //Remove objects in the list from realm
+      realm.write(() => realm.removeMany(list));
+
+      //Reload teams from realm and ensure they are removed
+      teams = realm.all<Team>();
```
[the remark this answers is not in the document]
desistefanova: Yes, we discussed this with @nirinchev. We agreed that both could be checked the collection before reload and after reload. I have it in most of the test. If you think it is too much I will remove reloading from all the test.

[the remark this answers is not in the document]
blagoev: no its fine, I noticed the discussion. This comment was done before I knew it was discussed. Feel free do it as you see fit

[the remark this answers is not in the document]
desistefanova: This test doesn't check whether the list is empty, because I don't clear the list. I only remove objects from realm database. We have to decide here, whether the user would expect the collection to be also cleared or not.

blagoev:
blocked by https://github.com/realm/realm-core/issues/5154

lib/src/results.dart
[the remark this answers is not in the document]
desistefanova: Taking in consideration your comments @nielsenko and @nirinchev the code was changed to the following. Snapshotting will be implemented with task #166. Are you OK to resolve this discussion?

`class _RealmResultsIterator<T extends RealmObject> implements Iterator<T> {
  final RealmResults<T> _results;
  int _index;
  T? _current;

  _RealmResultsIterator(RealmResults<T> results)
      : _results = results,
        _index = -1;

  @override
  T get current => _current as T;

  @override
  bool moveNext() {
    int length = _results.length;
    _index++;
    if (_index >= length) {
      _current = null;
      return false;
    }
    _current = _results[_index];

    return true;
  }`

test/realm_test.dart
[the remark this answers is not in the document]
desistefanova: Done.

lib/src/native/realm_core.dart
desistefanova: Renamed to realmListDeleteAll and realmResultsDeleteAll.

lib/src/realm_class.dart
desistefanova: Formatted.

lib/src/results.dart
desistefanova: Since @blagoev has created issue #163 I think we can resolve this discussion. @nirinchev, thank you for the helpful advice!

lib/src/results.dart
desistefanova: @nirinchev and @nielsenko thank you for the comments! I removed throwing concurrent exception. We will make a snapshot with issue #166. May I resolve this conversation?

lib/src/results.dart
desistefanova: I changed index to start from -1, so that now current and index are at the same point as @nirinchev required. I kept having _current member and evaluation in advance as @nielsenko suggested. Snapshot will be done with issue #166 - discussed with @blagoev. If you agree I would resolve this conversation.

test/realm_test.dart
```diff
@@ -552,5 +552,183 @@ Future<void> main([List<String>? args]) async {
...
+      realm.write(() {
+        realm.add(Team()..name = "Ferrari");
+        realm.add(Team()..name = "Maserati");
+      });
+
+      //Ensure the teams exist in realm
+      var teams = realm.all<Team>();
+      expect(teams.length, 2);
+
+      //Convert RealmResults to List (asList should be implemented)
```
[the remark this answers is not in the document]
desistefanova: The comment and converting to list are removed. They are no more needed.

lib/src/native/realm_core.dart
```diff
@@ -384,6 +384,13 @@ class _RealmCore {
     });
   }
 
+  void realmListDeleteAll(RealmList list) {
```
blagoev, line 387: are we deleting or removing items from the list. I guess we ment `deleting` is when we delte objects from the realm but when we `remove` we just remove them from the list and objects stay in the realm. So I guess this should be `realmListRemoveAll`

blagoev (changes requested):
consider name change

lib/src/native/realm_core.dart
```diff
@@ -384,6 +384,13 @@ class _RealmCore {
...
+    _realmLib.invokeGetBool(() => _realmLib.realm_list_remove_all(list.handle._pointer));
+  }
+
+  void realmResultsDeleteAll(RealmResults results) {
```
blagoev, line 391:
```suggestion
  void resultsDeleteAll(RealmResults results) {
```

lib/src/native/realm_core.dart
blagoev, line 387:
```suggestion
  void listDeleteAll(RealmList list) {
```

blagoev (changes requested):

nielsenko (approved):

lib/src/native/realm_core.dart
desistefanova: Done

lib/src/native/realm_core.dart
desistefanova: Done.

blagoev (approved):

PR #157, opened 2022-01-05
update realm core

nielsenko:
- Update realm-core
- Minor refactorings


nielsenko:
Updates to realm-core. In particular it uses the new version of `realm_get_last_error` that takes a `realm_error_t` as argument, so life-time can be controlled by Arena

lib/src/native/realm_core.dart
```diff
@@ -61,25 +61,25 @@ class _RealmCore {
...
-      if (error == nullptr) {
-        return null;
-      }
-
-      String? message;
-      if (error.ref.message != nullptr) {
-        message = error.ref.message.cast<Utf8>().toDartString();
-      }
-      _realmLib.realm_release_last_error(error);
+  LastError? getLastError(Arena arena) {
```
blagoev, line 64: why explicit `Arena` here. `Allocator` is more generic.

lib/src/native/realm_core.dart
```diff
@@ -460,33 +458,27 @@ class RealmListHandle extends Handle<realm_list> {
...
-  }
 }
 
 extension _RealmLibraryEx on RealmLibrary {
   void invokeGetBool(bool Function() callback, [String? errorMessage]) {
     bool success = callback();
     if (!success) {
-      final lastError = realmCore.getLastError();
-      throw RealmException("${errorMessage ?? ""} ${lastError.toString()}");
+      realmCore.raiseLastErrorAsRealmException(errorMessage);
```
blagoev, line 474:
```suggestion
      realmCore.throwLastError(errorMessage);
```

lib/src/native/realm_core.dart
```diff
@@ -61,25 +61,25 @@ class _RealmCore {
...
 
-    return using((Arena arena) {
-      return getLastError(arena);
+    return LastError(error.ref.error, message);
+  }
+
+  void raiseLastErrorAsRealmException([String? errorMessage]) {
+    using((Arena arena) {
+      final lastError = getLastError(arena);
+      throw RealmException("${errorMessage ?? ""} ${lastError.toString()}");
```
blagoev, line 82: `lastError` can be null here. same as the old code actually

blagoev: also this will result in a bit weird error message. 
Previously it was 
```Dart
"Class $className not found in ${realm.config.path}. Error: table not found"
```
now it will be 
```Dart
"Class $className not found in ${realm.config.path} table not found`
```

blagoev, line 82:
```suggestion
      throw RealmException("${errorMessage ?? ""} Error: ${lastError.toString()}");
```

blagoev (changes requested):
also , I suggest we wait for the snapshot C API then update this PR.

lib/src/native/realm_core.dart
nielsenko: Actually LastError.toString cause it to format like: `Class $className not found in ${realm.config.path} Error Code: ${lastError.code} Message: table not found`. I will add appropriage punctuation.

lib/src/native/realm_core.dart
nielsenko: Fixed

lib/src/native/realm_core.dart
nielsenko: Fixed

lib/src/native/realm_core.dart
nielsenko: Fixed

nielsenko: Recall that `lastError.toString` writes `Error code: ${lastError.code}` potentially followed by `Message: ${lastError.message}` if such exists. But I have added a '. ' between, if `message` is set.

blagoev: ok I see. resolving

blagoev: 👍 resolving

blagoev:
I suggest we wait for the C API result fixes in Core before we merge it.

blagoev (approved):
we decided to update core again later with the snapshots fixes

PR #159, opened 2022-01-06
realm.addAll

nielsenko:
- Rudimentrary implementation of Realm.addAll
- Add test


lib/src/realm_class.dart
```diff
@@ -83,6 +83,12 @@ class Realm {
     return object;
   }
 
+  void addAll<T extends RealmObject>(Iterable<T> items) {
+    for (final i in items) { // fallback
```
blagoev, line 87:
```suggestion
    for (final i in items) {
```

blagoev (approved):

PR #160, opened 2022-01-06
update macos runners

nielsenko:
- Use macos-latest instead of macos-10.15 (https://github.com/actions/virtual-environments/issues/4060)


blagoev:
seems like the base of this PR should be something else. too many changes

blagoev (approved):

PR #168, opened 2022-01-11
runs-on: macos-10.15

desistefanova:
A try to fix Android build failures


blagoev:
@desistefanova  I suggest we close this PR. The problem seems not related to using `macos-latest`. Its mostly the flaky nature of Android emulators

PR #169, opened 2022-01-12
realm.deleteRealm

### LINKED ISSUE #49: Implement Realm.deleteRealm
Static member of Realm that takes as an argument realm configuration and deletes all files and directories of realm. Throws RealmException if realm is not closed or if some IO exception happen. Use realm_delete_files from core API.

desistefanova:
Realm deleteRealm implemented - fixes #49


nirinchev (changes requested):
This should use the Core API instead of manually deleting the files. It's not a problem when you're the only process accessing a file, but it becomes an issue with multiprocess access.

desistefanova:
> This should use the Core API instead of manually deleting the files. It's not a problem when you're the only process accessing a file, but it becomes an issue with multiprocess access.

@nirinchev yes, we have delete_files function in the core API for .Net, but I couldn't find such method exposed in realm.h for Dart sdk. That's why we decided to implement it in Dart.

blagoev:
Indeed we are referencing core v11.8 which does not have the delete files api. It was only recently added in https://github.com/realm/realm-core/commit/9913f7dea74465274258796933aba6b031876c38
Let me update core to newer version and we can change the implementation to use that instead of Dart APIs. I think we will support only sync delete with the core version.

desistefanova:
> Indeed we are referencing core v11.8 which does not have the delete files api. It was only recently added in [realm/realm-core@9913f7d](https://github.com/realm/realm-core/commit/9913f7dea74465274258796933aba6b031876c38) Let me update core to newer version and we can change the implementation to use that instead of Dart APIs. I think we will support only sync delete with the core version.

It will be great. I will change it as soon as we have the new version.

blagoev:
@desistefanova ok I merged https://github.com/realm/realm-dart/pull/170 which includes `realm_delete_files` C API

lib/src/realm_class.dart
```diff
@@ -62,6 +63,11 @@ class Realm {
     }
   }
 
+  static void deleteRealm(Configuration config) {
```
nielsenko, line 66: Should we spawn an isolate and make this async?

lib/src/realm_class.dart
```diff
@@ -62,6 +63,11 @@ class Realm {
...
+    // TODO: To discuss: Do we need to return this bool result from realmCore?
```
nielsenko, line 67: Why not? This is IO, the User needs to know if operation succeeded or not

lib/src/native/realm_core.dart
```diff
@@ -186,6 +186,15 @@ class _RealmCore {
     return RealmHandle._(realmPtr);
   }
 
+  bool deleteRealmFiles(String path) {
+    return using((Arena arena) {
+      Pointer<Uint8> realmDeleted = arena<Uint8>();
+      _realmLib.invokeGetBool(() => _realmLib.realm_delete_files(path.toUtf8Ptr<Int8>(arena), realmDeleted), "Error deleting realm at path ${path}");
+      // TODO: To discuss: Is this the right way to check whether file is deleted according to the doc?
```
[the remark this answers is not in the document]
blagoev: yes this is correct way to check it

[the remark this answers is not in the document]
blagoev: I suggest we throw an exception here if the delete failed and not return a result.

lib/src/native/realm_core.dart
```diff
@@ -186,6 +186,15 @@ class _RealmCore {
     return RealmHandle._(realmPtr);
   }
 
+  bool deleteRealmFiles(String path) {
```
blagoev, line 189:
```suggestion
  void deleteRealmFiles(String path) {
```

lib/src/realm_class.dart
blagoev: I don't think we should try doing that. This should be a sync operation as all other core related.

lib/src/realm_class.dart
blagoev: this should throw a `RealmException` if it fails, imo

test/realm_test.dart
```diff
@@ -1012,5 +1012,29 @@ Future<void> main([List<String>? args]) async {
       //Ensure list size is the same like teams collection size
       expect(list.length, teams.length);
     });
+
+    test('Realm deleteRealm files after closing realm', () {
```
blagoev, line 1016:
```suggestion
    test('Realm deleteRealm succeeds', () {
```

test/realm_test.dart
```diff
@@ -1012,5 +1012,29 @@ Future<void> main([List<String>? args]) async {
...
+      var config = Configuration([Dog.schema, Person.schema]);
+      var realm = Realm(config);
+      
+      realm.close();
+      Realm.deleteRealm(config);
+
+      expect(File(config.path).existsSync(), false);
+      // TODO: For Discussion: Lock file is not deleted by core API.
```
blagoev, line 1024: this is fine as it is mentioned in the header. we should not check for it

test/realm_test.dart
```diff
@@ -1012,5 +1012,29 @@ Future<void> main([List<String>? args]) async {
...
+      //expect(File("${config.path}.lock").existsSync(), false);
+      expect(Directory("${config.path}.management").existsSync(), false);
+    });
+
+    test('Realm deleteRealm files before closing realm expects exception', () {
```
blagoev, line 1029:
```suggestion
    test('Realm deleteRealm throws exception on an open realm', () {
```

test/realm_test.dart
```diff
@@ -1012,5 +1012,29 @@ Future<void> main([List<String>? args]) async {
...
+      var config = Configuration([Dog.schema, Person.schema]);
+      var realm = Realm(config);
+
+      expect(() => Realm.deleteRealm(config), throws<RealmException>());
+      
+      expect(File(config.path).existsSync(), true);
+      expect(File("${config.path}.lock").existsSync(), true);
```
blagoev, line 1036: imo, we should not check for the lock file in the test.

blagoev (changes requested):

lib/src/realm_class.dart
blagoev, line 66: As the comment on `Realm.exists` I think we should require a path here. Please check my reasoning [here ](https://github.com/realm/realm-dart/pull/171#discussion_r783772877)

blagoev (changes requested):

lib/src/realm_class.dart
blagoev: @nielsenko @desistefanova ^

desistefanova:
> > Indeed we are referencing core v11.8 which does not have the delete files api. It was only recently added in [realm/realm-core@9913f7d](https://github.com/realm/realm-core/commit/9913f7dea74465274258796933aba6b031876c38) Let me update core to newer version and we can change the implementation to use that instead of Dart APIs. I think we will support only sync delete with the core version.
> 
> It will be great. I will change it as soon as we have the new version.

Done.

lib/src/realm_class.dart
nielsenko: That is fine with me as well. As long as it cannot fail silently.

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
desistefanova: Here is the description of `realm_delete_files`:
/// @param[out] did_delete_realm If non-null, set to true if the primary Realm file was deleted.
/// Discard value if the function returns an error.
/// @return true if no error occurred.

lib/src/realm_class.dart
nielsenko: Okay, I just don't like blocking IO on the main thread.

lib/src/realm_class.dart
nirinchev: It won't fail silently, but it may succeed even if an exception is called, which is why there's a return value.

nielsenko: I mean on the dart side.. Or maybe I misunderstand?

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
nielsenko: True, `invokeGetBool` will raise exception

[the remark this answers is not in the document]
blagoev: yep

lib/src/realm_class.dart
desistefanova: After a discussion we decided to have only one sync method deleteReam for now.

lib/src/realm_class.dart
desistefanova: After a discussion we decided to leave the exception to be thrown by `invokeGetBool`. Return type will be void.

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
desistefanova: Fixed in https://github.com/realm/realm-dart/pull/169/commits/0697dce4c21edbd697003715971e4e21b9d68d23

lib/src/realm_class.dart
desistefanova: Fixed in https://github.com/realm/realm-dart/pull/169/commits/0697dce4c21edbd697003715971e4e21b9d68d23

test/realm_test.dart
desistefanova: Fixed in https://github.com/realm/realm-dart/pull/169/commits/0697dce4c21edbd697003715971e4e21b9d68d23

test/realm_test.dart
desistefanova: Fixed in https://github.com/realm/realm-dart/pull/169/commits/0697dce4c21edbd697003715971e4e21b9d68d23

lib/src/native/realm_core.dart
desistefanova: Fixed in https://github.com/realm/realm-dart/pull/169/commits/0697dce4c21edbd697003715971e4e21b9d68d23

lib/src/native/realm_core.dart
```diff
@@ -186,6 +186,13 @@ class _RealmCore {
     return RealmHandle._(realmPtr);
   }
 
+  void deleteRealmFiles(String path) {
+    return using((Arena arena) {
```
blagoev, line 190:
```suggestion
    using((Arena arena) {
```

blagoev (commented):
👍 with small change requested

nielsenko (approved):

blagoev (approved):

desistefanova:
> This should use the Core API instead of manually deleting the files. It's not a problem when you're the only process accessing a file, but it becomes an issue with multiprocess access.

@nirinchev we've changed the implementation to use `realm_delete_files` C API. So it is done.

PR #171, opened 2022-01-13
realm.exists and realm.existsSync

### LINKED ISSUE #50: Implement Realm.exists and Realm.existsSync
Takes string path to a realm and checks if the realm file exists. Static member

desistefanova:
Implement Realm.exists - fixes #50

### DIFF
--- a/lib/src/realm_class.dart
+++ b/lib/src/realm_class.dart
@@ -63,21 +63,21 @@ class Realm {
     }
   }
 
-  static bool existsSync(Configuration config) {
+  static bool existsSync(String path) {
     try {
-      final fileEntity = File(config.path);
+      final fileEntity = File(path);
       return fileEntity.existsSync();
     } catch (e) {
-      throw RealmException("Error while checking if Realm exists at ${config.path}. Error: $e");
+      throw RealmException("Error while checking if Realm exists at ${path}. Error: $e");
     }
   }
 
-  static Future<bool> exists(Configuration config) async {
+  static Future<bool> exists(String path) async {
   try {
-      final fileEntity = File(config.path);
+      final fileEntity = File(path);
       return await fileEntity.exists();
     } catch (e) {
-      throw RealmException("Error while searching for realm file. Error: $e");
+      throw RealmException("Error while checking if Realm exists at ${path}. Error: $e");
     }
   }
 
--- a/lib/src/realm_class.dart
+++ b/lib/src/realm_class.dart
@@ -68,7 +68,7 @@ class Realm {
       final fileEntity = File(config.path);
       return fileEntity.existsSync();
     } catch (e) {
-      throw RealmException("Error while searching for realm file. Error: $e");
+      throw RealmException("Error while checking if Realm exists at ${config.path}. Error: $e");
     }
   }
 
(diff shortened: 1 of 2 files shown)


lib/src/realm_class.dart
```diff
@@ -62,6 +63,23 @@ class Realm {
...
+      throw RealmException("Error while searching for realm file. Error: $e");
+    }
+  }
+
+  static Future<bool> exists(Configuration config) async {
+    final fileEntity = File(config.path);
+    return fileEntity
+        .exists()
+        .then((value) => value)
+        .onError((error, stackTrace) => throw RealmException("Error while searching for realm file. Error: $error"));
```
nielsenko, line 80:
```suggestion
  try {
      final fileEntity = File(config.path);
      return await fileEntity.exists();
    } catch (e) {
      throw RealmException("Error while searching for realm file. Error: $e");
    }
```

lib/src/realm_class.dart
```diff
@@ -62,6 +63,23 @@ class Realm {
     }
   }
 
+  static bool existsSync(Configuration config) {
+    try {
+      final fileEntity = File(config.path);
+      return fileEntity.existsSync();
+    } catch (e) {
+      throw RealmException("Error while searching for realm file. Error: $e");
```
blagoev, line 71: we could include the path here.

blagoev, line 71:
```suggestion
      throw RealmException("Error while checking if Realm exists at ${config.path}. Error: $e");
```

lib/src/realm_class.dart
blagoev, line 80: same suggestion as above

blagoev: +1 for this suggestion

lib/src/realm_class.dart
```diff
@@ -62,6 +63,23 @@ class Realm {
     }
   }
 
+  static bool existsSync(Configuration config) {
```
blagoev, line 66: I know we discussed passing a configuration here but on a second thought we should probably require a path here imo. It's somewhat unnatural to require a config which requires a schema to be constructed. So if we require a path and there is a config around one can do 
```Dart
Realm.existsSync(config.path);
```
or if there is not a config around they can do 
```Dart
Realm.existsSync(Configuration.defaultPath); //or pass a custom string here. 
```

I think its more natural and especially since this is a static method we shouldn't require a config which is somewhat an instance concept.

blagoev: @nielsenko  @desistefanova what do you think about this ^

blagoev (changes requested):

lib/src/realm_class.dart
nielsenko: I have also changed my mind, and agree :-)

desistefanova: It is a strong argument. I agree @blagoev.

desistefanova: Fixed in https://github.com/realm/realm-dart/pull/171/commits/36ef83d9efb04bc3700725948649336692d23e86

lib/src/realm_class.dart
desistefanova: OK

lib/src/realm_class.dart
desistefanova: OK

blagoev (approved):

PR #173, opened 2022-01-13
results.isEmplty 

### LINKED ISSUE #38: Support Results empty
Results should support isEmpty field

desistefanova:
New test added for results.isEempty - fixes #38

### DIFF
--- a/test/realm_test.dart
+++ b/test/realm_test.dart
@@ -497,6 +497,33 @@ Future<void> main([List<String>? args]) async {
       expect(cars.isEmpty, true);
     });
 
+    test('Results.query() isEmpty', () {
+      var config = Configuration([Dog.schema, Person.schema]);
+      var realm = Realm(config);
+
+      final dogOne = Dog()
+        ..name = "Pupu"
+        ..age = 1;
+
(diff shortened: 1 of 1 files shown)


test/realm_test.dart
```diff
@@ -497,6 +497,33 @@ Future<void> main([List<String>? args]) async {
       expect(cars.isEmpty, true);
     });
 
+    test('Results.query() isEmpty', () {
```
blagoev, line 500: perhaps  `Results.all() isEmpty` test should be named `Results isEmpty` and then we don't need this one. But it does not hurt to have it. Feel free to leave it if you would like.

blagoev (approved):

test/realm_test.dart
desistefanova: Results.all() isEmpty was existing and I created the new one similar to it. But you are right we have both test cases in the new test. I will remove the first test and will rename the second test.

desistefanova: Actually, both tests are a little bit different. It is good to have them both. Just will rename them.

blagoev (approved):

PR #174, opened 2022-01-13
Fix adding object with list properties

### LINKED ISSUE #164: Fix default values for List members and adding Realm Objects with Lists members
Currently this throws an exception.  
```Dart
final team = Team()..name = "Ferrari"..players.add(Person()..name = "Player1");
realm.write(() => realm.add(team));
```
When setting the default values when adding objects we should handle List members explicitly and iterate them adding them to the realm. 
This should walk over the entire tree of references handling other lists member in nested objects as well. 

Lists members should have default values set as well

```Dart
_defaultsSet ??= RealmObject.setDefaults<Team>({"players": <Person?>[]});
```

blagoev:
fixes #164

### DIFF
--- a/lib/src/realm_object.dart
+++ b/lib/src/realm_object.dart
@@ -18,6 +18,7 @@
 
 import 'native/realm_core.dart';
 import 'realm_class.dart';
+import 'list.dart';
 
 abstract class RealmAccessor {
   Object? get<T extends Object>(RealmObject object, String name);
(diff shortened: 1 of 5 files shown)


blagoev:
CI fails so this needs more work

lib/src/realm_object.dart
```diff
@@ -141,6 +142,20 @@ class RealmCoreAccessor implements RealmAccessor {
   void set(RealmObject object, String name, Object? value, [bool isDefault = false]) {
     final propertyMeta = metadata[name];
     try {
+      if (value is List) {
+        if (value.isEmpty) {
+          return;
+        }
+
+        //This assumes the targe list property is empty. `valus is List` should happen only when making a RealmObject managed
```
desistefanova, line 150: //This assumes the target list property is empty. `value is List` should happen only when making a RealmObject managed

desistefanova (approved):

nielsenko:
There is a discrepancy between how we set and get a list on like players, ie `RealmObject.set<List<Person>>(this, 'players', <Person>[])` and `Realm.get<Person>(this, 'players') as List<Person>` ie. type argument to set and get differs.

Perhaps we could fix that eventually?

nielsenko (approved):

lib/src/realm_object.dart
blagoev: Could you elaborate?

blagoev: there is a typo. Fill fix

blagoev: fixed

PR #175, opened 2022-01-13
results.length

### LINKED ISSUE #39: Support Results length
Results should support reading its length (the total number of objects in this result)

desistefanova:
Results length tests implemented. Fixes #39

### DIFF
--- a/lib/src/realm_class.dart
+++ b/lib/src/realm_class.dart
@@ -19,6 +19,7 @@
 import 'dart:ffi';
 import 'dart:io';
 import 'dart:isolate';
+import 'dart:mirrors';
 
 import 'results.dart';
 import 'configuration.dart';
@@ -95,10 +96,15 @@ class Realm {
       throw RealmException("Object type ${object.runtimeType} not configured in the current Realm's schema."
           " Add type ${object.runtimeType} to your config before opening the Realm");
     }
-
-    final handle = metadata.class_.primaryKey == null
-        ? realmCore.createRealmObject(this, metadata.class_.key)
-        : realmCore.createRealmObjectWithPrimaryKey(this, metadata.class_.key, metadata.class_.primaryKey!);
+    RealmObjectHandle handle;
+    if (metadata.class_.primaryKey == null) {
+      handle = realmCore.createRealmObject(this, metadata.class_.key);
+    } else {
+      var reflection = reflect(object);
+      InstanceMirror field = reflection.getField(Symbol(metadata.class_.primaryKey!));
+      Object? primaryKeyValue = field.reflectee as Object?;
+      handle = realmCore.createRealmObjectWithPrimaryKey(this, metadata.class_.key, primaryKeyValue!);
+    }
 
     final accessor = RealmCoreAccessor(metadata);
     object.manage(this, handle, accessor);
--- a/lib/src/realm_class.dart
+++ b/lib/src/realm_class.dart
@@ -19,7 +19,6 @@
 import 'dart:ffi';
 import 'dart:io';
 import 'dart:isolate';
-import 'dart:mirrors';
 
 import 'results.dart';
 import 'configuration.dart';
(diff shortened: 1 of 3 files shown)


test/realm_test.dart
```diff
@@ -463,21 +463,46 @@ Future<void> main([List<String>? args]) async {
...
       expect(cars.length, 0);
+
+      final carOne = Car()..make = "Toyota";
+      final carTwo = Car()..make = "Toyota";
+      realm.write(() => realm.addAll([carOne, carTwo]));
+
+      expect(cars.length, 2);
+
+      final filteredCars = realm.query<Car>('make == "Toyota"');
+      // TODO: To be discussed. The length here is returned 1, should be 2
```
blagoev, line 503: Make is primary key on car, so it is really one object

blagoev (approved):

test/realm_test.dart
```diff
@@ -463,21 +463,46 @@ Future<void> main([List<String>? args]) async {
...
+      //expect(filteredCars.length, 2);
+
```
blagoev, line 505: Consider enabling this expect with 1

test/realm_test.dart
desistefanova: Oh, you are right. But then why it allows me to add two objects with the same key? And also in results coming from the query it returns 2 items, only the length is 1 since it is taken with `realmCore.getResultsCount(this)`

test/realm_test.dart
```diff
@@ -463,21 +463,46 @@ Future<void> main([List<String>? args]) async {
...
-      final car = Car();
-      realm.write(() => realm.add(car));
+      final carOne = Car()..make = "Toyota 1";
+      final carTwo = Car()..make = "Toyota 2";
+      final carThree = Car()..make = "Renault";
+      realm.write(() => realm.addAll([carOne, carTwo, carThree]));
+
+      expect(cars.length, 3);
+
+      final filteredCars = realm.query<Car>('make BEGINSWITH "Toyota"');
```
blagoev, line 480:
```suggestion
      final filteredCars = realm.query<Car>('make BEGINSWITH "Toyot"');
```

desistefanova: Done

test/realm_test.dart
desistefanova: Fixed in https://github.com/realm/realm-dart/pull/175/commits/e3083d94da8f9328738f03fcdef7c0daaa555871

test/realm_test.dart
desistefanova: Fixed method `add` in realm to throw exception for duplicate primary keys. Commit https://github.com/realm/realm-dart/pull/175/commits/e3083d94da8f9328738f03fcdef7c0daaa555871

desistefanova: C API method `realm_object_create_with_primary_key` expects primary key **value** instead of primary key column name.

test/realm_test.dart
```diff
@@ -463,21 +490,44 @@ Future<void> main([List<String>? args]) async {
       expect(cars, isNotNull);
     });
 
-    test('Results.all() length', () {
+    test('Results length with deleted by filter', () {
```
desistefanova, line 493:
```suggestion
    test('Results length after deleted objects by filter', () {
```

lib/src/realm_class.dart
```diff
@@ -95,10 +95,13 @@ class Realm {
...
     }
-
-    final handle = metadata.class_.primaryKey == null
-        ? realmCore.createRealmObject(this, metadata.class_.key)
-        : realmCore.createRealmObjectWithPrimaryKey(this, metadata.class_.key, metadata.class_.primaryKey!);
+    RealmObjectHandle handle;
+    if (metadata.class_.primaryKey == null) {
+      handle = realmCore.createRealmObject(this, metadata.class_.key);
+    } else {
+      Object? primaryKeyValue = RealmObject.get(object, metadata.class_.primaryKey!);
```
blagoev, line 102: we should use the `accessor` to get the value. In general RealmObject.get should be called only by the generated code since this is where we have the `T` (type) information and this is compile time only. I will push the fix into the branch as we discussed.

blagoev (changes requested):

test/realm_test.dart
blagoev: good catch.

lib/src/realm_class.dart
blagoev: fixed

blagoev (approved):

PR #176, opened 2022-01-16
Flutter desktop support

blagoev:
Adds Flutter Windows support.
Updates flutter example app
Bumps all flutter constraints to 2.8
Bumps all dart constraints to 2.15
Updates Realm flutter tests app to support Windows platform.
Adds CI support for Flutter Windows.

* This does not execute any metrics on build.


nielsenko (commented):
Consider dropping the pubspec.lock file on the example app.

blagoev:
@nielsenko will do. I am not sure this is the best approach, but lets do it. we can add it later if we encounter any issues.

nielsenko (approved):

desistefanova (approved):

PR #177, opened 2022-01-17
Test sort via Realm.query

nielsenko:


test/realm_test.dart
```diff
@@ -614,6 +614,22 @@ Future<void> main([List<String>? args]) async {
...
+            Person()..name = "Sebastian",
+            Person()..name = "Kimi",
+          ]));
+
+      final result = realm.all<Person>().sort('name');
+      final resultNames = result.map((p) => p.name).toList();
+      final sortedNames = [...resultNames]..sort();
+      expect(resultNames, sortedNames);
+    });
+
```
desistefanova, line 632: If we have items with equal names and different keys. After working with the collection (delete/add items) they could be returned with different indexes. See this test. It fails actually.

`test('Sort result twice returns different index', () {
      var config = Configuration([Cat.schema]);
      var realm = Realm(config);
      Cat cat1 = Cat(1)
        ..name = "Bella"
        ..age = 1;
      Cat cat2 = Cat(2)
        ..name = "Bella"
        ..age = 2;
      Cat cat3 = Cat(3)
        ..name = "Oliver"
        ..age = 3;
      realm.write(() => realm.addAll([cat1, cat2, cat3]));

      final result1 = realm.all<Cat>().sort('name');
      List<int> list1 = [];
      for (var item in result1) {
        list1.add(item.id);
      }
      realm.write(() => realm.delete(cat1));
      realm.write(() => realm.add(Cat(1)
        ..name = "Bella"
        ..age = 1));

      final result2 = realm.all<Cat>().sort('name');
      List<int> list2 = [];
      for (var item in result2) {
        list2.add(item.id);
      }
      expect(list1, list2);
    });`

desistefanova: In case of having some virtualized collection or some other specific logic that requires indexes to remains the same, the developers should be able to do sorting by name and then sorting by id. Something like .sort('name').sort("id"). Is there any way to do it at once with the query language? Or what about doing always sorting by primary key at the end if there is some?

desistefanova (approved):
With suggested test about changing sort indexes after delete/add operations.

test/realm_test.dart
nielsenko: You cannot expect the indexes of a live query to be stable under db operations, unless you snapshot the result. Regarding the double sort, I'm not sure actually

nielsenko: BTW, I'm retiring the sort method until we can build queries linq style. You can do the same with query though.

test/realm_test.dart
```diff
@@ -664,6 +664,54 @@ Future<void> main([List<String>? args]) async {
...
+      expect(result, orderedEquals(snapshot));
+      expect(result.map((d) => d.name), snapshot.map((d) => d.name));
+      result = realm.query<Dog>('TRUEPREDICATE SORT(name ASC)'); // redoing query won't change that
+      expect(result, orderedEquals(snapshot));
+      expect(result.map((d) => d.name), snapshot.map((d) => d.name));
+
+      realm.write(() => realm.delete(dog1)); // result will update, snapshot will not, but an object has died
+
+      expect(() => snapshot[0].name, throwsA(isA<RealmException>()));
```
blagoev, line 703:
```suggestion
      expect(() => snapshot[0].name, throws<RealmException>());
```

blagoev: we have an convenience method :)

PR #180, opened 2022-01-17
Override operator== for RealmObject

nielsenko:


lib/src/realm_object.dart
```diff
@@ -206,6 +206,16 @@ class RealmObject {
...
   }
+
+  @override
+  // ignore: hash_and_equals
+  bool operator ==(Object other) {
+    final o = other;
+    if (identical(this, o)) return true;
+    if (o is! RealmObject) return false;
+    if (!isManaged || !o.isManaged) return false;
+    return realmCore.equals(handle, o.handle);
```
blagoev, line 217: I would suggest we follow the other places and keep passing realm objects to realm_core and get the handles from inside the realm_core implementation. I will put suggestions on the relevant places

lib/src/native/realm_core.dart
```diff
@@ -442,6 +442,10 @@ class _RealmCore {
   void listClear(RealmList list) {
     _realmLib.invokeGetBool(() => _realmLib.realm_list_clear(list.handle._pointer));
   }
+
+  bool equals<T extends NativeType>(Handle<T> x, Handle<T> y) {
```
blagoev, line 446:
```suggestion
  bool equals(RealmObject first, RealmObject second) {
```

lib/src/realm_object.dart
blagoev, line 217:
```suggestion
    return realmCore.equals(this, other);
```

lib/src/realm_object.dart
```diff
@@ -206,6 +206,16 @@ class RealmObject {
     RealmAccessor.setDefaults<T>(values);
     return true;
   }
+
+  @override
+  // ignore: hash_and_equals
+  bool operator ==(Object other) {
+    final o = other;
```
blagoev, line 213: perhaps we don't need an additional local variable.

test/realm_test.dart
```diff
@@ -1073,5 +1069,29 @@ Future<void> main([List<String>? args]) async {
...
+        realm..add(person)..add(dog);
+      });
+
+      expect(person, person);
+      expect(person, isNot(1)); 
+      expect(person, isNot(dog));
+
+      final read = realm.query<Person>("name == 'Kasper'");
+
+      expect(read, [person]);
```
blagoev, line 1094: Minor: Not sure what the error message when comparing collections, I would have written `expect(read[0], person);` Hope it's readable.

blagoev (changes requested):
I suggest some small changes

test/realm_test.dart
blagoev: consider ignoring this if you think its alright

lib/src/realm_object.dart
nielsenko: Fixed

lib/src/realm_object.dart
nielsenko: Fixed

lib/src/native/realm_core.dart
nielsenko: Fixed

lib/src/realm_object.dart
nielsenko: Fixed

test/realm_test.dart
nielsenko: I think it is a bit safer like this. If read is empty, read[0] throws before expect gets to run. Also, the output is quite nice, but we should consider generating a toString method for realm objects.

nielsenko: Also `expect(read[0], person)` don't check if `read.length > 1`

lib/src/native/realm_core.dart
```diff
@@ -442,6 +442,10 @@ class _RealmCore {
   void listClear(RealmList list) {
     _realmLib.invokeGetBool(() => _realmLib.realm_list_clear(list.handle._pointer));
   }
+
+  bool equals<T extends NativeType>(RealmObject first, RealmObject second) {
```
blagoev, line 446:
```suggestion
  bool equals(RealmObject first, RealmObject second) {
```

test/realm_test.dart
blagoev: sure, scratch my comment

lib/src/native/realm_core.dart
blagoev: I think we don't need the type constraint on this method.  Added a suggestion

test/realm_test.dart
blagoev: scratch my comment

blagoev (approved):

PR #182, opened 2022-01-18
Implement Realm.operator==

### LINKED ISSUE #161: Support Realm equals
Support equals for Realm instance

nielsenko:


lib/src/realm_class.dart
```diff
@@ -194,6 +194,14 @@ class Realm {
     final handle = realmCore.queryClass(this, metadata.class_.key, query, args);
     return RealmResultsInternal.create<T>(handle, this);
   }
+
+  @override
+  // ignore: hash_and_equals
+  bool operator ==(Object other) {
+    if (identical(this, other)) return true;
+    if (other is! Realm) return false;
+    return realmCore.realmEquals(this, other);
```
nielsenko, line 203: I think this will always return false. I cannot invent a way, where this would return true, but `identical(this, other)` would not?

nielsenko: Even after exposing `RealmObject.realm`

desistefanova (approved):

lib/src/realm_class.dart
nirinchev: Are you caching realm instances at the dart layer? In most SDKs, you can get different managed instances that point to the same native one, in which case `identical` (which if I understand correctly is equivalent to `ReferenceEquals`) would return `false`.

nielsenko: Correct. All calls start from a realm, and all objects created will be given this realm, even query results. Anyway, as I have implemented it here, I don't depend on it.

nirinchev: I meant, like:

```dart
var config = Configuration()
  ..path = "/tmp/my-realm"

var realm1 = Realm(config);
var realm2 = Realm(config);

// I would expect
assert(identical(realm1, realm2), false);
assert(realm1 == realm2, true);
```

nielsenko: As you may notice in the test that is not the case. Good to know, it is the expected behaviour.

nirinchev: Yeah, this sounds like a bug in Core/the C API in that case. Those should be the same native instance due to OS caching.

nielsenko: We are using realm-core v11.9.0.

nielsenko:
Parked until after Alpha release

nielsenko:
I would suggest making `snapshot` public. There are other ways to walk around the result set that also require the collection to be stable.

lib/src/configuration.dart
```diff
@@ -97,6 +97,14 @@ class Configuration {
   /// The realm file must already exists at [path]
   bool get isReadOnly => realmCore.getConfigReadOnly(this);
   set isReadOnly(bool value) => realmCore.setConfigReadOnly(this, value);
+
+  @override
+  bool operator ==(Object other) {
+    if (identical(this, other)) return true;
+    if (other is! Configuration) return false;
+    bool areEquals = realmCore.configurationEquals(this, other);
+    return areEquals;
```
nielsenko, line 106:
```suggestion
    return realmCore.configurationEquals(this, other);
```

lib/src/realm_class.dart
```diff
@@ -232,6 +232,15 @@ class Realm {
...
     return RealmResultsInternal.create<T>(handle, this);
   }
+
+  @override
+  // ignore: hash_and_equals
+  bool operator ==(Object other) {
+    if (identical(this, other)) return true;
+    if (other is! Realm) return false;
+    bool areEquals = realmCore.realmEquals(this, other);
+    return areEquals;
```
nielsenko, line 242:
```suggestion
    return realmCore.realmEquals(this, other);
```

lib/src/configuration.dart
desistefanova: done

lib/src/realm_class.dart
desistefanova: done

PR #185, opened 2022-01-18
Add support for Flutter for MacOS

blagoev:
This adds Flutter for MacOS support.
updates flutter example and tests apps
Fixes library loading for Flutter on MacOS
*Does not run metrics on build


.github/workflows/realm-dart-windows.yml
```diff
@@ -26,10 +26,7 @@ jobs:
           submodules: 'recursive'
 
       - name: Build Ralm Dart for Windows
```
nielsenko, line 28: Ralm should be Realm

.github/workflows/realm-flutter-windows.yml
```diff
@@ -26,10 +26,7 @@ jobs:
           submodules: 'recursive'
 
       - name: Build Ralm Dart for Windows
```
nielsenko, line 28: Ralm should be Realm

blagoev: can you elaborate? This step name is  `Realm Dart for Windows` cause this is what it is doing. Or you refer to something else

nielsenko (changes requested):
Remove the flutter_XX.png files

.github/workflows/realm-flutter-windows.yml
blagoev: oh yeah there is a typo. got it. will fix

nielsenko: Minor type 'Ralm'

blagoev:
@nielsenko These are added by the flutter tooling.

.github/workflows/realm-flutter-windows.yml
nielsenko: Nothing to do with this PR, just noticed it

blagoev:
@nielsenko I got your meaning. will fix

nielsenko:
The example should work regardless. Feels a bit odd with 0 bytes files, and error screens in the repo.

nielsenko (approved):

.github/workflows/realm-flutter-windows.yml
blagoev: fixed

PR #186, opened 2022-01-19
Add notifications to RealmResults

nielsenko:


test/realm_test.dart
```diff
@@ -709,10 +709,46 @@ Future<void> main([List<String>? args]) async {
...
+        });
+        yield 0; // to allow event to fire, if any
+        expect(changeCount, 2); // once per transaction, not once per change
+
+        subscription.cancel();
+      }())
+          .drain(0), // test don't support an async* testFunction, so we drain to make it plain async
+    );
+
+    test('Lists create object with a list property', () async {
```
desistefanova, line 751: Should this test method be async?

desistefanova (approved):

test/realm_test.dart
nielsenko: No, that is not needed. My mistake.

nielsenko: Fixed

lib/src/results.dart
```diff
@@ -196,6 +197,9 @@ class RealmResults<T extends RealmObject> extends collection.IterableBase<T> {
   // void set length(int newLength) {
   //   throw new Exception("Setting length on Results<T> is not supported");
   // }
+
+  // TODO: For now changed don't look at the RealmCollectionChangesHandles
+  Stream<RealmResults> get changed => realmCore.resultChanged(this, _realm.scheduler.handle).map((_) => this);
```
nielsenko, line 202: Should we call it `changes` instead? So user will write: `results.changes.listen(() { ... });` ?

nielsenko: I think I prefer `changed` still.. Once we support full change sets, it will read like `results.changed.listen((changes) { ... });`

PR #187, opened 2022-01-19
Use Flutter 2.10 beta and Dart 2.16 beta on CI

blagoev:
Use Flutter 2.10 beta and Dart 2.16 beta on CI
update flutter action to v2
bump android compileSdkVersion 31 (required by Flutter 2.10 beta)


nielsenko:
We could consider building for both stable and beta, using a matrix

nielsenko (approved):

PR #196, opened 2022-01-21
cycle referenced objects tests

### LINKED ISSUE #195: Create tests for cycle referenced objects
Create new tests that add a graph of realm objects with cycle references.

desistefanova:
Some new tests created for adding a graph of realm objects with cycle references - fixes #195


test/realm_test.dart
```diff
@@ -56,6 +56,25 @@ class _Team {
...
+  late int? yearOfBirth;
+  late _School? school;
+}
+
+@RealmModel()
+class _School {
+  @PrimaryKey()
+  late final String name;
+  late String? city;
+  final List<_Student> students = [];
```
desistefanova, line 73: I suggest we can implement the generator to insist the lists to be define as **final** without **late** and to be initialized **[]**. This is how the model is convenient to be used, because we will have compile time error that setter of the list is not able to be used. Otherwise they can use it, but the data are not stored in realm.
If the generator insists declaration of the lists to be like this 
`final List<_Student> students = [];`
 then this mistake 
`school.students = [student1, student2]`
will not be allowed and early visible.

test/realm_test.dart
```diff
@@ -1182,20 +1201,78 @@ Future<void> main([List<String>? args]) async {
...
+      //Check schools
+      var schools = realm.all<School>();
+      expect(schools.length, 3);
+
+      //Check students
+      var students = realm.all<Student>();
+      expect(students.length, 3);
+
+      //Check branches
+      var branches = realm.all<School>().query(r'branchOfSchool != nil');
```
blagoev, line 1266:
```suggestion
      var branches = realm.all<School>().query('branchOfSchool != nil');
```
```suggestion
      var branches = realm.all<School>().query(r'branchOfSchool != nil');
```

test/realm_test.dart
```diff
@@ -1182,20 +1201,78 @@ Future<void> main([List<String>? args]) async {
...
+      expect(branches.length, 2);
+      expect(branches[0].students.length + branches[1].students.length, 3);
+
+      //Check main schools
+      var mainSchools = realm.all<School>().query(r'branchOfSchool = nil');
```
blagoev, line 1271:
```suggestion
      var mainSchools = realm.all<School>().query('branchOfSchool = nil');
```

blagoev: r"", means regex. and is only useful when we have a query with arguments like query(r"name == $0", [myname]) or one could write query("name == \$0", [myname]) escaping the `$`

test/realm_test.dart
```diff
@@ -1182,20 +1201,78 @@ Future<void> main([List<String>? args]) async {
...
       expect(person, isNot(dog));
 
       final read = realm.query<Person>("name == 'Kasper'");
 
       expect(read, [person]);
     });
   });
+  group('Cycle referenced objects tests:', () {
+    test('Realm adding objects graph', () {
+      var studentMichele = Student(1)
```
blagoev, line 1222: minor: could we write on one line 
but more importantly aren't these available in the constructor so we could write like below

test/realm_test.dart
```diff
@@ -1182,20 +1201,78 @@ Future<void> main([List<String>? args]) async {
...
+        ..name = "Michele Ernesto"
+        ..yearOfBirth = 2005;
+      var studentLoreta = Student(2)
```
blagoev, line 1225:
```suggestion
      var studentLoreta = Student(2, {name  = "Loreta", yearOfBirth = 2006})
```

blagoev: we could write some or all of these like that to demonstrate both,

blagoev (approved):
good test, added some comments

test/realm_test.dart
desistefanova: Fixed in https://github.com/realm/realm-dart/pull/196/commits/e058f5ce6ce1ce13fc88117feebaad19dd5896c7
Both ways are shown.

blagoev:
@desistefanova  merging this for you

PR #199, opened 2022-01-21
throw RealmUnsupportedSetError on call to setter of late final field on a realm object.

nielsenko:


common/lib/src/realm_types.dart
```diff
@@ -52,14 +52,23 @@ enum RealmCollectionType {
...
 }
 
+class RealmError extends Error {}
+
+class RealmUnsupportedError extends UnsupportedError implements RealmError {
+  RealmUnsupportedError(String message) : super(message);
+}
+
+class RealmUnsupportedSetError extends RealmUnsupportedError {
+  RealmUnsupportedSetError() : super('Cannot set late final field on realm object');
```
desistefanova, line 62: It will be nice to have some suggestion in the text about how they can set this field if the setter is not supported. But this also would work for now.

nielsenko: I would like to keep it simple. There are (typically) two kinds of late final fields: primary keys, and lists. It differs a bit what to do, ie. set it in ctor, or call `x.clear()..addAll([ ... ])`

desistefanova (approved):
Looks OK

common/lib/src/realm_types.dart
desistefanova: Yes it is better to be simple.

common/lib/src/realm_types.dart
```diff
@@ -52,14 +52,23 @@ enum RealmCollectionType {
   dictionary,
 }
 
+class RealmError extends Error {}
+
+class RealmUnsupportedError extends UnsupportedError implements RealmError {
```
blagoev, line 57: I wonder do we gain much by extending errors here. Since these are not supposed to be catched by user code? Wouldn't a simple `RealmUnsupportedError ` throwing the relevant message suffice here. I expect string constants are interned in Dart

lib/src/realm_class.dart
```diff
@@ -31,7 +31,7 @@ export 'results.dart' hide RealmResultsInternal;
 export 'realm_object.dart'
     hide RealmObjectInternal, RealmAccessor, RealmValuesAccessor, RealmMetadata, RealmCoreAccessor, RealmClassMetadata, RealmPropertyMetadata;
 export "configuration.dart" hide ConfigurationInternal;
-export 'package:realm_common/realm_common.dart' show RealmModel, PrimaryKey, Ignored, MapTo, Indexed, RealmPropertyType, RealmCollectionType;
+export 'package:realm_common/realm_common.dart' hide Decimal128, RealmAny, Uuid;
```
blagoev, line 34: I liked the explicit show more. Thus we have control what is exposed and we don't slip something unintentionally

common/lib/src/realm_types.dart
nielsenko: They are, I just like to have a type hierarchy in exceptions, if you need to catch them.

blagoev: Fully agree for exceptions, but for Errors seems unnecessary since these should fail the app and are not supposed to be catched.

lib/src/realm_class.dart
blagoev: yes indeed its long but it could be single Error type (per my comment above) and we better have a long list here than unintentionally exposing something. I agree it's not ideal, but since there are no visibility modifiers...

common/lib/src/realm_types.dart
nielsenko: I still like the distinction during development / debugging. You could argue the same for all the different types of error in dart. Why not just have one Error class?

lib/src/realm_class.dart
nielsenko: To me it is clearer to read that we explicitly would like to hide 3 classes. It would easy to hide one of these in a long list of show declarations.

common/lib/src/realm_types.dart
nielsenko: I reduced the hierarchy by one level

blagoev (approved):

lib/src/realm_class.dart
```diff
@@ -28,10 +28,18 @@ import 'list.dart';
...
+export 'realm_object.dart' show RealmException, RealmObject;
 export 'realm_object.dart'
     hide RealmObjectInternal, RealmAccessor, RealmValuesAccessor, RealmMetadata, RealmCoreAccessor, RealmClassMetadata, RealmPropertyMetadata;
+
+export "configuration.dart" show Configuration, RealmSchema, SchemaObject;
 export "configuration.dart" hide ConfigurationInternal;
-export 'package:realm_common/realm_common.dart' show RealmModel, PrimaryKey, Ignored, MapTo, Indexed, RealmPropertyType, RealmCollectionType;
+
+export 'package:realm_common/realm_common.dart'
+    show Ignored, Indexed, MapTo, RealmError, RealmModel, RealmUnsupportedSetError, RealmCollectionType, RealmPropertyType;
```
desistefanova, line 40: I don't see PrimaryKey.

blagoev: good catch

lib/src/realm_class.dart
```diff
@@ -28,10 +28,18 @@ import 'list.dart';
 
 export 'list.dart' hide RealmListInternal;
 export 'results.dart' hide RealmResultsInternal;
+
+export 'realm_object.dart' show RealmException, RealmObject;
 export 'realm_object.dart'
     hide RealmObjectInternal, RealmAccessor, RealmValuesAccessor, RealmMetadata, RealmCoreAccessor, RealmClassMetadata, RealmPropertyMetadata;
+
+export "configuration.dart" show Configuration, RealmSchema, SchemaObject;
```
desistefanova, line 36: Missing SchemaProperty.

lib/src/realm_class.dart
```diff
@@ -28,10 +28,18 @@ import 'list.dart';
...
 export "configuration.dart" hide ConfigurationInternal;
-export 'package:realm_common/realm_common.dart' show RealmModel, PrimaryKey, Ignored, MapTo, Indexed, RealmPropertyType, RealmCollectionType;
+
+export 'package:realm_common/realm_common.dart'
+    show Ignored, Indexed, MapTo, RealmError, RealmModel, RealmUnsupportedSetError, RealmCollectionType, RealmPropertyType;
+export 'package:realm_common/realm_common.dart' hide Decimal128, RealmAny, Uuid, ObjectId, RealmObjectMarker;
+
 export 'realm_property.dart';
```
desistefanova, line 43: Missing SchemaProperty exposed explicitly.

lib/src/realm_class.dart
nielsenko: Hmm.. why does it work then 🤔

nielsenko: Well, that kind of prove your point doesn't it 😄 .. I have removed the hide, and only use show now

PR #200, opened 2022-01-21
Fix need for cast with extension method

nielsenko:


lib/src/list.dart
```diff
@@ -25,6 +25,16 @@ import 'native/realm_core.dart';
 import 'realm_object.dart';
 import 'realm_class.dart';
 
+extension RealmListExtension<T extends Object> on List<T> {
+  bool get isValid {
+    final self = this;
+    if (self is RealmList<T>) {
+      return self.isValid;
+    }
+    return false;
```
desistefanova, line 34: If it is not RealmList, should return true.

nielsenko: Fixed

desistefanova: Actually this logic is only for RealmObject. .Net and Kotlin SDK don't have IsValid for their base collections.

PR #204, opened 2022-01-23
update readme

blagoev:
Update README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ This repository holds the source code for the Realm SDK for Flutterâ„¢ and Dartâ
     }
     ```
 
-* Generate Generates RealmObject class `Car` from data model class `_Car`.
+* Generate RealmObject class `Car` from data model class `_Car`.
 
     ```
     flutter run realm generate
--- a/README.md
+++ b/README.md
@@ -196,7 +196,7 @@ The Realm Dart package is `realm_dart`
     dart pub add realm_dart
     ```
 
-* Install the `realm_dart` package into the application. This downloads and copies the required native binaries to the app directory
+* Install the `realm_dart` package into the application. This downloads and copies the required native binaries to the app directory.
 
     ```
     dart run realm_dart install
@@ -220,7 +220,7 @@ The Realm Dart package is `realm_dart`
                         generate_for:
                             - bin/*.dart 
         ```
-    * To generate RealmObject classes with realm_dart use this command
+    * To generate RealmObject classes with realm_dart use this command.
 
         ```
         dart run realm_dart generate
@@ -230,7 +230,7 @@ The Realm Dart package is `realm_dart`
         _*This file should be committed to source control_
 ## Usage
 
-For usage of Realm Dart see the Realm Flutter usage above
+For usage of Realm Dart see the Realm Flutter usage above.
 
 
 # Building the source
@@ -287,7 +287,7 @@ For usage of Realm Dart see the Realm Flutter usage above
 
 ### Versioning
 
-Realm Flutter and Dart SDK packages follow [Semantic Versioning](https://semver.org/)
+Realm Flutter and Dart SDK packages follow [Semantic Versioning](https://semver.org/).
 During the initial development the packages will be versioned according the scheme `0.major.minor+release stage` until the first stable version is reached then packages will be versioned with `major.minor.patch` scheme.
 
 The first versions will follow `0.1.0+preview`, `0.1.1+preview` etc.
--- a/README.md
+++ b/README.md
@@ -5,12 +5,285 @@
 Realm is a mobile database that runs directly inside phones, tablets or wearables.
 This repository holds the source code for the Realm SDK for Flutter™ and Dart™.
 
-# Alpha
-
 **This project is in the Alpha stage, All API's might change without warning and no guarantees are given about stability. Do not use it in production.**
 
-The previous preview version of Realm Dart can be found on this branch: https://github.com/realm/realm-dart/tree/preview
+## Getting Started
 
+* Import Realm.
+    ```dart
+    import 'package:realm/realm.dart';
+    ```
+
+* Define a data model class `_Car`.
+    ```dart
+    @RealmModel()
+    class _Car {
+      late String make;
+
+      late String model;
+    
+      int? kilometers = 500;
+    }
+    ```
+
+* Generate Generates RealmObject class `Car` from data model class `_Car`.
+
+    ```
+    flutter run realm generate
+    ```
+
+* Open a Realm and add some objects.
+
+    ```dart
+    var config = Configuration([Car.schema]);
+    var realm = Realm(config);
+
+    var car = Car("Telsa", "Model Y", kilometers: 5);
+    realm.write(() {
+      realm.add(car);
+    }
+    ```
+
+* Query objects in Realm.
+
+    ```dart
+    var cars = realm.all<Car>();
+    Car myCar = cars[0];
+    print("My car is ${myCar.make} model ${myCar.model}");
+
+    cars = realm.all<Car>().query("make == 'Tesla'");
+    ```
+
+## Samples
+
+For complete samples check the [Realm Flutter and Dart Samples](https://github.com/realm/realm-dart-samples).
+
+## Documentation
+
+For API documentation go to 
+ * [Realm Flutter API Docs](https://pub.dev/documentation/realm/latest/)
+
+ * [Realm Dart API Docs](https://pub.dev/documentation/realm_dart/latest/)
+
+For a complete documentation go to [Realm Flutter and Dart SDK Docs](https://docs.mongodb.com/realm/sdk/flutter/).
+
+
+## Limitations
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,22 +5,288 @@
...
+        // Get object by index
+        var item = items[5];
+        
+        // Get object by primary key
+        var itemByKey = realm.find<Item>(0);
+        
+        // Filter and sort object
+        var objects = realm.query<Item>("name == 'Special Item'");
+        var mymake = 'Tesla';
+        var objects = realm.query<Item>(r'make == $0', [mymake]");
```
desistefanova, line 178:
```suggestion
        var objects = realm.query<Item>(r'make == $0', [mymake]);
```

README.md
```diff
@@ -5,22 +5,288 @@
...
+
+# Building the source
+
+## Building Realm Flutter
+
+    * clone the repo 
+    ```
+    git clone https://github.com/realm/realm-dart
+    git submodule udpate --init --recursive
+    ```
```
desistefanova, line 237:
```suggestion
* clone the repo
    ```
    git clone https://github.com/realm/realm-dart
    git submodule udpate --init --recursive
    ```
```

desistefanova: Moved tabulation because after generating the doc asterisk and "`" were displayed instead of ballet and code formatting.

README.md
```diff
@@ -5,22 +5,288 @@
...
+        scripts\build.bat
+        ```
+    * MacOS
+        ```
+        ./scripts/build-macos.sh
+        ```
+    * Linux
+        ```
+        ./scripts/build-linux.sh
+        ```
```
desistefanova, line 279:
```suggestion
* Android
    ```
    ./scripts/build-android.sh all
    scripts\build-android.bat all
    #or for emulator only
    ./scripts/build-android.sh x86
    scripts\build-android.bat x86
    ```

* iOS
    ```
    ./scripts/build-ios.sh
    #or for simulator only
    ./scripts/build-ios.sh simulator
    ```

* Windows
    ```
    scripts\build.bat
    ```

* MacOS
    ```
    ./scripts/build-macos.sh
    ```

## Building Realm Dart

* Windows
    ```
    scripts\build.bat
    ```
* MacOS
    ```
    ./scripts/build-macos.sh
    ```
* Linux
    ```
    ./scripts/build-linux.sh
    ```
```

desistefanova:
There are quite a lot declarative sentences with missing punctuation mark at the end.

nielsenko:
@desistefanova seems totally on top of this.

README.md
```diff
@@ -5,22 +5,288 @@
...
+        ```yaml
+        targets:
+            $default:
+                builders:
+                    realm_generator|realm_object_builder:
+                        enabled: true
+                        generate_for:
+                            - bin/*.dart 
+        ```
+
```
mongodben, line 223: i believe another step should be added around here with info on generating the models b/c it's a slightly different command from realm flutter. 

```sh
dart run realm_dart generate
```

README.md
blagoev: will fix

README.md
blagoev: good catch. The above command was wrong but I still will add this here as well

README.md
blagoev: fixed. good catch

README.md
blagoev: fixed

README.md
blagoev: fixed

README.md
```diff
@@ -5,22 +5,295 @@
...
+## Samples
+
+For complete samples check the [Realm Flutter and Dart Samples](https://github.com/realm/realm-dart-samples).
+
+## Documentation
+
+For API documentation go to 
+ * [Realm Flutter API Docs](https://pub.dev/documentation/realm/latest/)
+
+ * [Realm Dart API Docs](https://pub.dev/documentation/realm_dart/latest/)
```
desistefanova, line 66: Aren't these two links pointing again to this README?

blagoev: they point to the API docs on pub.dev

README.md
```diff
@@ -5,22 +5,295 @@
...
+
+* Define a data model class `_Car`.
+    ```dart
+    @RealmModel()
+    class _Car {
+      late String make;
+
+      late String model;
+    
+      int? kilometers = 500;
```
desistefanova, line 25: int? kilometers = 500;
Is this a good sample - nullable int with default value?

blagoev: I don't see why not. Do you think we can improve it?

README.md
```diff
@@ -5,22 +5,295 @@
...
+    }
+    ```
+
+* Generate Generates RealmObject class `Car` from data model class `_Car`.
```
desistefanova, line 29:
```suggestion
* Generate RealmObject class `Car` from data model class `_Car`.
```

README.md
```diff
@@ -5,22 +5,295 @@
...
+    flutter run realm generate
+    ```
+
+* Open a Realm and add some objects.
+
+    ```dart
+    var config = Configuration([Car.schema]);
+    var realm = Realm(config);
+
+    var car = Car("Telsa", "Model Y", kilometers: 5);
```
desistefanova, line 41: I think it will be nice to show how the generated constructor looks like before using it in the example.

blagoev: I think this is not a good place for that. The readme is our quick getting started guide. Such details are for elsewhere. This sample at the very beginning severs only one purpose show some code. A picture is worth a thousand words right. So sample code here is the same.

README.md
```diff
@@ -5,22 +5,295 @@
...
+    ```
+
+* Query objects in Realm.
+
+    ```dart
+    var cars = realm.all<Car>();
+    Car myCar = cars[0];
+    print("My car is ${myCar.make} model ${myCar.model}");
+
+    cars = realm.all<Car>().query("make == 'Tesla'");
```
desistefanova, line 54: Interesting: It seems the filter is working with both quotes combinations:
'make == "Tesla"' - this is how we are using it in tests
"make == 'Tesla'" - but this is more intuitive

blagoev: yes, this is parsed by core

README.md
```diff
@@ -5,22 +5,295 @@
...
+                        generate_for:
+                            - lib/*.dart 
+        ```
+
+    * Import Realm in a dart file (ex. `catalog.dart`).
+        ```dart
+        import 'package:realm/realm.dart';
+        ```
+
+    * Declare a part file `catalog.g.dart` in the begining of the `catalog.dart` dart file after all imports.
```
desistefanova, line 118: Maybe here we need a very short explanation that part file 'catalog.g.dart' will contain generated files. We have it below, but on this step is not clear why we are declaring part file.

README.md
desistefanova: Just wondering isn't it confusing.

README.md
blagoev: Lets leave it as it is. this is made clear on the generation step. Less words to read is better.

README.md
blagoev: I don't think so. It demonstrated both default values and nullable properties support

README.md
blagoev: fixed

README.md
```diff
@@ -5,22 +5,295 @@
...
+        ```dart
+        // Create a Configuration object
+        var config = Configuration(Item.schema);
+
+        // Opean a Realm
+        realm = Realm(config);
+
+        // Open a write transaction
+        realm.write(() {
+          var myItem = Item(0, 'Iteam', price: 4);
```
desistefanova, line 159: We usually create the objects in our tests before to open transaction. I think it is a good practice, but it could be like this also.

README.md
```diff
@@ -5,22 +5,295 @@
...
+
+## Usage
+
+* Add `realm_dart` package to a Dart application.
+
+    ```
+    dart pub add realm_dart
+    ```
+
+* Install the `realm_dart` package into the application. This downloads and copies the required native binaries to the app directory
```
desistefanova, line 199:
```suggestion
* Install the `realm_dart` package into the application. This downloads and copies the required native binaries to the app directory.
```

README.md
```diff
@@ -5,22 +5,295 @@
...
+
+        ```
+        dart run realm_dart generate
+        ```
+        A new file `catalog.g.dart` will be created next to the `catalog.dart`.
+        
+        _*This file should be committed to source control_
+## Usage
+
+For usage of Realm Dart see the Realm Flutter usage above
```
desistefanova, line 233:
```suggestion
For usage of Realm Dart see the Realm Flutter usage above.
```

README.md
```diff
@@ -5,22 +5,295 @@
...
+    ./scripts/build-macos.sh
+    ```
+* Linux
+    ```
+    ./scripts/build-linux.sh
+    ```
 
 ### Versioning
 
 Realm Flutter and Dart SDK packages follow [Semantic Versioning](https://semver.org/)
```
desistefanova, line 290:
```suggestion
Realm Flutter and Dart SDK packages follow [Semantic Versioning](https://semver.org/).
```

README.md
blagoev: it should work regardless. We shouldn't advise on that as a good practice. Doing both is ok

desistefanova (approved):

PR #209, opened 2022-01-23
Realm API documentation

desistefanova:
Documenting realm-dart library  - fixes private task RDART-150


common/lib/src/realm_common_base.dart
```diff
@@ -16,23 +16,59 @@
 //
 ////////////////////////////////////////////////////////////////////////////////
 
+///@nodoc
+late String _noDoc;
```
blagoev, line 20: seems a debug left over?

lib/realm.dart
```diff
@@ -1,3 +1,10 @@
+/// Realm SDK for Flutter and Dart
```
blagoev, line 1: copyright header should be at the top.

lib/realm.dart
```diff
@@ -15,8 +22,3 @@
 // limitations under the License.
 //
 ////////////////////////////////////////////////////////////////////////////////
-
```
blagoev, line 18: we need a library comment

blagoev (changes requested):

lib/realm.dart
blagoev: I will write one on the suggestions PR

common/lib/src/realm_common_base.dart
desistefanova: This is something I've added for discussion. I put this code here as workaround in order to prevent these slashes to appear in the doc API //////////. Is there some way I can do it?

lib/realm.dart
desistefanova: It is the same like the previous comment. The slashes ////////////// appear in the doc. I moved this line to prevent this.

blagoev:
closing this one in favor of #225

PR #210, opened 2022-01-24
Fix missing curly brackets in generated code

nielsenko:
Generator produces an if statement, that was missing curly brackets, causing a linter warning


nielsenko:
Maybe a bit unintended that this only runs the Flutter iOS & Android pipelines. Guess it happens since all changes are in the generator folder

blagoev (approved):

PR #214, opened 2022-01-24
disallow bool as index and primary key

nielsenko:
- Fixed.. may allow bools to be indexed later, even if it makes little sense (dotnet supports it)

### DIFF
--- a/generator/test/generator_test.dart
+++ b/generator/test/generator_test.dart
@@ -974,4 +974,86 @@ class _Bar {}
       )),
     );
   });
+
+  test('bool not allowed on indexed field (for now)', () async {
+    await expectLater(
+      () async => await testBuilder(
+        generateRealmObjects(),
+        {
+          'pkg|lib/src/test.dart': r'''
+import 'package:realm_common/realm_common.dart';
+
+part 'test.g.dart';
+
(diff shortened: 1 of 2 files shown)


generator/test/generator_test.dart
```diff
@@ -974,4 +974,86 @@ class _Bar {}
...
+            '9 │   late bool bad;\n'
+            '  │        ^^^^ bool is not an indexable type\n'
+            '  ╵\n'
+            'Change the type of \'bad\', or remove the @Indexed() annotation\n'
+            '',
+      )),
+    );
+  });
+
+  test('bool not allowed on indexed field (for now)', () async {
```
blagoev, line 1019:
```suggestion
  test('bool not allowed on indexed field', () async {
```

generator/test/generator_test.dart
```diff
@@ -974,4 +974,86 @@ class _Bar {}
       )),
     );
   });
+
+  test('bool not allowed on indexed field (for now)', () async {
```
blagoev, line 978:
```suggestion
  test('bool not allowed on indexed field', () async {
```

blagoev (approved):

PR #211, opened 2022-01-24
Ensure realm_generator will work without a build.yaml file

nielsenko:
Note this requires some changes to documentation. See the build.yaml files in the project

PR #220, opened 2022-01-25
Support realm object generation by default

blagoev:
This enables realm object generation after realm package is added to user project.
include build.yaml for realm_dart and realm package
removes dart example  build.yaml
leaves build.yaml for flutter example as an example for customized build.yaml
move build_runner as a dependency

This substitutes #211

### DIFF
--- a/build.yaml
+++ b/build.yaml
@@ -10,4 +10,12 @@ targets:
           - lib/src/cli/metrics/metrics.dart
       realm_generator|realm_object_builder:
         generate_for:
-          - test/**.dart
+          include:
+builders:
+  realm_generator:
+    import: "package:realm_generator/realm_generator.dart"
+    builder_factories: ["generateRealmObjects"]
+    build_extensions: { ".dart": ["realm_objects.g.part"] }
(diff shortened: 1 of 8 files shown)


desistefanova (approved):

nielsenko:
Well you know my stance on this. Quoting dart best practices on dependencies (https://dart.dev/tools/pub/dependencies):

> The rule for deciding between a regular or dev dependency is simple: If the dependency is imported from something in your lib or bin directories, it needs to be a regular dependency. If it’s only imported from test, example, etc. it can and should be a dev dependency.
> 
> Using dev dependencies makes dependency graphs smaller. That makes pub run faster, and makes it easier to find a set of package versions that satisfies all constraints.

I don't approve, but as technical solution it does the job

blagoev:
lets discuss this on a meeting

nielsenko (approved):
So we still don't agree fully on this, but we agree that we can always remove the direct dependency on `realm_generator` and `build_runner` later, if the constraints we pull in. causes trouble for our users. Let the world decide, as they say :-)

PR #221, opened 2022-01-25
use range for tar dep

blagoev:
use range for tar dep


desistefanova (approved):

nielsenko (commented):
Could you elaborate on why this is needed?

blagoev:
latest tar 0.5.1 does not work and crashes.

nielsenko (approved):

PR #223, opened 2022-01-26
warn on missing api docs

nielsenko:
- Warn on missing documentation of public api
- Fix type error (missing int)

### DIFF
--- a/generator/analysis_options.yaml
+++ b/generator/analysis_options.yaml
@@ -13,15 +13,18 @@
 
 include: package:lints/recommended.yaml
 
-# Uncomment the following section to specify additional rules.
+analyzer:
+  language:
+    strict-raw-types: true
 
-# linter:
-#   rules:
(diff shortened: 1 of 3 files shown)


generator/analysis_options.yaml
```diff
@@ -13,15 +13,18 @@
...
+    implicit-casts: false
+    implicit-dynamic: false
 
-# analyzer:
-#   exclude:
-#     - path/to/excluded/files/**
+linter:
+  rules:
+    avoid_relative_lib_imports: false
+    package_api_docs: true
```
blagoev, line 27: Should we enable this for the generator?

blagoev (approved):
Not sure we need API docs warnings in the generator

generator/analysis_options.yaml
nielsenko: Fixed

PR #224, opened 2022-01-26
Fix all linter violations

nielsenko:


blagoev:
@nielsenko I hope you don't mind I have hijacked this PR. most linter issues should be fixed now.

PR #225, opened 2022-01-26
V1 api docs

blagoev:
API Docs v1
This substitutes #209 
Removes `_RealmProperty` and moves all members to `SchemaProperty` to better document them


common/lib/src/realm_common_base.dart
```diff
@@ -1,4 +1,4 @@
-////////////////////////////////////////////////////////////////////////////////
+// //////////////////////////////////////////////////////////////////////////////
```
nielsenko, line 1: I assume this is to make pease with dart doc. Ie. it doesn't start with /// .. but it does look like an error. I suggest we use another "frame" eventually

nielsenko (approved):

PR #226, opened 2022-01-26
Update ffigen to latest stable

nielsenko:


ffigen/pubspec.yaml
```diff
@@ -5,16 +5,8 @@ description: >-
          dart run ffigen --config macos.yaml
          dart run ffigen --config linux.yaml
 
-documentation: 
-publish_to: none
```
blagoev, line 9: wouldn't it be better to keep this so we explicitly state that this is not to be published

lib/src/native/realm_core.dart
```diff
@@ -187,8 +187,8 @@ class _RealmCore {
 
   void deleteRealmFiles(String path) {
     using((Arena arena) {
-      final realm_deleted = arena<Uint8>();
```
blagoev, line 190: lets keep the underscore names as non idiomatic Dart names to keep treat them with care as pointers

lib/src/native/realm_core.dart
```diff
@@ -209,15 +209,19 @@ class _RealmCore {
   }
 
   void commitWrite(Realm realm) {
-    _realmLib.invokeGetBool(() => _realmLib.realm_commit(realm.handle._pointer), "Could commit write");
+    _realmLib.invokeGetBool(() => _realmLib.realm_commit(realm.handle._pointer), "Could not commit write");
```
blagoev, line 212: 👍

lib/src/native/realm_core.dart
```diff
@@ -209,15 +209,19 @@ class _RealmCore {
...
   }
 
   bool getIsWritable(Realm realm) {
     return _realmLib.realm_is_writable(realm.handle._pointer);
   }
 
   void rollbackWrite(Realm realm) {
-    _realmLib.invokeGetBool(() => _realmLib.realm_rollback(realm.handle._pointer), "Could rollback write");
+    _realmLib.invokeGetBool(() => _realmLib.realm_rollback(realm.handle._pointer), "Could not rollback write");
```
blagoev, line 220: 👍

lib/src/native/realm_core.dart
```diff
@@ -272,34 +276,48 @@ class _RealmCore {
...
+      final realmValue = _toRealmValue(value, arena);
+      _realmLib.invokeGetBool(() => _realmLib.realm_set_value(object.handle._pointer, propertyKey, realmValue.ref, isDefault));
     });
   }
 
-  // ignore: unused_element
-  int get _threadId => _realmLib.get_thread_id();
+  int get threadId => _realmLib.get_thread_id();
+
+  Object triggerGC() {
```
blagoev, line 302: this one does not really work. lets remove it. The only way to trigger the GC is through observatory manually.

lib/src/native/realm_core.dart
```diff
@@ -272,34 +276,48 @@ class _RealmCore {
...
-      final realm_value = _toRealmValue(value, arena);
-      _realmLib.invokeGetBool(() => _realmLib.realm_set_value(object.handle._pointer, propertyKey, realm_value.ref, isDefault));
+      final realmValue = _toRealmValue(value, arena);
+      _realmLib.invokeGetBool(() => _realmLib.realm_set_value(object.handle._pointer, propertyKey, realmValue.ref, isDefault));
     });
   }
 
-  // ignore: unused_element
-  int get _threadId => _realmLib.get_thread_id();
+  int get threadId => _realmLib.get_thread_id();
```
blagoev, line 300: this one I kept to ease debugging. Why expose it publicly?

lib/src/native/realm_core.dart
nielsenko: Not sure how that snuck in.. will remove

lib/src/native/realm_core.dart
nielsenko: I think I just fixed a warning. Will ignore instead.

lib/src/native/realm_core.dart
```diff
@@ -446,7 +452,7 @@ class LastError {
 }
 
 abstract class Handle<T extends NativeType> {
-  final Pointer<T> _pointer;
+  Pointer<T> _pointer;
```
blagoev, line 455: surprised to see final go here :)

nielsenko: I think it is a residue from when `releaseEarly` existed. I will add it back,

blagoev (approved):

PR #227, opened 2022-01-27
Fix premature realm close on Flutter

blagoev:
Fixes premature realm close in Flutter
Also update example to not include primarykeys to support multiple re-runs

### DIFF
--- a/lib/src/realm_class.dart
+++ b/lib/src/realm_class.dart
@@ -20,10 +20,9 @@ import 'dart:ffi';
 import 'dart:io';
 import 'dart:isolate';
 
-import 'package:realm_common/realm_common.dart';
-
-import 'results.dart';
+import '../realm.dart' show isFlutterPlatform;
 import 'configuration.dart';
+import 'results.dart';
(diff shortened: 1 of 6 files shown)


lib/src/realm_class.dart
```diff
@@ -256,8 +255,10 @@ class _Scheduler {
     final sendPort = receivePort.sendPort;
     handle = realmCore.createScheduler(Isolate.current.hashCode, sendPort.nativePort);
 
-    //We use this to receive a SCHEDULER_FINALIZE_OR_PROCESS_EXIT notification on process exit to close the receivePort or the process with hang.
-    Isolate.spawn(_handler, 2, onExit: sendPort);
+    if (!isFlutterPlatform) {
```
nielsenko, line 258: Are we sure it is safe when not running flutter?

blagoev: yes

nielsenko (dismissed):

nielsenko (approved):

PR #231, opened 2022-01-28
fix drop prepare command

nielsenko:
- Fix install command for macos
- Drop prepare_command from ios realm.podspec, remove comment with same from macos podspec, and do a bit of formatting

### DIFF
--- a/lib/src/cli/install/install_command.dart
+++ b/lib/src/cli/install/install_command.dart
@@ -203,11 +203,11 @@ class InstallCommand extends Command<void> {
 
     //We create a symlink to the binary file to be included as a app resource on build which needs to have an txt extension
     var binaryName = "librealm_dart.dylib";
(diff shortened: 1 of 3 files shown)


lib/src/cli/install/install_command.dart
```diff
@@ -203,11 +203,11 @@ class InstallCommand extends Command<void> {
 
     //We create a symlink to the binary file to be included as a app resource on build which needs to have an txt extension
```
nielsenko, line 204: Will be rebased on master soon

blagoev (changes requested):
for ios the prepare command is needed. 
for macos this is needed but lets leave it commented for now.

nielsenko:
Apparently prepare_command _is_ needed on ios. Will look into this later

PR #232, opened 2022-01-28
fix running package commands on flutter

blagoev:
This fixes running package commands on Flutter.


nielsenko (approved):
This is unexpected

PR #233, opened 2022-01-28
Preparea alpha.1

blagoev:
Moves example\myapp to example
Adds dardoc config to common
removes test dev_dep for realm


nielsenko (approved):

flutter/realm_flutter/pubspec.yaml
```diff
@@ -39,7 +39,6 @@ dev_dependencies:
   build_cli: ^2.1.3
   json_serializable: ^6.1.0
   lints: ^1.0.1
-  test: ^1.20.1
 
 # TODO: Remove once flutter_test from sdk is updated.
```
nielsenko, line 43: You should remove the dep override as well..

nielsenko (changes requested):

flutter/realm_flutter/pubspec.yaml
blagoev: removed it

nielsenko (approved):

PR #236, opened 2022-01-31
prepare alpha release

blagoev:
prepare alpha release


nielsenko:
./flutter/realm_flutter/example/pubspec.yaml version is not updated. Not sure it should even follow the rest, but right now it does

blagoev:
@nielsenko  updated in a follow-up commit

nielsenko (approved):

PR #238, opened 2022-02-03
Use dart specific C api for notification callbacks

nielsenko:
This has the advantage of not needing the static map in CallbackBridge, and hence no extra cleanup is needed dart side.

Another advantage is that we don't have to wait for dart side WeakReference in 2.17, since we can use the native side Dart_WeakPersistentHandle that already exists.

The extra C/C++ code is fairly negligible.


nielsenko:
This has been superseded

PR #239, opened 2022-02-04
Realm list.query

nielsenko:
Add query method to RealmList class

### DIFF
--- a/lib/src/list.dart
+++ b/lib/src/list.dart
@@ -87,12 +87,17 @@ class RealmList<T extends Object> extends collection.ListBase<T> {
 // The query operations on lists only work for list of objects (core restriction),
 // so we add it as an extension method to allow the compiler to prevent misuse.
 extension RealmListOfObject<T extends RealmObject> on RealmList<T> {
-  /// Filter list.
+  /// Filters the list and returns a new [RealmResults] according to the provided query.
   ///
-  /// @param query The query
-  /// @param args Possible parameters for substition in query
+  /// Only works for lists of Realm objects.
+  /// 
+  /// @param query The query used to filter the list
+  /// @param args Optional parameters for substitution in the query
   ///
   /// @return The live result
+  ///
+  /// The Realm Dart and Realm Flutter SDKs supports querying based on a language inspired by [NSPredicate](https://academy.realm.io/posts/nspredicate-cheatsheet/)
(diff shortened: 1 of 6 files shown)


test/realm_test.dart
```diff
@@ -788,6 +788,24 @@ Future<void> main([List<String>? args]) async {
...
+        final person = Person('Kasper');
+        final team = Team('Realm-dart', players: [
+          Person('Lubo'),
+          person,
+          Person('Desi'),
+        ]);
+
+        realm.write(() => realm.add(team));
+
+        final result = (team.players as RealmList<Person>).query(r'name BEGINSWITH $0', ['K']);
```
blagoev, line 804: I would add a TODO here to change this when `players is RealmList`

blagoev (approved):

test/realm_test.dart
nielsenko: Fixed

lib/src/list.dart
```diff
@@ -81,6 +81,11 @@ class RealmList<T extends Object> extends collection.ListBase<T> {
   void clear() {
     realmCore.listClear(this);
   }
+
+  RealmResults<T> query(String query, [List<Object> args = const []]) {
```
blagoev, line 85: we should add API doc here

nielsenko:
I'll convert to draft for now. This really depends on #162. The type restriction on `RealmResult` has been lifted in this PR, but we need to ensure we actually support it.

lib/src/list.dart
nielsenko: Fixed

nielsenko:
> I'll convert to draft for now. This really depends on #162. The type restriction on `RealmResult` has been lifted in this PR, but we need to ensure we actually support it.

Actually core does not support filters on primitives. PR has been refactored to align.

blagoev (approved):
with some suggestions

lib/src/native/realm_core.dart
```diff
@@ -364,6 +364,27 @@ class _RealmCore {
     });
   }
 
+    RealmResultsHandle queryList(RealmList target, String query, List<Object> args) {
+    return using((arena) {
```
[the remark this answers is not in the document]
nielsenko: You are right..

PR #240, opened 2022-02-04
Drop final requirement on primary keys for now

nielsenko:


desistefanova:
Do we already allow PK editing? Which is the related issue?

nielsenko:
> Do we already allow PK editing? Which is the related issue?

No. It will be a runtime error, if you try. But the point is that defining it as `late final String pk` it is still not a compile time error, due to the somewhat odd behaviour of public late final fields without initialisers. So @blagoev would like us to raise this restriction.

blagoev (changes requested):
please add changelog entry

blagoev (approved):

PR #241, opened 2022-02-04
Run generator test on CI

nielsenko:
Run generator tests on any changes to ./generator. Tests are run latest version of ubuntu, windows and macos, for both stable and beta.

### DIFF
--- a/.github/workflows/realm-dart-linux.yml
+++ b/.github/workflows/realm-dart-linux.yml
@@ -2,19 +2,22 @@ name: Realm Dart for Linux
 
 on:
   push:
-    branches: 
+    branches:
       - master
-    # Do not run if change is in these paths
     paths-ignore:
       - 'flutter/**'
       - 'generator/**'
+      - '.github/**'
+      - '!.github/workflows/realm-dart-linux.yml' # except this one
     tags:
       - '[0-9]+.[0-9]+.[0-9]+**' # matches tags like number(s).number(s).number(s)(any) for ex: 1.0.0 and also 1.0.0+beta
+
   pull_request:
-    # Do not run if change is in these paths
     paths-ignore:
-        - 'flutter/**'
-        - 'generator/**'
+      - 'flutter/**'
+      - 'generator/**'
+      - '.github/**'
+      - '!.github/workflows/realm-dart-linux.yml' # except this one
 
 jobs:
   CI:
--- a/.github/workflows/realm-generator.yml
+++ b/.github/workflows/realm-generator.yml
@@ -1,0 +1,65 @@
+name: Realm Generator
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'generator/**'
+      - .github/workflows/realm-generator.yml
+
+  pull_request:
+    paths:
+      - 'generator/**'
+      - .github/workflows/realm-generator.yml
+
+jobs:
+  Generator_CI:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest] # windows is not tested ue to bug in how error messages are rendered
+        channel: [stable, beta]
+
+    defaults:
+      run:
+        working-directory: ./generator/
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name : Setup Dart SDK
+        uses: dart-lang/setup-dart@main
+        with:
+          sdk: ${{ matrix.channel }}
+
+      - name: Install dependencies
+        run: dart pub get
+
+      - name: Run generator tests
+        run: dart test --reporter expanded --coverage ./coverage/
+
+      - name: Generate generator coverage report
+        run: |
+          dart run coverage:format_coverage \
+            --in coverage/ \
+            --out ./coverage/lcov.info \
+            --check-ignore \
+            --lcov \
+            --packages .packages \
+            --report-on lib
+
+      - name: Publish Generator Coverage
+        id: publish-generator-coverage
+        uses: coverallsapp/github-action@1.1.3
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          flag-name: realm_generator
+          path-to-lcov: ./coverage/lcov.info
+        continue-on-error: true
+
+      - name: Output Coveralls response
+        if: ${{ success() }}
+        run: echo ${{ steps.publish-generator-coverage.outputs.coveralls-api-result }}
--- a/.github/workflows/realm-dart-macos.yml
+++ b/.github/workflows/realm-dart-macos.yml
@@ -4,18 +4,20 @@ on:
   push:
     branches:
       - master
-    # Do not run if change is in these paths
     paths-ignore:
       - 'flutter/**'
       - 'generator/**'
+      - '.github/**'
+      - '!.github/workflows/realm-dart-macos.yml' # except this one
     tags:
       - '[0-9]+.[0-9]+.[0-9]+**' # matches tags like number(s).number(s).number(s)(any) for ex: 1.0.0 and also 1.0.0+beta
 
   pull_request:
-    # Do not run if change is in these paths
     paths-ignore:
-        - 'flutter/**'
-        - 'generator/**'
+      - 'flutter/**'
+      - 'generator/**'
+      - '.github/**'
+      - '!.github/workflows/realm-dart-macos.yml' # except this one
 
 jobs:
   CI:
--- a/.github/workflows/realm-dart-windows.yml
+++ b/.github/workflows/realm-dart-windows.yml
@@ -4,17 +4,20 @@ on:
   push:
     branches:
       - master
-    # Do not run if change is in these paths
     paths-ignore:
       - 'flutter/**'
       - 'generator/**'
+      - '.github/**'
+      - '!.github/workflows/realm-dart-windows.yml' # except this one
     tags:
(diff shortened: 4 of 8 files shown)


.github/workflows/realm-generator.yml
```diff
@@ -0,0 +1,64 @@
...
+      - 'generator/**'
+      - .github/workflows/realm-generator.yml
+
+jobs:
+  Generator_CI:
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest] # windows is not tested ue to bug in how error messages are rendered
```
desistefanova, line 22:
```suggestion
        os: [ubuntu-latest, macos-latest] # windows is not tested due to bug in how error messages are rendered
```

desistefanova, line 22:
```suggestion
        os: [ubuntu-latest, macos-latest] # windows is not tested due to bug in how error messages are rendered
```

desistefanova, line 22:
```suggestion
        os: [ubuntu-latest, macos-latest] # windows is not tested due to bug in how error messages are rendered
```

desistefanova, line 22:
```suggestion
        os: [ubuntu-latest, macos-latest] # windows is not tested due to bug in how error messages are rendered
```

desistefanova, line 22:
```suggestion
        os: [ubuntu-latest, macos-latest] # windows is not tested due to bug in how error messages are rendered
```

.github/workflows/realm-dart-linux.yml
```diff
@@ -2,19 +2,22 @@ name: Realm Dart for Linux
...
 on:
   push:
-    branches: 
+    branches:
       - master
-    # Do not run if change is in these paths
     paths-ignore:
       - 'flutter/**'
       - 'generator/**'
+      - '.github/**'
```
blagoev, line 10: Why not run on changes in these files? Isn't it required to run CI when changing CI scripts?

.github/workflows/realm-dart-linux.yml
```diff
@@ -2,19 +2,22 @@ name: Realm Dart for Linux
...
+      - '!.github/workflows/realm-dart-linux.yml' # except this one
```
blagoev, line 11: as above. seems counter intuitive not to run the linux CI when it's script changes.

.github/workflows/realm-generator.yml
blagoev, line 22: I suggest we test on linux only. It works the same on all platforms and fails the same as well.

.github/workflows/realm-generator.yml
```diff
@@ -0,0 +1,64 @@
...
+      - 'generator/**'
+      - .github/workflows/realm-generator.yml
+
+  pull_request:
+    paths:
+      - 'generator/**'
+      - .github/workflows/realm-generator.yml
+
+jobs:
+  Generator_CI:
```
blagoev, line 17: we use `CI` name  for the others. should this be any different?

.github/workflows/realm-generator.yml
```diff
@@ -0,0 +1,64 @@
...
+    runs-on: ${{ matrix.os }}
+
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest] # windows is not tested ue to bug in how error messages are rendered
+        channel: [stable, beta]
```
blagoev, line 23: I suggest testing this on stable only. Later we could handle `beta`  in some scheduled nightly runs

.github/workflows/realm-generator.yml
```diff
@@ -0,0 +1,64 @@
...
+
+      - name: Generate generator coverage report
+        run: |
+          dart run coverage:format_coverage \
+            --in coverage/ \
+            --out ./coverage/lcov.info \
+            --check-ignore \
+            --lcov \
+            --packages .packages \
+            --report-on lib
```
blagoev, line 52: Should this have `continue-on-error: true`?

.github/workflows/realm-generator.yml
```diff
@@ -0,0 +1,64 @@
...
+        id: publish-generator-coverage
+        uses: coverallsapp/github-action@1.1.3
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          flag-name: realm_generator
+          path-to-lcov: ./coverage/lcov.info
+        continue-on-error: true
+
+      - name: Output Coveralls response
+        run: echo ${{ steps.publish-generator-coverage.outputs.coveralls-api-result }}
```
blagoev, line 64: If the above fails does this fail and should it be skipped if the above fails?

.github/workflows/realm-dart-linux.yml
```diff
@@ -2,19 +2,22 @@ name: Realm Dart for Linux
 
 on:
   push:
-    branches: 
+    branches:
       - master
```
blagoev, line 6: these needs to run on any branch not just master

blagoev (changes requested):

.github/workflows/realm-dart-linux.yml
nielsenko: The ! means this path is not ignored.. I'm ignoring changes to the other actions

.github/workflows/realm-generator.yml
nielsenko: As you can see from the comment on windows, that is actually not guaranteed. This should run rarely, so I think it is okay to test both platform.. I'll raise a bug report regarding the windows issue

.github/workflows/realm-dart-linux.yml
nielsenko: It does on PRs

.github/workflows/realm-generator.yml
nielsenko: 👌

nielsenko: Fixed

.github/workflows/realm-dart-linux.yml
blagoev: yes and it should run on any other branch as well. Merging into `dev` branch should build if we have it. Now this change restricts us working only in master

.github/workflows/realm-dart-linux.yml
blagoev: so this ignores everything in github except linux workflow?

.github/workflows/realm-generator.yml
blagoev: I wonder why do we name it differently? Any value in that?

.github/workflows/realm-generator.yml
blagoev: `continue-on-error: true`  here will break the build? The way I read it this will fail the build currently  if `Generate generator coverage report` fails.

.github/workflows/realm-dart-linux.yml
nielsenko: Yes

.github/workflows/realm-dart-linux.yml
nielsenko: The only thing I'm changing in that line is that I remove a space at the end. I don't think you want us to do full CI on push to any branch, including feature branches we are still working on? Or maybe I don't understand your comment..

.github/workflows/realm-generator.yml
nielsenko: No `continue-on-error: true` means that the job will continue, even if this step fails.

.github/workflows/realm-generator.yml
nielsenko: No, but let me test for `${{ success() }}`

.github/workflows/realm-generator.yml
nielsenko: It shows up separately in the github UI. 

image

nielsenko: Notice btw the failed steps, that don't break the job

.github/workflows/realm-generator.yml
blagoev: This is my understanding as well. This is why I asked. So should this step have the same configuration?

.github/workflows/realm-dart-linux.yml
blagoev: Pardon me. I didn't read this right.

.github/workflows/realm-generator.yml
blagoev: Is this a bug in Actions or Windows in general

.github/workflows/realm-generator.yml
blagoev: ok I think my question is not valid since this has if {success}

.github/workflows/realm-generator.yml
nielsenko: Sorry, now I understand.. was confusing with the next step.. my mistake. No - this does not depend on an external service. I think it should just break the build, if it fails

.github/workflows/realm-generator.yml
nielsenko: This is bug in error output on windows... it still works, it just doesn't look as good as on the other platforms

blagoev: I see, ok lets leave it as a matrix test

.github/workflows/realm-generator.yml
blagoev: I am not sure we care so much for the coverage and if tests are green than this should be a nice to have. But I don't expect this to fail to often or at all, so lets leave it as it is.

blagoev (approved):

PR #242, opened 2022-02-05
Results notifications

blagoev:
This fixes up results notifications.
* Public API needs to be discussed before merging.


src/realm_dart_collections.cpp
```diff
@@ -19,84 +19,82 @@
...
-            auto h = Dart_HandleFromWeakPersistent_DL(handle_);
-            // Note Dart_IsNull(h) is not exposed in DL, so we cannot check.
-            // Hence the callback has to be prepared for this eventuality.
-            callback_(h, changes);
+    void callback(const realm_collection_changes_t* changes) {
+        if (m_handle) {
+            HandleScope scope;
+            //TODO: HACK. We can not release Dart persitent handles in delete_handle on Isolate teardown since the IsolateGroup is destroyed before it.
+            //This works since Dart_WeakPersistentHandle is equivalent to Dart_FinalizableHandle. They both are FinalizablePersistentHandle internally.
+            Dart_WeakPersistentHandle weakHnd = reinterpret_cast<Dart_WeakPersistentHandle>(m_handle);
```
nielsenko, line 60: This is not safe. `m_handle` is really an `Dart_FinalizableHandle` which will be auto deleted. We do not know that at this point that it is not (Depends on the order the finalizers for the notification_token_t and m_handle).

Also, this depends on an implementation detail in the dart VM. Google can change this without even bumping the minor version, and still be compliant with semantic versioning.. not that I think they are highly likely to change this soon.

nielsenko, line 60: This is not safe. `m_handle` is really an `Dart_FinalizableHandle` which will be auto deleted. We do not know that at this point that it is not (Depends on the order the finalizers for the notification_token_t and m_handle).

Also, this depends on an implementation detail in the dart VM. Google can change this without even bumping the minor version, and still be compliant with semantic versioning.. not that I think they are highly likely to change this soon.

src/realm_dart_collections.cpp
```diff
@@ -19,84 +19,82 @@
...
 }
 
-RLM_API realm_notification_token_t *
-realm_dart_results_add_notification_callback(realm_results_t *results,
-                                             Dart_Handle userdata,
-                                             realm_dart_on_collection_change_func_t on_change,
-                                             realm_scheduler_t *scheduler)
+realm_notification_token_t* realm_dart_results_add_notification_callback(
+    realm_results_t* results, 
+    Dart_Handle notification_controller,
```
nielsenko, line 88: `userdata` is a better name. This does not need to be a `NotificationController`  being parsed. Also, i think it should match the realm-core c-api function this is a front for.

src/realm_dart_collections.cpp
nielsenko, line 60: This is not safe. `m_handle` is really an `Dart_FinalizableHandle` which will be auto deleted. We do not know that at this point that it is not (Depends on the order the finalizers for the notification_token_t and m_handle).

Also, this depends on an implementation detail in the dart VM. Google can change this without even bumping the minor version, and still be compliant with semantic versioning.. not that I think they are highly likely to change this soon.

nielsenko, line 60: This is not safe. `m_handle` is really an `Dart_FinalizableHandle` which will be auto deleted. We do not know that at this point that it is not (Depends on the order the finalizers for the notification_token_t and m_handle).

Also, this depends on an implementation detail in the dart VM. Google can change this without even bumping the minor version, and still be compliant with semantic versioning.. not that I think they are highly likely to change this soon.

blagoev: his is safe, we are sure that this handle is persisted and is safe to be used here. Lets discuss this on our meeting today.

src/realm_dart_collections.cpp
blagoev: I changed the name here since this needs to be exactly notification_controller instance. The name is supposed to guide the caller. The C-API uses the name `userdata` since they don't know the context their API will be used hence a more generic term.

nielsenko (approved):

PR #253, opened 2022-02-11
throw error on primary key set

### LINKED ISSUE #254: Relax final restrictions on collections and throw error on primary key set
Remove `final` restriction on collections in models
throw an error if a primary key setter is called at runtime.

blagoev:
Generates throw error when setting primary keys
Relaxes final constraint on collection in models
Changes realm generator tests to run the dart tests as well
Run generator CI on Linux only
Refactors generated constructors
fixes #254


generator/lib/src/realm_field_info.dart
```diff
@@ -60,11 +61,13 @@ class RealmFieldInfo {
...
     yield "$typeName get $name => RealmObject.get<$basicTypeName>(this, '$realmName') as $typeName;";
-    if (!isFinal) {
+    bool generateSetter = !isFinal && !primaryKey && !isRealmCollection;
+    if (generateSetter) {
       yield '@override';
       yield "set $name(${typeName != typeModelName ? 'covariant ' : ''}$typeName value) => RealmObject.set(this, '$realmName', value);";
     } else {
-      if (isLate) { // we still need to override setter on late final, but just throw error!
+      bool generateThrowError = isLate || primaryKey || isRealmCollection;
+      if (generateThrowError) { // we still need to override setter on late final, but just throw error!
```
nirinchev, line 70: Have you folks looked into how migrations are going to be done? Because during a migration, the user should be able to change the primary key. This is how the PK setter looks like in C# to handle this use case: https://github.com/realm/realm-dotnet/blob/aba49e08af41066389f34cc18169976e2e1b7e05/Realm/Realm/DatabaseTypes/RealmObjectBase.cs#L223-L235

blagoev: no we have not reach that point. We will change the generator accordingly for that.
thanks for the heads up.

.github/workflows/realm-generator.yml
```diff
@@ -14,32 +14,46 @@ on:
...
       - name : Setup Dart SDK
         uses: dart-lang/setup-dart@main
         with:
-          sdk: ${{ matrix.channel }}
+          sdk: stable
 
-      - name: Install dependencies
+      - name: Install generator dependencies
         run: dart pub get
+        working-directory: ./generator/
```
desistefanova, line 35: Isn't it correct first to set working directory to `./generator/` and then to do `dart pub get`

.github/workflows/realm-generator.yml
```diff
@@ -14,32 +14,46 @@ on:
...
 
       - name: Run generator tests
-        run: dart test --reporter expanded --coverage ./coverage/
+        run: dart test --reporter expanded --coverage ./coverage/ --test-randomize-ordering-seed random
+        working-directory: ./generator/
```
desistefanova, line 39: This working directory is set twice. If it is because of the following step I suppose it should be set to realm-dart.

.github/workflows/realm-generator.yml
```diff
@@ -14,32 +14,46 @@ on:
...
+      - name: Setup Ninja
+        uses: seanmiddleditch/gha-setup-ninja@master
+
+      - name: Build Realm Dart for Linux
+        run: ./scripts/build-linux.sh
+
+      - name: Run realm_dart tests
+        run: dart test -r expanded --test-randomize-ordering-seed random
 
       - name: Generate generator coverage report
```
desistefanova, line 58: These steps about coverage reports are related to generator tests. Maybe it is better to be moved above right after running generator tests.

generator/lib/src/realm_model_info.dart
```diff
@@ -65,10 +65,7 @@ class RealmModelInfo {
         }
 
         yield* allExceptCollections.map((f) {
-          if (f.isFinal) {
-            return "RealmObject.set(this, '${f.name}', ${f.name});"; // since no setter will be created!
-          }
-          return 'this.${f.name} = ${f.name};'; // defer to generated setter
+          return "RealmObject.set(this, '${f.name}', ${f.name});";
```
desistefanova, line 68: Isn't it better to be like this for all the properties with setters `return 'this.${f.name} = ${f.name};'; // defer to generated setter`?

.github/workflows/realm-generator.yml
blagoev: I am not sure what you mean.  This does both. There is not first and second run.

.github/workflows/realm-generator.yml
blagoev: this is the working directory of this step. What do you mean it is set twice?

.github/workflows/realm-generator.yml
blagoev: The idea here is that we should generate coverage only after all tests have passed

generator/lib/src/realm_model_info.dart
blagoev: This change simplifies the code and is somewhat more performant. Why do you think it is better?

desistefanova: Just suggesting because the code will be more readable. If the other way is more performant, OK.

.github/workflows/realm-generator.yml
desistefanova: I understand. Order doesn't matter.

desistefanova (approved):

PR #265, opened 2022-02-20
line 38 causes error with CMAKE<3.19

domenicopopolizio:
COMMAND_ERROR_IS_FATAL, found on line 38, causes an error with CMAKE version < 3.19
See https://cmake.org/cmake/help/v3.18/command/execute_process.html compared to https://cmake.org/cmake/help/v3.19/command/execute_process.html
Since it has already been commented out on line 49, commenting it out on line 38 too could be an idea for solving the issue.
Alternatively the user should be asked to update CMAKE


domenicopopolizio:
@cla-bot check

blagoev:
@domenicopopolizio Thanks for reporting. We will bump the cmake minimum required version to 3.19 so there is a proper message to update cmake instead. 
This command should fail the build when it fails otherwise the build will fail later on with more obscure messages.

blagoev:
superseded by #266

PR #270, opened 2022-02-22
Allow unmanaged RealmList, and Use RealmList on interface of RealmObjects

nielsenko:


lib/src/list.dart
```diff
@@ -66,64 +73,73 @@ class RealmList<T extends Object> extends collection.ListBase<T> {
...
-  ///
-  /// The length of the list becomes zero.
-  /// If the elements are managed [RealmObject]s, they all remain in the Realm.
   @override
-  void clear() {
-    realmCore.listClear(this);
+  bool get isValid => realmCore.listIsValid(this);
+}
+
+class _UnmanagedRealmList<T extends Object> extends collection.ListBase<T> with RealmEntity implements RealmList<T> {
```
desistefanova, line 88: I remember we decided to remove underscore from _UnmanagedRealmList.

desistefanova (approved):

lib/src/list.dart
nielsenko: Yes.. apparently I only updated the `ManagedRealmList`. 2s.

nielsenko: Fixed

lib/src/realm_class.dart
```diff
@@ -33,6 +33,7 @@ import 'results.dart';
 export 'package:realm_common/realm_common.dart'
     show Ignored, Indexed, MapTo, PrimaryKey, RealmError, RealmModel, RealmUnsupportedSetError, RealmStateError, RealmCollectionType, RealmPropertyType;
 export "configuration.dart" show Configuration, RealmSchema, SchemaObject;
+export 'entity.dart' show RealmEntity;
```
blagoev, line 36: Could we place this class in `realm_object.dart` and remove `entity.dart`. Seems we have a chance here to keep the files number low.

lib/src/list.dart
```diff
@@ -66,64 +73,73 @@ class RealmList<T extends Object> extends collection.ListBase<T> {
...
-    final handle = realmCore.queryList(this, query, args);
+  RealmResults<T> query(String query, [List<Object> arguments = const []]) {
+    final handle = realmCore.queryList(this, query, arguments);
     return RealmResultsInternal.create<T>(handle, realm);
   }
 
   /// Allows listening for changes when the contents of this collection changes.
   Stream<RealmListChanges<T>> get changes {
-    final controller = ListNotificationsController<T>(this);
+    final controller = ListNotificationsController<T>(managed);
```
blagoev, line 133: This seems odd. Shouldn't this method throw a `RealmStateError` if the list is not managed. Then I think we don't need the `managed` getter below.

lib/src/list.dart
```diff
@@ -66,64 +73,73 @@ class RealmList<T extends Object> extends collection.ListBase<T> {
...
     return controller.createStream();
   }
 }
 
 /// @nodoc
-extension RealmListInternal on RealmList {
-  RealmListHandle get handle => _handle;
+extension RealmListInternal<T extends Object> on RealmList<T> {
+  ManagedRealmList<T> get managed => this as ManagedRealmList<T>;
```
blagoev, line 140: I suggest we remove this getter. It allows very easy to forget and throw a type error somewhere.

lib/src/entity.dart
```diff
@@ -0,0 +1,30 @@
...
+import 'realm_class.dart';
+
+mixin RealmEntity {
+  Realm? _realm;
+  Realm get realm => _realm ?? (throw RealmException('$this not managed'));
+  bool get isManaged => _realm != null;
+}
+
+extension RealmEntityInternal on RealmEntity {
+  void setRealm(Realm value) => _realm = value;
```
blagoev, line 28: As I suggest in a comment I suggest we move RealmEntity into `realm_object.dart`. Then we avoid the need to expose this method on all user types.

lib/src/list.dart
```diff
@@ -20,33 +20,40 @@ import 'dart:async';
...
 import 'collections.dart';
+import 'entity.dart';
 import 'native/realm_core.dart';
 import 'realm_class.dart';
-import 'realm_object.dart';
 import 'results.dart';
 
 /// Instances of this class are live collections and will update as new elements are either
-/// added to or deleted from the Realm that match the underlying query.
+/// added to or deleted to the list or from the Realm.
```
blagoev, line 29:
```suggestion
/// added to or deleted from the collection or from the Realm.
```

lib/src/list.dart
```diff
@@ -66,64 +73,73 @@ class RealmList<T extends Object> extends collection.ListBase<T> {
...
 
-  /// Sets the element at the specified index in the list.
   @override
   void operator []=(int index, T value) {
     RealmListInternal.setValue(handle, realm, index, value);
   }
 
-  /// Clears the collection in memory and the references
-  /// to the objects in this collection in Realm.
+  @override
```
blagoev, line 81: something is off with the doc comments. The doc comment on `clear` is deleted.

blagoev: oh ok, I remembered we talked about this.
In this case I believe we have to include the doc comment here since it has Realm specifics

lib/src/list.dart
```diff
@@ -20,33 +20,40 @@ import 'dart:async';
...
-  /// The length of this [RealmList].
   @override
-  int get length => realmCore.getListSize(handle);
+  int get length => realmCore.getListSize(_handle);
 
-  /// @nodoc
-  //settng lenght is no operation
   @override
-  set length(int length) {}
+  set length(int length) {} // no-op for managed lists
```
blagoev, line 55: I wonder if we should have this (`settng lenght on managed lists is a no operation`) in the doc comment as well. Then we will need the doc comment itself.

lib/src/list.dart
```diff
@@ -66,64 +73,73 @@ class RealmList<T extends Object> extends collection.ListBase<T> {
...
-  /// The length of the list becomes zero.
-  /// If the elements are managed [RealmObject]s, they all remain in the Realm.
   @override
-  void clear() {
-    realmCore.listClear(this);
+  bool get isValid => realmCore.listIsValid(this);
+}
+
+class UnmanagedRealmList<T extends Object> extends collection.ListBase<T> with RealmEntity implements RealmList<T> {
+  late final _unmanaged = <T?>[]; // lazy ctor'ed (use T? for length=)
```
blagoev, line 89:
```suggestion
  late final _unmanaged = <T?>[];
```

blagoev: I guess the comment is outdated, no?

lib/src/list.dart
blagoev: the same for query above

lib/src/list.dart
```diff
@@ -66,64 +73,73 @@ class RealmList<T extends Object> extends collection.ListBase<T> {
...
+  /// Only works for lists of [RealmObject]s.
   ///
   /// The Realm Dart and Realm Flutter SDKs supports querying based on a language inspired by [NSPredicate](https://academy.realm.io/posts/nspredicate-cheatsheet/)
   /// and [Predicate Programming Guide.](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Predicates/AdditionalChapters/Introduction.html#//apple_ref/doc/uid/TP40001789)
-  ///
-  /// Only works for lists of objects.
-  RealmResults<T> query(String query, [List<Object> args = const []]) {
-    final handle = realmCore.queryList(this, query, args);
+  RealmResults<T> query(String query, [List<Object> arguments = const []]) {
+    final handle = realmCore.queryList(this, query, arguments);
```
blagoev, line 127: should throw `RealmStateError` if the list is not managedlist

lib/src/list.dart
blagoev: also being lazy here brings no value since we initialize it every time in the ctor.

blagoev (changes requested):

lib/src/realm_class.dart
nielsenko: If we import `realm_object.dart` from `list.dart` then yes, but since it is just a file in `lib/src` (ie. not `lib`) I'm not sure I see the value in keeping the file count low? Anyway, I have prepped a commit, where `RealmEntity` is moved to `realm_object.dart`.

blagoev (commented):
Also please add changelog entry

lib/src/list.dart
nielsenko: Actually we don't if the optional `items` is null, but it is a very minor optimisation.

nielsenko:
> Also please add changelog entry
I was wondering about this. This is mostly invisible to the user in that we didn't have any public methods on `RealmList` over `List` in the previous release (since we didn't have `changes`, `isManaged`, etc.). But I'll add it

lib/src/list.dart
nielsenko: I see your point. I suggest making it private and renaming it to `_ensureManaged`, and have it throw a `RealmStateError` instead. The same has to happen a few places.

lib/src/list.dart
nielsenko: Fixed

lib/src/list.dart
nielsenko: I have gone about it slightly different. See suggestion

lib/src/list.dart
nielsenko: Does it have any realm specific behaviour? The fact that the items are not deleted just because they are removed from the list, is no different than a regular list.. Or am I missing something?

lib/src/list.dart
```diff
@@ -20,33 +20,40 @@ import 'dart:async';
...
+  bool get isValid;
+
+  factory RealmList._(RealmListHandle handle, Realm realm) => ManagedRealmList._(handle, realm);
+  factory RealmList(Iterable<T> items) => UnmanagedRealmList(items);
+}
+
+class ManagedRealmList<T extends Object> extends collection.ListBase<T> with RealmEntity implements RealmList<T> {
   final RealmListHandle _handle;
-  
-  /// The Realm isntance this collection belongs to.
```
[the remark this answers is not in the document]
nielsenko: Moved to `RealmEntity`

lib/src/entity.dart
nielsenko: It is needed in `list.dart` as well. I use the common pattern of create a public class that is hidden outside package.

lib/src/list.dart
nielsenko: Fixed

lib/src/list.dart
nielsenko: Included in latest commit

lib/src/list.dart
nielsenko: The fact we use `T?` not T, is a requirement to use `ListBase`.

lib/src/list.dart
nielsenko: Fixed

lib/src/list.dart
nielsenko: dropped `late` and changed comment

lib/src/realm_object.dart
```diff
@@ -272,7 +276,7 @@ extension RealmObjectInternal on RealmObject {
     }
 
     _handle = handle;
-    _realm = realm;
+    setRealm(realm);
```
blagoev, line 279: we can set the `_realm` directly now

lib/src/realm_object.dart
```diff
@@ -289,15 +293,12 @@ extension RealmObjectInternal on RealmObject {
     final object = RealmObject._factories[type]!();
     object._handle = handle;
     object._accessor = accessor;
-    object._realm = realm;
+    object.setRealm(realm);
```
blagoev, line 296: as above

lib/src/realm_object.dart
```diff
@@ -327,7 +328,7 @@ class RealmObjectChanges<T extends RealmObject> {
   /// The property names that have changed.
   List<String> get properties {
     final propertyKeys = realmCore.getObjectChangesProperties(_handle);
-    return object._realm!.getPropertyNames(object.runtimeType, propertyKeys);
+    return object.realm.getPropertyNames(object.runtimeType, propertyKeys);
```
blagoev, line 331: Now we can access the `_realm`, no need to go through the `realm` getter hop

lib/src/realm_object.dart
```diff
@@ -342,7 +343,7 @@ class RealmObjectNotificationsController<T extends RealmObject> extends Notifica
 
   @override
   RealmNotificationTokenHandle subscribe() {
-    return realmCore.subscribeObjectNotifications(realmObject._handle!, this, realmObject._realm!.scheduler.handle);
+    return realmCore.subscribeObjectNotifications(realmObject._handle!, this, realmObject.realm.scheduler.handle);
```
blagoev, line 346: as above

lib/src/realm_object.dart
```diff
@@ -178,6 +178,18 @@ class RealmCoreAccessor implements RealmAccessor {
   }
 }
 
+mixin RealmEntity {
+  Realm? _realm;
+  /// The [Realm] instance this collection belongs to.
```
blagoev, line 183: realm entity is not a collection. The doc comment needs a fix. Perhaps
`The [Realm] instance this object belongs to`

blagoev (changes requested):

lib/src/realm_object.dart
```diff
@@ -178,6 +178,18 @@ class RealmCoreAccessor implements RealmAccessor {
...
+  Realm get realm => _realm ?? (throw RealmStateError('$this not managed'));
+  /// True if the entity belongs to a realm.
+  bool get isManaged => _realm != null;
+}
+
+extension RealmEntityInternal on RealmEntity {
```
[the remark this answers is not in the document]
nielsenko: We need it for `list.dart`

lib/src/realm_object.dart
nielsenko: Fixed

lib/src/realm_object.dart
nielsenko: Erhh.. do you prefer `_realm!.`over `realm`?

lib/src/realm_object.dart
nielsenko: Same

lib/src/list.dart
```diff
@@ -66,64 +75,73 @@ class RealmList<T extends Object> extends collection.ListBase<T> {
...
-    final handle = realmCore.queryList(this, query, args);
+  RealmResults<T> query(String query, [List<Object> arguments = const []]) {
+    final handle = realmCore.queryList(_ensureManaged, query, arguments);
     return RealmResultsInternal.create<T>(handle, realm);
   }
 
   /// Allows listening for changes when the contents of this collection changes.
   Stream<RealmListChanges<T>> get changes {
-    final controller = ListNotificationsController<T>(this);
+    final controller = ListNotificationsController<T>(_ensureManaged);
```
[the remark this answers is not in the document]
nielsenko: Okay

lib/src/realm_object.dart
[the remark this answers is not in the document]
blagoev: I see.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * Added support for opening Realm in read-only mode. ([#260](https://github.com/realm/realm-dart/pull/260))
 * Primary key fields no longer required to be `final` in data model classes ([#240](https://github.com/realm/realm-dart/pull/240))
 * List fields no longer required to be `final` in data model classes. ([#253](https://github.com/realm/realm-dart/pull/253))
+* Generated classes no longer use `List` class on interface, but instead use `RealmList`. This allow user to avoid casting, when accessing `RealmList` specific functionality. ([#270](https://github.com/realm/realm-dart/pull/270))
```
blagoev, line 15: I am sorry, should have said it clearer. I suggest we have this something like this

### Internal
 * Use `RealmList` instead of `List` in generated Realm object classes. It enables access to a Realm list specific APIs.

lib/src/realm_object.dart
blagoev: I don't mind either. I wonder about performance, but maybe its negligible. Lets leave it like that

lib/src/realm_object.dart
```diff
@@ -358,7 +359,7 @@ class RealmObjectNotificationsController<T extends RealmObject> extends Notifica
 
     final changes = RealmObjectChanges<T>._(changesHandle, realmObject);
     streamController.add(changes);
-  }
+  }<
```
blagoev, line 362: a weird symbol

nielsenko: Yes.. it broke the build. Fixed

lib/src/list.dart
nielsenko: now called `asManaged()`

CHANGELOG.md

blagoev (approved):

PR #274, opened 2022-02-23
add explanation for using _ and $ prefixes

swifthing:


swifthing:
@cla-bot check

blagoev:
@swifthing Thanks for the suggestion. We will close this issue since we want to keep our getting started guide concise and to the point. This will be documented in our Docs pages.
Also as a best practice our recommendation is to not use $ prefix like `$MyModel` unless really necessary.

PR #280, opened 2022-02-28
config inMemory

### LINKED ISSUE #91: Config: Support inMemory Realm

desistefanova:
Implemented inMemory property of realm configuration. Fixes #91

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * Added support checking if Realm lists and Realm objects are valid. ([#183](https://github.com/realm/realm-dart/pull/183))
 * Support query on lists of realm objects. ([#239](https://github.com/realm/realm-dart/pull/239))
 * Added support for opening Realm in read-only mode. ([#260](https://github.com/realm/realm-dart/pull/260))
-* Added support for opening Realm in memory. ([#280](https://github.com/realm/realm-dart/pull/280))
+* Added support for opening in-memory Realms. ([#280](https://github.com/realm/realm-dart/pull/280))
 * Primary key fields no longer required to be `final` in data model classes ([#240](https://github.com/realm/realm-dart/pull/240))
 * List fields no longer required to be `final` in data model classes. ([#253](https://github.com/realm/realm-dart/pull/253))
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * Added support checking if Realm lists and Realm objects are valid. ([#183](https://github.com/realm/realm-dart/pull/183))
 * Support query on lists of realm objects. ([#239](https://github.com/realm/realm-dart/pull/239))
 * Added support for opening Realm in read-only mode. ([#260](https://github.com/realm/realm-dart/pull/260))
+* Added support for opening Realm in memory. ([#280](https://github.com/realm/realm-dart/pull/280))
 * Primary key fields no longer required to be `final` in data model classes ([#240](https://github.com/realm/realm-dart/pull/240))
 * List fields no longer required to be `final` in data model classes. ([#253](https://github.com/realm/realm-dart/pull/253))
 
--- a/lib/src/configuration.dart
+++ b/lib/src/configuration.dart
@@ -40,7 +40,11 @@ class Configuration {
   /// This allows opening it from locked locations such as resources,
   /// bundled with an application.  The realm file must already exists.
   ///
-  /// If [inMemoryOnly] is set, then the [Realm] instance will not be persisted, and hence be _deleted_ after it is closed.
+  /// [inMemory] specifies if this Realm should be opened in-memory.
+  /// This still requires a path (can be the default path) to identify the [Realm] so other processes can open the same [Realm].
+  /// The file will also be used as swap space if the [Realm] becomes bigger than what fits in memory,
+  /// but it is not persistent and will be removed when the last instance is closed.
+  /// When all in-memory instance of [Realm] is closed all data in that [Realm] is deleted.
   Configuration(List<SchemaObject> schemaObjects, {bool readOnly = false, bool inMemory = false})
       : _schema = RealmSchema(schemaObjects),
         _handle = realmCore.createConfig() {
@@ -49,8 +53,8 @@ class Configuration {
     if (readOnly) {
       isReadOnly = true;
     }
-    if (inMemoryOnly) {
-      inMemory = true;
+    if (inMemory) {
+      isInMemory = true;
     }
     realmCore.setSchema(this);
   }
@@ -103,10 +107,14 @@ class Configuration {
   bool get isReadOnly => realmCore.getConfigReadOnly(this);
   set isReadOnly(bool value) => realmCore.setConfigReadOnly(this, value);
 
-  /// Specifies if this Realm should be opened in-memory. This still requires a path (can be the default path) to identify the Realm so other processes can open the same Realm. The file will also be used as swap space if the Realm becomes bigger than what fits in memory, but it is not persistent and will be removed when the last instance is closed.
+  /// Specifies if this Realm should be opened in-memory.
+  ///
+  /// This still requires a path (can be the default path) to identify the Realm so other processes can open the same Realm.
+  /// The file will also be used as swap space if the Realm becomes bigger than what fits in memory,
+  /// but it is not persistent and will be removed when the last instance is closed.
   /// When all in-memory instance of Realm is closed all data in that Realm is deleted.
-  bool get inMemory => realmCore.getConfigInMemory(this);
-  set inMemory(bool value) => realmCore.setConfigInMemory(this, value);
+  bool get isInMemory => realmCore.getConfigInMemory(this);
+  set isInMemory(bool value) => realmCore.setConfigInMemory(this, value);
 }
 
 /// A collection of properties describing the underlying schema of a [RealmObject].
--- a/lib/src/configuration.dart
+++ b/lib/src/configuration.dart
@@ -38,7 +38,7 @@ class Configuration {
   /// Creates a [Configuration] with schema objects for opening a [Realm].
   /// [readOnly] controls whether a [Realm] is opened as read-only.
   ///
-  /// [inMemory] specifies if this Realm should be opened in-memory.
+  /// [inMemory] specifies if this [Realm] should be opened in-memory.
   Configuration(List<SchemaObject> schemaObjects, {bool readOnly = false, bool inMemory = false})
       : _schema = RealmSchema(schemaObjects),
         _handle = realmCore.createConfig() {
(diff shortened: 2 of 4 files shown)


lib/src/configuration.dart
```diff
@@ -97,6 +100,11 @@ class Configuration {
   /// The realm file must already exists at [path]
   bool get isReadOnly => realmCore.getConfigReadOnly(this);
   set isReadOnly(bool value) => realmCore.setConfigReadOnly(this, value);
+
+  /// Gets or sets a value specifying settings for an in-memory Realm.
+  /// When all in-memory instance of Realm is closed all data in that Realm is deleted.
+  bool get inMemory => realmCore.getConfigInMemory(this);
+  set inMemory(bool value) => realmCore.setConfigInMemory(this, value);
```
nielsenko, line 107: Do we want this to be public?

lib/src/configuration.dart
```diff
@@ -97,6 +100,11 @@ class Configuration {
   /// The realm file must already exists at [path]
   bool get isReadOnly => realmCore.getConfigReadOnly(this);
   set isReadOnly(bool value) => realmCore.setConfigReadOnly(this, value);
```
nielsenko, line 102: Do we want this to be public?

nielsenko (changes requested):
I would prefer `Configuration` to be immutable, ie. to drop the public setters, so that everything must be set on construction. We hold a reference to the `Configuration` from a `Realm`. You cannot really manipulate the `Configuration` after you have opened the `Realm`, so why allow stuff like `realm.Configuration.inMemory = true` in the type system (even if it does throw an exception).

nielsenko:
> I would prefer `Configuration` to be immutable, ie. to drop the public setters, so that everything must be set on construction. We hold a reference to the `Configuration` from a `Realm`. You cannot really manipulate the `Configuration` after you have opened the `Realm`, so why allow stuff like `realm.Configuration.inMemory = true` in the type system (even if it does throw an exception).

On round-table today, we decided to merge as is, and defer making `Configuration` immutable to later

nielsenko (approved):

lib/src/configuration.dart
desistefanova: Discussed on a meeting. Leaving it public.

lib/src/configuration.dart
```diff
@@ -39,14 +39,19 @@ class Configuration {
   /// [readOnly] controls whether a [Realm] is opened as readonly.
   /// This allows opening it from locked locations such as resources,
   /// bundled with an application.  The realm file must already exists.
-  Configuration(List<SchemaObject> schemaObjects, {bool readOnly = false})
+  ///
+  /// [inMemoryOnly] configures [Realm] instance to be deleted after it is closed.
```
nielsenko, line 43:
```suggestion
  /// If [inMemoryOnly] is set, then the [Realm] instance will not be persisted, and hence deleted after it is closed.
```

nielsenko: .. but maybe I'm wrong and it is actually persisted, and just gets deleted when closed? I would be surprised, but then again it wouldn't be the first time 😊

nielsenko: Okay 😄

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * Added support checking if Realm lists and Realm objects are valid. ([#183](https://github.com/realm/realm-dart/pull/183))
 * Support query on lists of realm objects. ([#239](https://github.com/realm/realm-dart/pull/239))
 * Added support for opening Realm in read-only mode. ([#260](https://github.com/realm/realm-dart/pull/260))
+* Added support for opening Realm in memory. ([#280](https://github.com/realm/realm-dart/pull/280))
```
blagoev, line 13:
```suggestion
* Added support for opening in-memory Realms. ([#280](https://github.com/realm/realm-dart/pull/280))
```

lib/src/configuration.dart
```diff
@@ -39,14 +39,19 @@ class Configuration {
   /// [readOnly] controls whether a [Realm] is opened as readonly.
   /// This allows opening it from locked locations such as resources,
   /// bundled with an application.  The realm file must already exists.
-  Configuration(List<SchemaObject> schemaObjects, {bool readOnly = false})
+  ///
+  /// If [inMemoryOnly] is set, then the [Realm] instance will not be persisted, and hence be _deleted_ after it is closed.
```
blagoev, line 43: I think other SDKs have different doc here.

lib/src/configuration.dart
```diff
@@ -39,14 +39,19 @@ class Configuration {
...
+  Configuration(List<SchemaObject> schemaObjects, {bool readOnly = false, bool inMemoryOnly = false})
```
blagoev, line 44:
```suggestion
  Configuration(List<SchemaObject> schemaObjects, {bool readOnly = false, bool inMemory = false})
```

lib/src/configuration.dart
```diff
@@ -97,6 +102,11 @@ class Configuration {
   /// The realm file must already exists at [path]
   bool get isReadOnly => realmCore.getConfigReadOnly(this);
   set isReadOnly(bool value) => realmCore.setConfigReadOnly(this, value);
+
+  /// Gets or sets a value specifying settings for an in-memory Realm.
+  /// When all in-memory instance of Realm is closed all data in that Realm is deleted.
+  bool get inMemory => realmCore.getConfigInMemory(this);
+  set inMemory(bool value) => realmCore.setConfigInMemory(this, value);
```
blagoev, line 109: This name diverges with `isReadOnly` we could syncing it. So `isInMemory` or we rename `isReadOnly` to `readOnly`

lib/src/configuration.dart
```diff
@@ -97,6 +102,11 @@ class Configuration {
   /// The realm file must already exists at [path]
   bool get isReadOnly => realmCore.getConfigReadOnly(this);
   set isReadOnly(bool value) => realmCore.setConfigReadOnly(this, value);
+
+  /// Gets or sets a value specifying settings for an in-memory Realm.
+  /// When all in-memory instance of Realm is closed all data in that Realm is deleted.
```
blagoev, line 107: Other SDKs have different doc comments here.

lib/src/configuration.dart
desistefanova: Could you please check this comment: https://github.com/realm/realm-dart/pull/280#r817680594

lib/src/configuration.dart
```diff
@@ -97,6 +102,11 @@ class Configuration {
   /// The realm file must already exists at [path]
   bool get isReadOnly => realmCore.getConfigReadOnly(this);
   set isReadOnly(bool value) => realmCore.setConfigReadOnly(this, value);
+
+  /// Gets or sets a value specifying settings for an in-memory Realm.
```
blagoev, line 106:
```suggestion
  /// Specifies if this Realm should be opened in-memory
```

blagoev:
```suggestion
  /// Specifies if this Realm should be opened in-memory. This still requires a path (can be the default path) to identify the Realm so other processes can open the same Realm. The file will also be used as swap space if the Realm becomes bigger than what fits in memory, but it is not persistent and will be removed when the last instance is closed.
```

test/configuration_test.dart
```diff
@@ -108,4 +108,58 @@ Future<void> main([List<String>? args]) async {
...
+
+  test('Configuration inMemory - no files after closing realm', () {
+    Configuration config = Configuration([Car.schema], inMemoryOnly: true);
+    var realm = Realm(config);
+    realm.write(() => realm.add(Car('Tesla')));
+    realm.close();
+    expect(Realm.existsSync(config.path), false);
+  });
+
+  test('Configuration inMemory - lost objects after closing', () {
```
blagoev, line 120: I think we could delete this test. Does not make much sense to test that if it is inMemory only.

test/configuration_test.dart
```diff
@@ -108,4 +108,58 @@ Future<void> main([List<String>? args]) async {
...
+  test('Configuration inMemory can not be readOnly', () {
+    Configuration config = Configuration([Car.schema], inMemoryOnly: true);
+    var realm = Realm(config);
+
+    config.isReadOnly = true;
+    expect(() => Realm(config), throws<RealmException>("Realm at path '${config.path}' already opened with different read permissions"));
+    realm.close();
+  });
+
+  test('Configuration inMemory and readOnly', () {
```
blagoev, line 141: This test is testing readonly instead of inMemory. I suggest delete it.

lib/src/configuration.dart
blagoev: yes it should be public

lib/src/configuration.dart
blagoev: add a review with a suggestion

desistefanova: Done.

lib/src/configuration.dart
desistefanova: Renamed to isInMemory .

lib/src/configuration.dart
desistefanova: Done.

test/configuration_test.dart
desistefanova: Deleted.

test/configuration_test.dart
desistefanova: Deleted.

lib/src/configuration.dart
```diff
@@ -39,14 +39,23 @@ class Configuration {
   /// [readOnly] controls whether a [Realm] is opened as readonly.
   /// This allows opening it from locked locations such as resources,
   /// bundled with an application.  The realm file must already exists.
-  Configuration(List<SchemaObject> schemaObjects, {bool readOnly = false})
+  ///
+  /// [inMemory] specifies if this Realm should be opened in-memory.
+  /// This still requires a path (can be the default path) to identify the [Realm] so other processes can open the same [Realm].
```
blagoev, line 44: Since we document this below I suggest we keep it short here. as the fifo api doc. So to remove these comments

lib/src/configuration.dart
```diff
@@ -39,14 +39,23 @@ class Configuration {
   /// [readOnly] controls whether a [Realm] is opened as readonly.
   /// This allows opening it from locked locations such as resources,
   /// bundled with an application.  The realm file must already exists.
-  Configuration(List<SchemaObject> schemaObjects, {bool readOnly = false})
+  ///
+  /// [inMemory] specifies if this Realm should be opened in-memory.
```
blagoev, line 43: Only this one to stay
/// [inMemory] specifies if this Realm should be opened in-memory.

blagoev (changes requested):

lib/src/configuration.dart
desistefanova: done

lib/src/configuration.dart
desistefanova: done

blagoev (approved):

PR #283, opened 2022-03-02
Implement Realm.deleteAll. Fixes #59

### LINKED ISSUE #59: Implement deleteAll by type
Deletes all RealmObjects of given type

nielsenko:


desistefanova (approved):

CHANGELOG.md
```diff
@@ -7,11 +7,12 @@ x.x.x Release notes (yyyy-MM-dd)
 * Support change notifications on query results. ([#208](https://github.com/realm/realm-dart/pull/208))
 * Support change notifications on list collections. ([#261](https://github.com/realm/realm-dart/pull/261))
 * Support change notifications on realm objects. ([#262](https://github.com/realm/realm-dart/pull/262))
-* Added support checking if Realm lists and Realm objects are valid. ([#183](https://github.com/realm/realm-dart/pull/183))
+* Added support for checking if realm lists and realm objects are valid. ([#183](https://github.com/realm/realm-dart/pull/183))
```
blagoev, line 10: Isn't it better to have it as capital letter?

lib/src/realm_class.dart
```diff
@@ -232,6 +232,9 @@ class Realm {
     final handle = realmCore.queryClass(this, metadata.class_.key, query, args);
     return RealmResultsInternal.create<T>(handle, this);
   }
+
+  /// Deletes all [RealmObject]s of type `T` in the `Realm`
```
blagoev, line 236:
```suggestion
  /// Deletes all [RealmObject]s of type `T` from the `Realm`
```

CHANGELOG.md
nielsenko: I just think we should be consistent. For change notifications we used lower case.

blagoev: I agree, we should try to be consistent. Other SDKs are all over the place.  But I wonder if its better to use capital letter for these. I lean towards using capital letter. `Realm object, Realm list` vs `realm list, realm object` etc. What do you think?

blagoev (approved):

PR #284, opened 2022-03-02
Implemented setting fifoFilesFallbackPath in configuration

### LINKED ISSUE #93: Config: Support custom FIFO files paths

desistefanova:
Support custom FIFO files paths. Fixes #93


lib/src/configuration.dart
```diff
@@ -36,14 +36,23 @@ class Configuration {
...
       : _schema = RealmSchema(schemaObjects),
         _handle = realmCore.createConfig() {
     schemaVersion = 0;
     path = defaultPath;
+
+    if (fifoFilesFallbackPath?.isNotEmpty ?? false) {
+      this.fifoFilesFallbackPath = fifoFilesFallbackPath!;
+    }
+
     if (readOnly) {
```
nielsenko, line 56: Not part of this PR, but I noticed this check. Should we not just set property regardless?

lib/src/configuration.dart
```diff
@@ -97,6 +106,16 @@ class Configuration {
   /// The realm file must already exists at [path]
   bool get isReadOnly => realmCore.getConfigReadOnly(this);
   set isReadOnly(bool value) => realmCore.setConfigReadOnly(this, value);
+
+  /// Gets or sets a value of FIFO special files location.
```
nielsenko, line 110: I think this comment will only show up on the getter

lib/src/configuration.dart
desistefanova: FIFO path is empty by default in C-API. It doesn't need to be set. This is some kind as a small optimization.
If you mean ReadOnly, it is a mode and the value is enum value. I would like to keep the default enum value of schema mode, which is currently "Automatic". But you are right it is not visible here, you can see it inside _realmLib.realm_config_set_in_memory.

lib/src/configuration.dart
```diff
@@ -97,6 +106,16 @@ class Configuration {
...
+  /// Opening a Realm creates a number of FIFO special files in order to
+  /// coordinate access to the Realm across threads and processes. If the Realm file is stored in a location
+  /// that does not allow the creation of FIFO special files (e.g. FAT32 filesystems), then the Realm cannot be opened.
+  /// In that case Realm needs a different location to store these files and this property defines that location.
+  /// The FIFO special files are very lightweight and the main Realm file will still be stored in the location defined
+  /// by the `path` property. This property is ignored if the directory defined by `path` allow FIFO special files.
```
nielsenko, line 116:
```suggestion
  /// by the [path] you  property. This property is ignored if the directory defined by [path] allow FIFO special files.
```

lib/src/configuration.dart
nielsenko: They appear on the same page, so not really an issue

lib/src/configuration.dart
```diff
@@ -36,14 +36,23 @@ class Configuration {
...
     schemaVersion = 0;
     path = defaultPath;
+
+    if (fifoFilesFallbackPath?.isNotEmpty ?? false) {
+      this.fifoFilesFallbackPath = fifoFilesFallbackPath!;
+    }
+
     if (readOnly) {
       isReadOnly = true;
     }
```
nielsenko, line 58:
```suggestion
    isReadOnly = true;
```

nielsenko: .. was all I meant

lib/src/configuration.dart
nielsenko: Okay. Should we have an enum here as well then? (eventually)

nielsenko (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * Added support for opening Realm in read-only mode. ([#260](https://github.com/realm/realm-dart/pull/260))
 * Primary key fields no longer required to be `final` in data model classes ([#240](https://github.com/realm/realm-dart/pull/240))
 * List fields no longer required to be `final` in data model classes. ([#253](https://github.com/realm/realm-dart/pull/253))
+* Support custom FIFO sperial files paths. ([#284](https://github.com/realm/realm-dart/pull/284))
```
blagoev, line 15:
```suggestion
* Support custom FIFO special files. ([#284](https://github.com/realm/realm-dart/pull/284))
```

lib/src/configuration.dart
```diff
@@ -36,14 +36,23 @@ class Configuration {
   RealmSchema get schema => _schema;
 
   /// Creates a [Configuration] with schema objects for opening a [Realm].
+  ///
+  /// [fifoFilesFallbackPath] is a custom directory for storing FIFO special files
```
blagoev, line 40: JS has this doc API doc https://docs.mongodb.com/realm-sdks/js/latest/Realm.html#~Configuration

So something like: 
```suggestion
/// [fifoFilesFallbackPath] enables FIFO special files.
/// Opening a Realm creates a number of FIFO special files in order to coordinate access to the Realm across threads and processes. If the Realm file is stored in a location that does not allow the creation of FIFO special files (e.g. FAT32 filesystems), then the Realm cannot be opened. The FIFO special files are very lightweight and the main Realm file will still be stored in the location defined by the path property. Ignored if the filesystem supports FIFO special files. 
/// 
```

lib/src/configuration.dart
```diff
@@ -36,14 +36,23 @@ class Configuration {
...
   /// This allows opening it from locked locations such as resources,
   /// bundled with an application.  The realm file must already exists.
-  Configuration(List<SchemaObject> schemaObjects, {bool readOnly = false})
+  Configuration(List<SchemaObject> schemaObjects, {String? fifoFilesFallbackPath, bool readOnly = false})
       : _schema = RealmSchema(schemaObjects),
         _handle = realmCore.createConfig() {
     schemaVersion = 0;
     path = defaultPath;
+
+    if (fifoFilesFallbackPath?.isNotEmpty ?? false) {
```
blagoev, line 52: could check for null only instead. if you really want to check for empty string then we need to throw a meaningful exception. I suggest checking for null only here

lib/src/configuration.dart
blagoev: there is sense in doing that. The least we call core the better. I think this is should be the case for all optional properties.

lib/src/configuration.dart
blagoev: OK I saw that this is documented below. So lets do this one instead.

blagoev:
```suggestion
  /// [fifoFilesFallbackPath] enables FIFO special files. 
```

blagoev (changes requested):

lib/src/configuration.dart
desistefanova: Done.

lib/src/configuration.dart
```diff
@@ -36,14 +36,23 @@ class Configuration {
   RealmSchema get schema => _schema;
 
   /// Creates a [Configuration] with schema objects for opening a [Realm].
+  ///
+  /// [fifoFilesFallbackPath] enables FIFO special files.
+  /// in case the Realm file is in a location that does not allow the creation of FIFO special files.
```
blagoev, line 41:
```suggestion
```

blagoev (approved):

PR #286, opened 2022-03-02
results - requested wrong index

### LINKED ISSUE #285: Add a test for results wrong index error

desistefanova:
New test added for getting element from results by not existing index. Fixes #285 
Fails for Mac OS. Tests [issue](https://github.com/realm/realm-core/issues/5233).


test/results_test.dart
```diff
@@ -138,6 +138,21 @@ Future<void> main([List<String>? args]) async {
...
+    var config = Configuration([Car.schema]);
+    var realm = Realm(config);
+
+    final car = Car('');
+    realm.write(() => realm.add(car));
+
+    final cars = realm.all<Car>();
+    realm.write(() => realm.deleteMany(cars));
+
+    expect(() => cars[0], throws<RealmException>("Requested index 0 in empty Results"));
```
blagoev, line 151: Why not just test this after opening the Realm since its empty we could skip adding objects to the realm

blagoev (changes requested):

test/results_test.dart
desistefanova: Done

blagoev (approved):

PR #289, opened 2022-03-09
Add build directories to .gitignore file

awh6al:
Generated with `./scripts/build-android.sh all`, So we should add them.


blagoev:
Thanks for reporting this. 
I have done this PR #291 that fixes the android build directory location, which was the initial issue. Now all these platform build directories are already ignored in `build-android` and should not pollute the main repo dir.

cheers

PR #294, opened 2022-03-14
return from write

### LINKED ISSUE #67: Support result value from write transaction callbacks
A single value should be possible to be returned as a result of the write transaction callback.
var value = realm.write(() {})

desistefanova:
Return result from write method callback. Fixes #67


nielsenko (approved):

lib/src/realm_class.dart
```diff
@@ -166,11 +166,12 @@ class Realm {
   ///
   /// If no exception is thrown from within the callback, the transaction will be committed.
   /// It is more efficient to update several properties or even create multiple objects in a single write transaction.
-  void write(void Function() writeCallback) {
+  TResult write<TResult>(TResult Function() writeCallback) {
```
blagoev, line 169: I suggest we use `T` here.

blagoev, line 169:
```suggestion
  T write<T>(T Function() writeCallback) {
```

test/realm_test.dart
```diff
@@ -595,4 +595,18 @@ Future<void> main([List<String>? args]) async {
     expect(mainSchools[0].branches[0].students.length + mainSchools[0].branches[1].students.length, 3);
     realm.close();
   });
+
+  test('Realm write returns added object', () {
```
blagoev, line 599:
```suggestion
  test('Realm write returns result', () {
```

blagoev (changes requested):

lib/src/realm_class.dart
desistefanova: done

blagoev (approved):

PR #301, opened 2022-03-15
generator test input/output files

desistefanova:
Refactor generator tests with input/output files.
goot_test.dart - is for tests that generate realm object classes successfully;
error_test.dart - is for all the tests that throw errors;
info_test.dart - is for tests that don't throw errors, but write some info in the log;
generator_test.dart is split between error_test.dart, info_test.dart and goot_test.dart.


generator/test/test_util.dart
```diff
@@ -0,0 +1,64 @@
...
+import 'package:path/path.dart' as _path;
+import 'package:dart_style/dart_style.dart';
+import 'package:build/build.dart';
+import 'package:build_test/build_test.dart';
+import 'package:realm_generator/realm_generator.dart';
+
+String _stringReplacements(String content) {
+  final formatter = DartFormatter();
+  var lines = LineSplitter.split(content);
+  String formattedContent = lines.where((element) => !element.startsWith("part of")).join('\n');
```
desistefanova, line 13: @nielsenko  do you it would be correct to generate also `part of ` ? Currently I remove programmatically this line from expected file output in order to match with the actual result.

nielsenko: As you may have noticed the `part of` is declared when you run the generator. It is just done by the `source_gen|combining_builder` that is setup in `build.yaml`. This allows us to co-exists with other builders that output to the same file.

generator/test/test_util.dart
```diff
@@ -0,0 +1,64 @@
...
+}
+
+Future<Map<String, Object>> getOutputFileAsset(String path) async {
+  String output = await readFileAsFormattedString(path);
+  String fileNameWithoutExtensions = _path.basenameWithoutExtension(_path.basenameWithoutExtension(path));
+  var generatedFile = '${_path.dirname(path)}/$fileNameWithoutExtensions.realm_objects.g.part';
+  return {'pkg|$generatedFile': output};
+}
+
+Future<dynamic> ioTtestBuilder(String directoryName, String testFilesPreffix) async {
```
nielsenko, line 56: Maybe just have a `path` argument and switch the extension with `setExtension` (https://pub.dev/documentation/path/latest/path/setExtension.html). Could consider building the path from the test _name_ and _group_?

generator/test/test_util.dart
```diff
@@ -0,0 +1,64 @@
...
+  return testBuilder(
+    generateRealmObjects(),
+    await getInputFileAsset('test/$directoryName/$testFilesPreffix.dart'),
+    outputs: await getOutputFileAsset('test/$directoryName/$testFilesPreffix.g.dart'),
+    reader: await PackageAssetReader.currentIsolate(),
+    writer: _OutputFileWriter(),
```
nielsenko, line 62: I have a vague recollection that setting `writer` changes how `outputs` are compared, but maybe I' mistaken.

desistefanova: The writer is no more needed.

generator/test/test_util.dart
desistefanova: Thanks!

generator/test/test_util.dart
desistefanova: OK

generator/test/test_util.dart
desistefanova: I implemented resolving test names by existing files in the group folder.

blagoev (approved):

PR #306, opened 2022-03-23
Added ApplicationConfiguration

### LINKED ISSUE #401: App: support basic properties
Support App basic properties base url, local app name, local app version, default timeout, platform, platform version, sdk version

nielsenko:
Implement ApplicationConfiguration basic properties like: base url, local app name, local app version, default timeout, platform, platform version, sdk version. Fixes #401


nielsenko:
Will rebase on master once current base is merged

desistefanova (approved):

nirinchev (commented):
The PR looks fine, but I'm not sure I understand why we're only implementing the configuration halfway? As far as I can tell, we're missing the 3 metadata properties - base path, encryption key, and persistence mode. Those are just primitive types, so unless there's a strong argument against, I'd suggest we just add them.

nielsenko:
> The PR looks fine, but I'm not sure I understand why we're only implementing the configuration halfway? As far as I can tell, we're missing the 3 metadata properties - base path, encryption key, and persistence mode. Those are just primitive types, so unless there's a strong argument against, I'd suggest we just add them.

As is, I only included the properties that are exposed with `realm_app_config_set_*` functions in `realm.h`. The properties you refer to, are part of the sync configuration, ie. exposed with `realm_sync_client_config_set_*`.

I like the idea of combining them into `ApplicationConfiguration` on the dart side, but perhaps we should discuss that in a wider group?

nirinchev:
To create an app we need both an app config and sync client config. Even though they are split in the C API there's no reason to split them in the SDK API and I believe every SDK exposes them on the app config object.

nielsenko:
> To create an app we need both an app config and sync client config. Even though they are split in the C API there's no reason to split them in the SDK API and I believe every SDK exposes them on the app config object.

Updated PR as suggested

lib/src/application_configuration.dart
```diff
@@ -0,0 +1,79 @@
...
+////////////////////////////////////////////////////////////////////////////////
+import 'dart:io';
+
+import 'package:meta/meta.dart';
+import 'package:pub_semver/pub_semver.dart';
+
+enum MetadataPersistenceMode {
+  plainText,
+  encrypted,
+}
```
nirinchev, line 26: Missing docs

lib/src/application_configuration.dart
```diff
@@ -0,0 +1,79 @@
...
+
+@immutable
+class ApplicationConfiguration {
+  /// The [appId] is the unique id that identifies the Realm application.
+  final String appId;
+
+  /// The [baseFilePath] is the [Directory] relative to which all local data for this application will be stored.
+  /// This data includes metadata for users and synchronized Realms.
+  final Directory? baseFilePath;
```
nirinchev, line 35: We should clarify if the expectation is that this directory exists or if we'll create it automatically.

lib/src/application_configuration.dart
```diff
@@ -0,0 +1,79 @@
...
+  final Duration? defaultRequestTimeout;
+
+  /// The [localAppName] is the friendly name identifying the current client application.
+  /// This is typically used to differentiate between client applications that use the same
+  /// MongoDB Realm app. These can be the same conceptual app developed for different platforms, or
+  /// significantly different client side applications that operate on the same data - e.g. an event managing
+  /// service that has different clients apps for organizers and attendees.
+  final String? localAppName;
+
+  /// The [localAppVersion]
```
nirinchev, line 52: This looks like Apple docs - can we expand it a little?

lib/src/application_configuration.dart
```diff
@@ -0,0 +1,79 @@
...
+  final Version? localAppVersion;
+
+  final MetadataPersistenceMode metadataPersistenceMode;
```
nirinchev, line 55: Missing docs

lib/src/application_configuration.dart
```diff
@@ -0,0 +1,79 @@
...
+
+  /// The encryption key for user metadata on this device.
+  /// This will not change the encryption key for individual Realms, which is set in the [Configuration]
+  final List<int>? metadataEncryptionKey;
```
nirinchev, line 59: Clarify the requirements here - e.g. that this should be 64 bytes. Also, is there a better type we can use than List<int> that is more expressive? I found https://api.dart.dev/stable/2.16.1/dart-typed_data/ByteData-class.html but not sure if it would fit our needs.

lib/src/native/realm_core.dart
```diff
@@ -809,6 +811,33 @@ class _RealmCore {
   RealmAppCredentialsHandle createAppCredentialsAnonymous() {
     return RealmAppCredentialsHandle._(_realmLib.realm_app_credentials_new_anonymous());
   }
+
+  AppConfigHandle createAppConfig(ApplicationConfiguration configuration, RealmHttpTransportHandle httpTransport) {
```
nirinchev, line 815: I don't think this needs to be public - we should probably have a `AppHandle createApp(ApplicationConfiguration configuration)` method that internally creates the `AppConfigHandle` and the `SyncConfigHandle`.

nirinchev (changes requested):
Some docs comments + we seem to be missing the method to create sync config handle.

lib/src/application_configuration.dart
nielsenko: Typically byte arrays are parsed around as `List<int>` in dart.  I think `ByteData` is a bit overkill here. It comes with a lot of functionality for working with the data. We could use `Uint8List`, but basically the user would have to use `Uint8List.fromBytes(List<int>)`.. which is used internally, but it will just be a bother for the user. We could roll over own `EncryptionKey` class, but maybe a bit overkill.

I will update the comment to mention the 64 bytes requirement.

lib/src/application_configuration.dart
nielsenko: We will user `unencrypted` instead of `plainText`. Also, added forgotten `disabled` value, and added doc comments

lib/src/application_configuration.dart
```diff
@@ -0,0 +1,96 @@
...
+  final Directory? baseFilePath;
+
+  /// The [baseUrl] is the [Uri] used to reach the MongoDB Realm server.
+  /// 
+  /// [baseUrl] only needs to be set if for some reason your application isn't hosted on realm.mongodb.com.
+  /// This can be the case if you're testing locally or are using a pre-production environment.
+  final Uri? baseUrl;
+
+  /// The [defaultRequestTimeout] for HTTP requests performed as part of authentication.
+  final Duration? defaultRequestTimeout;
```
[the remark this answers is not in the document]
nielsenko: I agree and I have pushed a fix. One caveat is that it repeats the defaults in core.

lib/src/native/realm_core.dart
nielsenko: The method is called `getApp(ApplicationConfiguration configuration)` but is part of the top PR (branch `realm_app`).

lib/src/application_configuration.dart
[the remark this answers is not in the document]
nirinchev: That's even better because Core is well known for accidentally shipping breaking changes 😄

PR #309, opened 2022-03-23
Add jobs to deploy and cleanup baas on cloud-dev

nirinchev:
This is work in progress. Initially will be added only on Linux job and once that works fine, I'll add to remaining jobs.

TODO:
- [x] Revert unrelated workflow changes
- [x] Changelog entry


nirinchev:
@nielsenko @blagoev this should be ready-ish for review. I was hoping to get a CI run to confirm, but feel free to start going over it as I don't expect it'll fundamentally change.

.github/workflows/pr.yml
```diff
@@ -0,0 +1,104 @@
...
+          sdk: stable
+      - name: Install dependencies
+        run: dart pub get
+      - name: Run tests
+        run: |
+          dart test -r expanded -j 1 --test-randomize-ordering-seed random
+
+  cleanup-baas:
+    runs-on: ubuntu-latest
+    name: Cleanup Linux
```
nielsenko, line 91:
```suggestion
    name: Cleanup BaaS
```

.github/workflows/pr.yml
```diff
@@ -0,0 +1,104 @@
...
+    needs:
+      - tests-linux
+    if: always()
+    timeout-minutes: 5
+    steps:
+      - uses: realm/ci-actions/mdb-realm/cleanup@fac1d6958f03d71de743305ce3ab27594efbe7b7
```
nielsenko, line 97: Could we tag this commit and reference it by tag?

.github/workflows/pr.yml
```diff
@@ -0,0 +1,104 @@
...
+    name: BaaS Linux
+    outputs:
+      clusterName: ${{ steps.deploy-mdb-apps.outputs.clusterName }}
+    timeout-minutes: 20
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: false
+      - uses: realm/ci-actions/mdb-realm/deployApps@fac1d6958f03d71de743305ce3ab27594efbe7b7
```
nielsenko, line 23: Could we tag this commit and reference it by tag?

.github/workflows/pr.yml
nirinchev: This is in the PR build which will eventually contain more than one platform. Each platform where we run sync tests will need its own deploy and cleanup jobs, which is why I tried to name them similarly for consistency.

.github/workflows/pr.yml
nirinchev: This is a bad practice because tags can be overwritten. It probably doesn't matter that much for our internal repositories, but I've always had it as a policy to use the commit sha for actions. [Here's the relevant section](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) from the Github docs.

lib/src/cli/deployapps/baas_client.dart
```diff
@@ -0,0 +1,294 @@
...
+
+  late String _groupId;
+
+  BaasClient._(String baseUrl, [this._clusterName])
+      : _baseUrl = "$baseUrl/api/admin/v3.0",
+        _headers = <String, String>{'Accept': 'application/json'},
+        _appSuffix = "-$_clusterName";
+
+  static Future<BaasClient> docker(String baseUrl) async {
+    final BaasClient result = BaasClient._(baseUrl);
```
nielsenko, line 54:
```suggestion
    final result = BaasClient._(baseUrl);
```

.github/workflows/pr.yml
nirinchev: Ah, good call - I didn't see that.

.github/workflows/pr.yml
nielsenko: I see your point, but it makes it very hard to see if we are using the latest action version. Do you suggest we do the same for external actions?

lib/src/cli/deployapps/baas_client.dart
```diff
@@ -0,0 +1,294 @@
...
+    final appId = doc['_id'] as String;
+    final clientAppId = doc['client_app_id'] as String;
+
+    final app = BaasApp(appId, clientAppId, name);
+
+    final confirmFuncId = await _createFunction(app, "confirmFunc", _confirmFuncSource);
+    final resetFuncId = await _createFunction(app, "resetFunc", _resetFuncSource);
+
+    enableProvider(app, "anon-user");
+    enableProvider(app, "local-userpass", '''{
```
nielsenko, line 122: Have you considered doing a proper type, and use `json_serializable` to source generate the serialization code?

test/test.dart
```diff
@@ -154,7 +159,23 @@ void parseTestNameFromArguments(List<String>? arguments) {
...
     testName = arguments[nameArgIndex + 1];
   }
 }
+
+Future<void> setupBaas() async {
+  var baasUrl = Platform.environment['BAAS_URL'];
+  var cluster = Platform.environment['BAAS_CLUSTER'];
+  var apiKey = Platform.environment['BAAS_API_KEY'];
+  var privateApiKey = Platform.environment['BAAS_PRIVATE_API_KEY'];
+  var projectId = Platform.environment['BAAS_PROJECT_ID'];
```
nielsenko, line 172:
```suggestion
  final baasUrl = Platform.environment['BAAS_URL'];
  final cluster = Platform.environment['BAAS_CLUSTER'];
  final apiKey = Platform.environment['BAAS_API_KEY'];
  final privateApiKey = Platform.environment['BAAS_PRIVATE_API_KEY'];
  final projectId = Platform.environment['BAAS_PROJECT_ID'];
```

test/test.dart
```diff
@@ -154,7 +159,23 @@ void parseTestNameFromArguments(List<String>? arguments) {
...
+
+  if (baasUrl == null) {
+    return;
+  }
+
+  final BaasClient client = await (cluster == null ? BaasClient.docker(baasUrl) : BaasClient.atlas(baasUrl, cluster, apiKey!, privateApiKey!, projectId!));
```
nielsenko, line 178:
```suggestion
  final client = await (cluster == null ? BaasClient.docker(baasUrl) : BaasClient.atlas(baasUrl, cluster, apiKey!, privateApiKey!, projectId!));
```

.github/workflows/pr.yml
nirinchev: Ideally, this should be done for all actions, but at the very least, we should do it for actions where we pass secrets.

test/test.dart
```diff
@@ -154,7 +159,23 @@ void parseTestNameFromArguments(List<String>? arguments) {
   }
 
   int nameArgIndex = arguments.indexOf("--name");
-  if (nameArgIndex >= 0 && arguments.length > 1) {
+  if (nameArgIndex >= 0 && arguments.length > nameArgIndex) {
```
nielsenko, line 162: What if `nameArgIndex` == `arguments.length - 1`?

lib/src/cli/deployapps/baas_client.dart
nirinchev: Yeah - I thought about it, but felt it's an overkill considering we're mostly serializing hardcoded values. Almost none of the payloads or responses are reused meaning we get very little value from having them be strongly typed.

lib/src/cli/deployapps/deployapps_command.dart
```diff
@@ -0,0 +1,78 @@
...
+      if (options.privateApiKey == null) {
+        abort('--private-api-key must be supplied when --atlas-cluster is not set');
+      }
+
+      if (options.projectId == null) {
+        abort('--project-id must be supplied when --atlas-cluster is not set');
+      }
+    }
+
+    final BaasClient client = await (options.atlasCluster == null
```
nielsenko, line 61:
```suggestion
    final client = await (options.atlasCluster == null
```

test/test.dart
nirinchev: That would mean that whoever ran the tests didn't actually pass in the name? The way I read the original code, the arguments would look like `*args* --name someName *more args*`. With the way it was written, we would get a crash if someone ran the command with `*args* --name` - was that desirable?

nielsenko: You are right, I believe the args package will bail before we get here

lib/src/cli/deployapps/baas_client.dart
```diff
@@ -0,0 +1,294 @@
...
+
+  static Future<BaasClient> docker(String baseUrl) async {
+    final result = BaasClient._(baseUrl);
+
+    await result._authenticate('local-userpass', '{ "username": "unique_user@domain.com", "password": "password" }');
+
+    dynamic groupDoc = await result._get('auth/profile');
+    result._groupId = (groupDoc['roles'] as List<dynamic>)[0]['group_id'] as String;
+
+    print('Current GroupID ${result._groupId}');
```
nielsenko, line 61: Consider avoiding print directly and use logger instead (see metrics)

lib/src/cli/deployapps/baas_client.dart
```diff
@@ -0,0 +1,294 @@
...
+      result[defaultApp.name] = defaultApp;
+    }
+
+    return result;
+  }
+
+  Future<List<BaasApp>> _getApps() async {
+    final apps = await _get('groups/$_groupId/apps') as List<dynamic>;
+    return apps
+        .map((dynamic doc) {
```
nielsenko, line 95:
```suggestion
        .map((doc) {
```

nielsenko: Yes indeed we do..

.github/workflows/pr.yml
```diff
@@ -0,0 +1,154 @@
...
+        with:
+          submodules: 'recursive'
+      - name: Setup Ninja
+        uses: seanmiddleditch/gha-setup-ninja@master
+      - name: Build Realm Dart for Linux
+        run: ./scripts/build-linux.sh
+      - name: Store artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: librealm-linux
```
nielsenko, line 51: You need to name this after a unique finger-print of the sources. Previously I used `git submodule status` on `src/realm-core`, but that was not deemed good enough, since it doesn't capture changes to the dart specific files in `src`. Hence I suggest also adding `sha1sum src/* | sha1sum` as well. Also, I suggest you use `actions/cache@v3` instead.

.github/workflows/pr.yml
```diff
@@ -0,0 +1,154 @@
...
+          projectId: ${{ secrets.ATLAS_PROJECT_ID }}
+          apiKey: ${{ secrets.ATLAS_PUBLIC_API_KEY }}
+          privateApiKey: ${{ secrets.ATLAS_PRIVATE_API_KEY }}
+          differentiator: dart-linux
+      - name : Setup Dart SDK
+        uses: dart-lang/setup-dart@main
+        with:
+          sdk: stable
+      - name: Deploy Apps
+        run: dart run realm_dart deploy-apps --baas-url https://realm-dev.mongodb.com --atlas-cluster ${{ steps.deploy-mdb-apps.outputs.clusterName }} --api-key ${{ secrets.ATLAS_PUBLIC_API_KEY }} --private-api-key ${{ secrets.ATLAS_PRIVATE_API_KEY }} --project-id ${{ secrets.ATLAS_PROJECT_ID }}
```
nielsenko, line 35:
```suggestion
        run: |
          dart run realm_dart deploy-apps \
            --baas-url https://realm-dev.mongodb.com \
            --atlas-cluster ${{ steps.deploy-mdb-apps.outputs.clusterName }} \
            --api-key ${{ secrets.ATLAS_PUBLIC_API_KEY }} \
            --private-api-key ${{ secrets.ATLAS_PRIVATE_API_KEY }} \
            --project-id ${{ secrets.ATLAS_PROJECT_ID }}
```

.github/workflows/pr.yml
nirinchev: This is different from caching as it creates an artifact for this particular run. Artifacts are not shared between runs, so they can have the same name. The general recommendation for passing build outputs from one job another is to use artifacts as it has several advantages:
1. They can not be accidentally overwritten by other runs.
2. They can be downloaded after the run completes (they appear as files in the workflow summary).
3. They can be downloaded deterministically from other workflows - e.g. once we have a proper release workflow, we can find the run for the correct commit and download the artifacts from it.

nielsenko: But I would like to re-use it another run, so that I can avoid building `librealm-dart` unless there are changes in `src`. We can also upload it as an artifact, to satisfy 2 and 3.

nielsenko: So what about building and caching, but still use an artifact to pass output to next job.

nirinchev: Yes, I do plan to add ccache to speed up builds, but that's out of scope for what this PR tries to achieve. In the .NET workflows, we do use a combination of caching and artifact uploads to avoid rebuilding the native binaries and we'll bring some of these mechanics to dart, but I'd rather do it step by step rather than as a single massive PR that'd be a lot harder to review.

nielsenko: Okay. Point taken 👍

lib/src/cli/deployapps/baas_client.dart
nielsenko: I see it is a repeated issue across other commands as well. Lets consider this separately

nielsenko (approved):

.github/workflows/pr.yml
```diff
@@ -0,0 +1,160 @@
...
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          submodules: false
+      - name: Setup GTK
+        run: |
+          sudo apt-get update -y
+          sudo apt-get install -y libgtk-3-dev xvfb
+      - name: Setup Ninja
+        uses: seanmiddleditch/gha-setup-ninja@master
```
fealebenpae, line 128: Why do we need Ninja here if we're using the artifact from the previous job already?

test/test.dart
```diff
@@ -92,9 +95,11 @@ void xtest(String? name, dynamic Function() testFunction) {
   testing.test(name, testFunction, skip: "Test is disabled");
 }
 
-void setupTests(List<String>? args) {
+Future<void> setupTests(List<String>? args) async {
```
desistefanova, line 98: I think it will be good to separate local tests from sync tests, as it is in .Net. Maybe having another setupSyncTests method would be better. And then to have another sync_test.dart file using this new test setup. So that the old tests will still be working without BaaS.

nirinchev: The old tests should still work without baas. The only change I made to `setupTests` is to call `setupBaas`, which is a no-op if the `BAAS_URL` environment variable is not set:

https://github.com/realm/realm-dart/blob/2c43087906deef2c1d25aa6546afc329b70271ae/test/test.dart#L174-L176

.github/workflows/pr.yml
nirinchev: I believe this is needed to compile the native flutter application: https://github.com/realm/realm-dart/blob/2c43087906deef2c1d25aa6546afc329b70271ae/.github/workflows/pr.yml#L143

fealebenpae: _Interesting_. I guess Flutter doesn't ship precompiled binaries yet like Electron does.

lib/src/cli/deployapps/baas_client.dart
```diff
@@ -0,0 +1,302 @@
...
+    // takes a few seconds to provision a user for BaaS, meaning enabling sync
+    // will fail if we attempt to do it with the same request. It's nondeterministic
+    // how long it'll take, so we must retry for a while.
+    var attempt = 0;
+    while (true) {
+      try {
+        await _patch('groups/$_groupId/apps/$app/services/$mongoServiceId/config', syncConfig);
+        break;
+      } catch (err) {
+        if (attempt++ < 1) {
```
desistefanova, line 221: It will cause only 1 retry, if I'm not wrong. In the message is written "Will keep retrying". Is one retry enough?

test/test.dart
desistefanova: OK, thanks! Still wondering, what will happen if baasUrl is set but it is wrong or not available?

desistefanova (approved):

lib/src/cli/deployapps/baas_client.dart
nirinchev: Well spotted - I changed it to make sure that the error gets propagated correctly and forgot to return it to a more sane value.

test/test.dart
nirinchev: The tests will fail, which I think is the desired behavior - if you are relying on testing some sync functionality (which you probably are if you've set the `BAAS_URL` env variable), then you probably want to know straight away that something is wrong.

.github/workflows/pr.yml
```diff
@@ -0,0 +1,160 @@
...
+        uses: actions/checkout@v2
+        with:
+          submodules: false
+      - uses: realm/ci-actions/mdb-realm/deployApps@fac1d6958f03d71de743305ce3ab27594efbe7b7
+        id: deploy-mdb-apps
+        with:
+          projectId: ${{ secrets.ATLAS_PROJECT_ID }}
+          apiKey: ${{ secrets.ATLAS_PUBLIC_API_KEY }}
+          privateApiKey: ${{ secrets.ATLAS_PRIVATE_API_KEY }}
+          differentiator: dart-linux
```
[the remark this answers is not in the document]
nirinchev: The name is fairly irrelevant - it gets hashed in combination with some other variables. As long as each job in a workflow has a unique name, it doesn't really matter what that name is.

bin/README.md
```diff
@@ -1,10 +1,10 @@
-Dart packages should distribute their prebuilt native extension binaries (if any) in the lib directory of the package. 
-This is where Dart VM looks for the binary when an application loads the Dart package. 
-On Windows Dart VM have a bug which looks for the package with incorrect path. 
-For Realm Dart this is something like C:\C:\<DART_PACKAGES_CACHE_PATH>\realm_dart\lib\src\realm_dart.dartrealm_dart_extension.dll. 
-The name of the native extension binary and the device letter is incorreth path on Windows. 
+Dart packages should distribute their prebuilt native extension binaries (if any) in the lib directory of the package.
```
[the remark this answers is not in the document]
nirinchev: I think it's prettier.

lib/src/cli/deployapps/baas_client.dart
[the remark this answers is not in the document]
nirinchev: You're right - I forgot to return it to a sane value after testing something.

.github/workflows/pr.yml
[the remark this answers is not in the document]
blagoev: could we rename to `realm-dart` to avoid confusion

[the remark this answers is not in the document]
nirinchev: No, because we're going to have tests for different platforms - e.g. linux, macOS, windows. They need to each deploy their own BaaS server and run tests against them. So we'll have `dart-linux`, `dart-macos`, etc. This also enables us to use the same naming scheme if we want to have flutter tests on desktop platforms - e.g. `flutter-linux`, etc. I don't think we need to run all those tests on each PR build as that will slow things down, but might be a good idea to run them on every merge to master.

lib/src/cli/deployapps/baas_client.dart
```diff
@@ -0,0 +1,302 @@
...
+
+    dynamic groupDoc = await result._get('auth/profile');
+    result._groupId = (groupDoc['roles'] as List<dynamic>)[0]['group_id'] as String;
+
+    print('Current GroupID ${result._groupId}');
+
+    return result;
+  }
+
+  /// A client that imports apps to an MongoDB Realm environment (typically realm-dev or realm-qa).
```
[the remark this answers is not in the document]
nirinchev: Added `@nodoc` to the client doc strings. The command itself didn't have any doc strings, so I didn't change that - let me know if you meant that you'd like me to add docstrings to command.

.github/workflows/pr.yml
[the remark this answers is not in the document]
blagoev: I see, thanks.

lib/src/cli/deployapps/baas_client.dart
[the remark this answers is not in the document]
blagoev: no its good as it is now. The api docs are usefull for us internally and my idea was to hide it just so we don't accidentally include it in the SDK API docs.

blagoev (approved):

blagoev (approved):

PR #310, opened 2022-03-23
configuration - disable format upgrade

### LINKED ISSUE #297: Support realm disable format upgrade

desistefanova:
Implement configuration property - disableFormatUpgrade. Fixes #297.


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ x.x.x Release notes (yyyy-MM-dd)
 **This project is in the Alpha stage. All API's might change without warning and no guarantees are given about stability. Do not use it in production.**
 
 ### Enhancements
+* Support disableFormatUpgrade ([#310](https://github.com/realm/realm-dart/pull/310))
```
nirinchev, line 7: [nit] This is clear to me, but is likely not clear for end-users. Might not be super critical right now, but we should aim to keep changelog entries as understandable as possible as for some users they're the first documentation for each new feature. In this case it might be better to have something along the lines of:

```suggestion
* Added a new `Configuration` option: `disableFormatUpgrade`. When set to `true`, opening a Realm with an older file format will throw an exception to avoid automatically migrating it. ([#310](https://github.com/realm/realm-dart/pull/310))
```

test/configuration_test.dart
```diff
@@ -153,4 +153,48 @@ Future<void> main([List<String>? args]) async {
...
+    final realmPath = combineFilePath(realmDir, Configuration.defaultPath);
+
+    await Directory(realmDir).create();
+    var config = Configuration([Car.schema])..path = realmPath;
+    var realm = Realm(config);
+    realm.close();
+
+    Realm.deleteRealm(realmPath);
+    var file = File(realmBundleFile);
+    await file.copy(realmPath);
```
nirinchev, line 169: This seems to be unnecessarily complicated. It shouldn't be necessary to open the Realm the first time. Instead, we could do something like:

```dart
final realmBundleFile = "test/data/realm_files/realm-bundle.realm";
final config = Configuration([Car.schema], disableFormatUpgrade: true);
await File(realmBundleFile).copy(config.path);

// try to open the Realm with config
```

test/test.dart
```diff
@@ -164,6 +164,10 @@ void parseTestNameFromArguments(List<String>? arguments) {
   }
 }
 
+String combineFilePath(String directory, String filename) {
+  return _path.join(directory, _path.basename(filename));
+}
```
nirinchev, line 169: This function seems confusing to me - it sounds like it's going to combine two paths (similarly to `_path.join`), but that's not what it does. In any case, we should be able to remove it if my suggestion above works.

test/configuration_test.dart
```diff
@@ -153,4 +153,48 @@ Future<void> main([List<String>? args]) async {
...
+    expect(() {
+      realm = Realm(config);
+    }, throws<RealmException>("The Realm file format must be allowed to be upgraded in order to proceed"));
+    realm.close();
+
+    await Directory(realmDir).delete(recursive: true);
+  }, skip: isFlutterPlatform);
+
+  test('Configuration - disableFormatUpgrade=false', () async {
+    final realmBundleFile = "test/data/realm_files/realm-bundle.realm";
```
nirinchev, line 181: Can't comment on the right thing, but `realm-bundle.realm` is an overly vague name. We'll likely need to add more such files, e.g. files for migration, etc. so it's important to name them in a more descriptive way. This one could be `old-format.realm`.

nirinchev (changes requested):

test/configuration_test.dart
desistefanova: Done.

test/test.dart
desistefanova: I removed this method.

nirinchev (approved):

lib/src/configuration.dart
```diff
@@ -99,6 +99,11 @@ class Configuration {
   /// The FIFO special files are very lightweight and the main [Realm] file will still be stored in the location defined
   /// by the [path] you  property. This property is ignored if the directory defined by [path] allow FIFO special files.
   final String? fifoFilesFallbackPath;
+
+  /// Specifies if a [Realm] file format should be automatically upgraded
+  /// if it was created with an older version of the [Realm] library.
+  /// If set to true and a file format upgrade is required, an error will be thrown instead.
```
blagoev, line 105:
```suggestion
  /// If set to true and a file format upgrade is required, an exception will be thrown instead.
```

blagoev, line 105:
```suggestion
  /// An exception will be thrown if a file format upgrade is required.
```

lib/src/native/realm_core.dart
```diff
@@ -151,6 +151,9 @@ class _RealmCore {
       if (config.fifoFilesFallbackPath != null) {
         _realmLib.realm_config_set_fifo_path(configHandle._pointer, config.fifoFilesFallbackPath!.toUtf8Ptr(arena));
       }
+
+      _realmLib.realm_config_set_disable_format_upgrade(configHandle._pointer, config.disableFormatUpgrade);
```
blagoev, line 155: This should be called only if the flag is set to true

lib/src/configuration.dart
blagoev: One note: We should be careful when wording `exception` and `error` since these have different meaning.

blagoev (changes requested):

lib/src/native/realm_core.dart
desistefanova: I wondered whether to do it because the similar checks were deleted for the rest of properties that have default value `false`. For example isInMemory and isReadOnly.

blagoev: hm, I am not sure why

blagoev (changes requested):

lib/src/native/realm_core.dart
```diff
@@ -143,17 +143,21 @@ class _RealmCore {
...
       _realmLib.realm_config_set_schema(configHandle._pointer, schemaHandle._pointer);
       _realmLib.realm_config_set_schema_version(configHandle._pointer, config.schemaVersion);
-
-      final schemaMode = config.isReadOnly ? realm_schema_mode.RLM_SCHEMA_MODE_IMMUTABLE : realm_schema_mode.RLM_SCHEMA_MODE_AUTOMATIC;
-      _realmLib.realm_config_set_schema_mode(configHandle._pointer, schemaMode);
-      _realmLib.realm_config_set_in_memory(configHandle._pointer, config.isInMemory);
-
+      _realmLib.realm_config_set_path(configHandle._pointer, config.path.toUtf8Ptr(arena));
+      _realmLib.realm_config_set_scheduler(configHandle._pointer, schedulerHandle._pointer);
+      if (config.isReadOnly) {
```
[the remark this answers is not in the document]
desistefanova: Yes, right. Actually it doesn't matter since RLM_SCHEMA_MODE_AUTOMATIC is the default one. I will return it back.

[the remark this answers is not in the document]
desistefanova: Isn't it better to be like this:
```dart
   if (config.isReadOnly) {
        _realmLib.realm_config_set_schema_mode(configHandle._pointer, realm_schema_mode.RLM_SCHEMA_MODE_IMMUTABLE);
      }
```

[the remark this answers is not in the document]
blagoev: i think this one is ok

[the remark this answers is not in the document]
blagoev: Are you sure RLM_SCHEMA_MODE_AUTOMATIC is the default one?

PR #313, opened 2022-03-25
Move the macos workflows to pr.yml

nirinchev:
Based on https://github.com/realm/realm-dart/pull/309, will rebase once merged.


.github/workflows/pr.yml
```diff
@@ -51,15 +53,14 @@ jobs:
           submodules: 'recursive'
       - name: Setup Ninja
         uses: seanmiddleditch/gha-setup-ninja@master
-      - name: Build Realm Dart for Linux
+      - name: Build Realm for Linux
```
nielsenko, line 56: I'm okay with this change, but the library being build is `librealm_dart.so`. I have wondered, if it would be better to build two libs, ie. also a `librealm.so`, so that the later can be shared across SDKs, with the Dart VM specific parts only, being in `librealm_dart.so`. Anyway - that is not for this PR, just something to think about.

src/CMakeLists.txt
```diff
@@ -79,9 +79,9 @@ endif()
 
 
 string(APPEND OUTPUT_DIR "${PROJECT_SOURCE_DIR}/binary")
-if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Windows")
+if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
     string(APPEND OUTPUT_DIR "/windows")
-elseif(CMAKE_HOST_SYSTEM_NAME STREQUAL "Linux")
+elseif(CMAKE_SYSTEM_NAME STREQUAL "Linux")
```
nielsenko, line 84: 👍

nielsenko (approved):

.github/workflows/pr.yml
nirinchev: This is the name of the library used by both flutter and dart - at least looking at the [latest release](https://github.com/realm/realm-dart/releases/tag/0.2.1%2Balpha), all libraries except for iOS are librealm_dart. I'm fine with renaming it to librealm, but I agree it should be done as a separate PR as it will likely be more involved than what I wanted to do in this one.

nirinchev:
The CI failures are related to a botched deployment of Atlas on cloud-dev. The team is investigating but they're not a merge stopper.

desistefanova (approved):

.github/workflows/pr.yml
blagoev: Why are we thinking renaming the library. I think we are totally sidetracking here. It's fine to rename the step.

nirinchev: Not sure I understand - why would we rename the step? This is building the native Realm library for Linux. It's used by both dart and flutter, which is why I think it's incorrect to call it "Build Realm Dart".

nirinchev: Also, I don't think it's particularly important to rename the library and in any case, not right now.

blagoev: yes indeed.

blagoev (approved):

PR #314, opened 2022-03-25
Consolidate on single config.yaml file

nielsenko:
One File to rule them all, One File to find them, One File to bring them all, and in the darkness bind them,


blagoev (changes requested):
Are we sure this works for all platforms. Does this work on Linux? I see that right now it has some custom paths which should be changed to a generic one, but still it needs different config than macos.

nielsenko:
> Are we sure this works for all platforms. Does this work on Linux? I see that right now it has some custom paths which should be changed to a generic one, but still it needs different config than macos.

Assuming you have clang 11 lib installed correctly. Fx, on ubuntu: 
```
sudo apt-get install libclang-11-dev
```

blagoev:
@nielsenko I don't mind the change in this case. Could you update che README.md in ffigen with the these details about Linux.

blagoev (approved):

PR #441, opened 2022-03-29
Add support for data initialization function

nirinchev:
This adds a `Configuration` property that allows developers to specify a function that imports an initial set of data the first time the Realm is opened.

```dart
final config = Configuration([Person.schema]);
config.initialDataCallback = (realm) {
  // realm is already in a write transaction
  realm.add(Person('peter'));
  realm.add(Person('george'));
};
```

Fixes https://github.com/realm/realm-dart/issues/298


src/realm_dart_realm.h
```diff
@@ -0,0 +1,39 @@
...
+//
+////////////////////////////////////////////////////////////////////////////////
+
+#ifndef REALM_DART_REALM_H
+#define REALM_DART_REALM_H
+
+#include "realm.h"
+#include "dart_api_dl.h"
+
+typedef int (*realm_dart_data_initialization_func_t)(Dart_Handle callback_func, realm_t*);
```
nirinchev, line 25: Wish I could declare that as a boolean return type but that generates Uin8 in the ffi bindings and I couldn't figure out the dart signature that would be compatible with it.

lib/src/configuration.dart
```diff
@@ -133,6 +133,12 @@ class Configuration {
   /// by the [path] you  property. This property is ignored if the directory defined by [path] allow FIFO special files.
   String get fifoFilesFallbackPath => realmCore.getConfigFifoPath(this);
   set fifoFilesFallbackPath(String value) => realmCore.setConfigFifoPath(this, value);
+
+  Function(Realm realm)? get initialDataCallback => _initialDataCallback;
```
nielsenko, line 137: This looks a bit odd to me.

When dealing with callbacks on the Dart side I expect it either maps to a Future that you can await (if it is a one time thing), or Stream you can listen to, if it can occur multiple times.

So instead of a callback you can register on the `Configuration`, I would expect a `Future` on the `Realm`, something like `Future<void> get initialData => _initialDataCompleter.future` where `Completer<void> _initialDataCompleter`.. this `Completer` (think TaskCompletionSource) is then parsed through to native and back again in the `userdata` parameter.

nielsenko: This would be complicated slightly if we allowed configs to be shared between realms - as core does, but we currently don't allow that in Dart.

lib/src/configuration.dart
```diff
@@ -133,6 +133,12 @@ class Configuration {
...
+  set initialDataCallback(Function(Realm realm)? value) {
+    _initialDataCallback = value;
+    realmCore.setConfigInitialDataCallback(this);
```
nielsenko, line 140: We are highly inconsistent already, but I think we should try to only traffic handles to `realmCore` to separate concerns a bit more concisely.

lib/src/native/realm_core.dart
```diff
@@ -196,6 +198,28 @@ class _RealmCore {
     });
   }
 
+  static int initial_data_callback(Object configHandle, Pointer<shared_realm> realmHandle) {
```
nielsenko, line 201: Handle has a very particular meaning, so I prefer naming this parameter different. Also, I would like it to be a `Completer` instead of a `Configuration`.

lib/src/native/realm_core.dart
```diff
@@ -196,6 +198,28 @@ class _RealmCore {
...
+    try {
+      final config = configHandle as Configuration;
+      final realm = RealmInternal.getUnowned(config, RealmHandle._unowned(realmHandle));
```
nielsenko, line 204: A bit unintuitive you need to guard against `configHandle` being null. Has to with the way `Dart_HandleFromWeakPersistent_DL` works, and the fact that Google in their infinite wisdom, has not exposed `Dart_IsNull_DL` (https://github.com/dart-lang/sdk/issues/48331)

lib/src/native/realm_core.dart
```diff
@@ -196,6 +198,28 @@ class _RealmCore {
...
+      config.initialDataCallback!(realm);
+      return 1;
+    } catch (ex) {
+      // TODO: return false here!
+      print(ex);
+      return 0;
+    }
+  }
+
+  void setConfigInitialDataCallback(Configuration config) {
```
nielsenko, line 214: I would like this to take a `ConfigHandle` and a `Completer` instead

lib/src/native/realm_core.dart
```diff
@@ -196,6 +198,28 @@ class _RealmCore {
     });
   }
 
+  static int initial_data_callback(Object configHandle, Pointer<shared_realm> realmHandle) {
+    try {
+      final config = configHandle as Configuration;
+      final realm = RealmInternal.getUnowned(config, RealmHandle._unowned(realmHandle));
+      config.initialDataCallback!(realm);
```
nielsenko, line 205: Here we could just `complete` the `Completer`

src/realm_dart_realm.h
nielsenko: When you do `Pointer.fromFunction` the argument can just be a Dart function returning bool. It should work 🤔

test/configuration_test.dart
```diff
@@ -153,4 +153,186 @@ Future<void> main([List<String>? args]) async {
...
+  });
+
+  test('Configuration.initialDataCallback can be unset', () {
+    var invoked = false;
+    var config = Configuration([Dog.schema, Person.schema]);
+    config.initialDataCallback = (realm) {
+      invoked = true;
+      realm.add(Dog('fido', owner: Person('john')));
+    };
+    config.initialDataCallback = null;
```
nielsenko, line 186: This obviously won't be possible with my idea, but instead you can just choose not to await the `initialData` future which will have the same effect.

test/configuration_test.dart
```diff
@@ -153,4 +153,186 @@ Future<void> main([List<String>? args]) async {
...
+  test('Configuration.initialDataCallback invoked', () {
+    var invoked = false;
+    var config = Configuration([Dog.schema, Person.schema]);
+    config.initialDataCallback = (realm) {
+      invoked = true;
+      realm.add(Dog('fido', owner: Person('john')));
+    };
+
+    final realm = Realm(config);
+    expect(invoked, true);
```
nielsenko, line 166: Here you depend on core raising the callback immediately on the same thread that did the call. I expect this is an implementation detail, that core might change later.

nirinchev: It's not - the event is raised just before the Realm is returned to the SDK and it's even the same Realm instance that will be returned to SDK. It can't be raised on another thread because we must block returning a Realm instance until the initial data callback has completed.

lib/src/configuration.dart
nielsenko: Maybe I misunderstand the purpose of this callback. I thought it was a way to ensure that you would not block while the realm is setup to be ready for use?

nielsenko: Why would you need to populate the realm before returning the instance. You may as well do that after?

nirinchev: It's a niche use case, but it's something that only happens once - when the Realm is created, which is not something you can easily do with the SDK API as there's no difference between opening an existing Realm vs creating a new one. An example use case would be if you wanted to populate some demo data in your app the first time it's opened and then allow the user to delete it if they want to (think a note taking app with a prepopulated note explaining some of the app functionality).

nielsenko: Why not just check if db is empty, and then populate it? 

But what you are saying is that:
```
final r = Realm();
await r.initialData; 
```
would actually wait forever on restart of the app? Because the callback really signifies that the db file has been created, not that the realm is wired up.

Well, I guess that makes many of my comments mood.

lib/src/native/realm_core.dart
nielsenko: If you do a `if (config is Configuration)` you are golden

test/configuration_test.dart
nielsenko: Ties in with my misunderstanding of the API

lib/src/configuration.dart
nielsenko: Well.. leave a crumb in the db that the user cannot delete. Or check if the realm db path exists first, to see if it is the first.. Anyway, the API is there, and now we support it in dart as well :-)

lib/src/configuration.dart
nielsenko: We did a discussion on this at the round-table. It was agreed that it is inconsistent, but not what to do about it.

There where basically two viewpoints:
1) We should use the higher level objects (RealmList, Configuration, etc) on the interface of `_RealmCore` and let the methods extract the handles they need for these.
2) We should have a clean layering, so that methods on `_RealmCore` never know about the higher level objects, hence always trafficking in handles.

The arguments was not settled, but I believe it was decided to live with this inconsistency for now, and revisist later

nirinchev: Discussed offline - we're keeping it as it is. This will change in any case if/when #455 gets approved.

blagoev (changes requested):

lib/src/native/realm_core.dart
```diff
@@ -196,6 +199,29 @@ class _RealmCore {
...
+  static int initial_data_callback(Object config, Pointer<shared_realm> realmHandle) {
+    try {
+      if (config is Configuration) {
+        final realm = RealmInternal.getUnowned(config, RealmHandle._unowned(realmHandle));
+        config.initialDataCallback!(realm);
+        return TRUE;
+      }
+    } catch (ex) {
+      // TODO: this should propagate the error to Core: https://github.com/realm/realm-core/issues/5366
+      print(ex);
```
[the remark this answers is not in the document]
nirinchev: If we bring down the process, there'll be no way to write a test for this case. Additionally, it's not like we can't tell Core that an exception happened - it's just that we can't tell the user what the precise error is. Right now, it'll just throw a generic error from the realm ctor. This is covered by this test case:

https://github.com/realm/realm-dart/blob/7b04d9d0d1b6312f285b79e1cb8a324ad5896b66/test/configuration_test.dart#L229-L238

The reason this test is skipped is an unrelated bug in Core that leaves the Realm open when such an error occurs, which then prevents us from deleting it.

lib/src/native/realm_core.dart
```diff
@@ -196,6 +199,29 @@ class _RealmCore {
...
+    }
+
+    return FALSE;
+  }
+
+  void setConfigInitialDataCallback(Configuration config) {
+    if (config.initialDataCallback == null) {
+      _realmLib.realm_dart_realm_config_set_initial_data_callback(config.handle._pointer, nullptr, nullptr);
```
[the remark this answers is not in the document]
blagoev: I see. thanks

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
blagoev: Still I wonder if we really should be propagating the unhandled error to core. If we throw an `RealmError` here it will mean Dart/Flutter will provide the necessary logging and stacktrace to the developer. Also it will force the developer to think of unhandled errors from such callbacks. Or said differently if we allow the propagation of unhandled errors core will only throw a more specific exception but loosing the stacktrace in the process versus if we throw `RealmError` from the Dart code then the VM will log the exact stacktrace and will allow Isolate unhandled error callback to run. So in a sense telling Core about the exception is hurting the developer experience a bit. Also as a result the realm creation will fail versus if the user is forced to handle the exception on their own inside the callback than it might turn out that an exception happened while initializing the data but still the realm could be created with no initial data. 
Hope it makes sense. I guess we could discuss it more on the round table.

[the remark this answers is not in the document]
blagoev: Forgot to mention there is a way to test this using `Isolate.spawn`

blagoev (changes requested):

lib/src/configuration.dart
blagoev, line 137: We need API docs and CHANGELOG entry

blagoev (changes requested):

src/realm_dart.cpp
```diff
@@ -54,6 +54,76 @@ RLM_API void realm_delete_finalizable(Dart_FinalizableHandle finalizable_handle,
...
+    delete initial_data_callback_data;
+    return result;
+}
+
+RLM_API void realm_dart_realm_config_set_initial_data_callback(realm_config_t* config,
+                                          Dart_Handle managed_config,
+                                          realm_dart_data_initialization_func_t on_data_initialization)
+{
+    if (managed_config && on_data_initialization) {
+        auto initial_data_callback_data = new InitialDataCallbackData(managed_config, on_data_initialization);
```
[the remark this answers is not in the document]
nirinchev: Good point - I think I'll rebase this on https://github.com/realm/realm-dart/pull/455 to make sure we make the entire config immutable.

lib/src/native/realm_core.dart
```diff
@@ -196,6 +199,29 @@ class _RealmCore {
     });
   }
 
+  static int initial_data_callback(Object config, Pointer<shared_realm> realmHandle) {
+    try {
+      if (config is Configuration) {
+        final realm = RealmInternal.getUnowned(config, RealmHandle._unowned(realmHandle));
+        config.initialDataCallback!(realm);
```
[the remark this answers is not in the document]
nirinchev: I don't have strong feelings about it - java uses `initialData` and js uses `onFirstOpen` - we can either choose to align with them or invent a new name that is closer to what dart users will expect. I don't have much experience there, so will defer to you, @nielsenko and @desistefanova for advice.

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
nirinchev: Hm... looks like the ffi will prevent us from bringing down the process here, even if we throw. When we construct the function pointer, we have to provide `exceptionalReturn` argument:

https://github.com/realm/realm-dart/blob/a8502e156635e3d8470270e784315283a160e7b6/lib/src/native/realm_core.dart#L160-L161

My guess is `Pointer.fromFunction` does a try-catch internally and returns the exceptional return value (`FALSE`), even if we don't handle the exception ourselves.

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
desistefanova: I like it as it is `initial_data_callback`. The dart user will use only `initialDataCallback`.

test/configuration_test.dart
```diff
@@ -143,4 +143,143 @@ Future<void> main([List<String>? args]) async {
...
+    realm.close();
+
+    var invokedAgain = false;
+    final configAgain = Configuration([Person.schema], initialDataCallback: (realm) {
+      invokedAgain = true;
+      realm.add(Person('p1'));
+      realm.add(Person('p2'));
+    });
+
+    final realmAgain = getRealm(config);
```
desistefanova, line 185:
```suggestion
    final realmAgain = getRealm(configAgain);
```

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
blagoev: Oh, my bad. I have tested it and indeed we can't throw through int callbacks. I guess my suggestion is to just remove the `print` in this case. Later when its possible we could report the exact error to core to be included in the generic error they throw.

blagoev:
my PR review comments are all over the place. Sorry about that. Don't know how it happens

nirinchev:
@blagoev @nielsenko @desistefanova this is now rebased on master and cleaned up, so should be ready for another look. It changed significantly since I integrated the GCHandle approach, so it's likely worth it reviewing the core functionality from scratch.

lib/src/native/realm_core.dart
```diff
@@ -157,10 +160,31 @@ class _RealmCore {
         _realmLib.realm_config_set_disable_format_upgrade(configHandle._pointer, config.disableFormatUpgrade);
       }
 
+      if (config.initialDataCallback != null) {
+        _realmLib.realm_config_set_data_initialization_function(
+            configHandle._pointer, Pointer.fromFunction(initial_data_callback, FALSE), _realmLib.gc_handle_weak_new(config));
```
desistefanova, line 165:
```suggestion
        _realmLib.realm_config_set_data_initialization_function(
            configHandle._pointer, Pointer.fromFunction(initial_data_callback, FALSE), config.toGCHandle());
```

lib/src/native/realm_core.dart
```diff
@@ -157,10 +160,31 @@ class _RealmCore {
...
+      }
+
       return configHandle;
     });
   }
 
+  static int initial_data_callback(Pointer<Void> userdata, Pointer<shared_realm> realmHandle) {
+    try {
+      final config = _realmLib.gc_handle_deref(userdata);
```
desistefanova, line 174:
```suggestion
      final Configuration? config =  userdata.toObject();
      if (config== null) {
          return;
      }
```

lib/src/native/realm_core.dart
```diff
@@ -157,10 +160,31 @@ class _RealmCore {
...
+      if (config is Configuration) {
```
desistefanova, line 175: Maybe will have to delete this if statement after applying the suggestion above.

desistefanova (changes requested):

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
desistefanova: onFirstOpen is explayning the exact occurance of this event. It is great.

desistefanova (approved):

blagoev (approved):

PR #445, opened 2022-03-30
application credentials email/password

### LINKED ISSUE #392: Support App Credentials email\password
realm_app_credentials_new_email_password

desistefanova:
Implemented application credentials email/password. Fixes #392 
` final credentials = Credentials.emailPassword("test@email.com", "000000");`


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 * Support anonymouse application credentials ([#443](https://github.com/realm/realm-dart/pull/443/))
```
```suggestion
* Support anonymous application credentials ([#443](https://github.com/realm/realm-dart/pull/443/))
```

blagoev (changes requested):

lib/src/native/realm_core.dart
```diff
@@ -762,6 +770,18 @@ extension _RealmLibraryEx on RealmLibrary {
   }
 }
 
+Pointer<realm_string_t> _toRealmString(String value, Allocator allocator) {
```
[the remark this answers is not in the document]
desistefanova: OK.

blagoev (approved):

lib/src/app_credentials.dart
```diff
@@ -41,6 +41,14 @@ class Credentials {
   Credentials.anonymous()
       : _handle = realmCore.createAppCredentialsAnonymous(),
         provider = AuthProvider.anonymous;
+
+  /// Returns a [Credentials] object that can be used to authenticate a user with their email and password.
+  /// A user can login with email and password only after they have registered their account and verified their
+  /// email.
+  /// [Email/Password Authentication Docs](https://docs.mongodb.com/realm/authentication/email-password)
+  Credentials.emailPassword(String email, String password)
+      : _handle = realmCore.createAppCredentialsEmailPassword(email, password),
```
nielsenko, line 50: This ties into @nirinchev's proposal, that we should not create the handle until actually needed. If we do agree on that then this needs to change somewhat.. 

Fx. we could make a subclass `EmailAndPassworsCredentials` that has two final fields (email & password) and have `_handle` be late with same initializer as here.

But again.. depends on what stance we take on @nirinchev 's PR..

lib/src/native/realm_core.dart
```diff
@@ -743,6 +751,14 @@ extension _StringEx on String {
     nativeString.last = 0; // zero terminate
     return result.cast();
   }
+
+  Pointer<realm_string_t> toRealmString(Allocator allocator) {
```
nielsenko, line 755: I feel this belongs somewhere else.. not really relates to `LastError`

lib/src/app_credentials.dart
nielsenko: To untangle the PR jam, I think we should merge as is. We can always revisit

nielsenko (approved):
May or may not refactor later depending on: #455

nirinchev (approved):
This looks good to me, but why is the Core API accepting different types for the email and password? Seems kinda weird.

lib/src/app_credentials.dart
desistefanova: Yes, we can change it later.

lib/src/native/realm_core.dart
desistefanova: It is in `extension _StringEx on String {`

Issue #442, opened 2022-03-29
Starting a write transaction inside a write transaction will rollback the original transaction


nirinchev:
The following awkward code fails in an even more awkward way:

```dart
final config = Configuration([Car.schema]);
final realm = Realm(config);
realm.write(() {
  // Second write inside the first one fails but the error is caught
  try {
    realm.write(() {});
  } catch (ex) {}
});
```

The error is:

```
RealmException: Could not commit write. Error code: 5 . Message: Can't commit a non-existing write transaction
```

What ends up happening is
1. We start the outer write
2. We start inner write
3. Inner write fails
4. We hit the [`realm.write`](https://github.com/realm/realm-dart/blob/9e7e9796abcb90958cf0290e15e20432a4d21545/lib/src/realm_class.dart#L180-L182) catch block, which rolls back the outer transaction and rethrows
5. We catch the exception from the inner write
6. Realm tries to commit the outer write, which fails because the transaction was rolled back in 4.

PR #448, opened 2022-03-30
Fix write inside write rollback; expose Realm.isInTransaction

nirinchev:
Fixes https://github.com/realm/realm-dart/issues/442

The `Transaction` class is currently undocumented as we don't have explicit `beginWrite` method. We may introduce it at a later point, especially once we get async writes.

I made `isInTransaction` public because it's quite useful when you are calling methods from inside your write transaction. Since starting a second transaction will throw, we should expose a mechanism for developers to guard against it. Example:

```dart

void markComplete(Task task) {
  if (task.realm.isInTransaction) {
    task.isComplete = true;
  } else {
    task.realm.write(() {
      task.isComplete = true;
    });
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,10 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 * Support result value from write transaction callbacks ([#294](https://github.com/realm/realm-dart/pull/294/))
+* Added a property `Realm.isInTransaction` that indicates whether the Realm instance has an open write transaction associated with it.
 
 ### Fixed
-* None
+* Fixed an issue that would result in the wrong transaction being rolled back if you start a write transaction inside a write transaction. (Issue [#442](https://github.com/realm/realm-dart/issues/442))
 
 ### Compatibility
 * Dart ^2.15 on Windows, MacOS and Linux
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ x.x.x Release notes (yyyy-MM-dd)
 * Added a property `Realm.isInTransaction` that indicates whether the Realm instance has an open write transaction associated with it.
 
 ### Fixed
-* Fixed an issue that would result in the wrong transaction being rolled back if you start a write transaction inside a write transaction. (Issue [#442](https://github.com/realm/realm-dart/issues/442))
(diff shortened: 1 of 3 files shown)


desistefanova (approved):

CHANGELOG.md
```diff
@@ -5,9 +5,10 @@ x.x.x Release notes (yyyy-MM-dd)
 
 ### Enhancements
 * Support result value from write transaction callbacks ([#294](https://github.com/realm/realm-dart/pull/294/))
+* Added a property `Realm.isInTransaction` that indicates whether the Realm instance has an open write transaction associated with it.
 
 ### Fixed
-* None
+* Fixed an issue that would result in the wrong transaction being rolled back if you start a write transaction inside a write transaction. (Issue [#442](https://github.com/realm/realm-dart/issues/442))
```
blagoev, line 11:
```suggestion
* Fixed an issue that would result in the wrong transaction being rolled back if you start a write transaction inside a write transaction. ([#442](https://github.com/realm/realm-dart/issues/442))
```

lib/src/realm_class.dart
```diff
@@ -164,22 +164,22 @@ class Realm {
...
-  bool get _isInTransaction => realmCore.getIsWritable(this);
+  /// Checks whether the `Realm` is in write transaction.
+  bool get isInTransaction => realmCore.getIsWritable(this);
 
   /// Synchronously calls the provided callback inside a write transaction.
   ///
   /// If no exception is thrown from within the callback, the transaction will be committed.
   /// It is more efficient to update several properties or even create multiple objects in a single write transaction.
   T write<T>(T Function() writeCallback) {
+    final transaction = Transaction._(this);
```
blagoev, line 175: I don't understand why we need the `Transaction` class. The change really seems only to move the beginWrite outside the try block. I would like to understand more so perhaps we could discuss. Added an agenda topic

blagoev (changes requested):

lib/src/realm_class.dart
nielsenko: Eventually we may want an API where the transaction is handed to user code, like some of the other SDKs.

nirinchev: Discussed offline, we're keeping the transaction approach - @blagoev do you have a preference - should I expose it publicly in this PR or should I open a separate ticket for it?

blagoev: I think lets do it in another pr. This one only fixes the nested writes.

blagoev (approved):

PR #452, opened 2022-03-31
email/password provider - register email

### LINKED ISSUE #426: Auth Provider: register email
realm_app_email_password_provider_client_register_email

desistefanova:
Implement EmailPasswordProvider and method registerUser for registering a new user with the given email and password. Fixes #426.


lib/src/native/realm_core.dart
```diff
@@ -850,6 +850,45 @@ class _RealmCore {
...
+    }
+  }
+
+  static void _app_email_password_provider_free_userdata_callback(Pointer<Void> userdata) {
+    final Completer<bool>? completer = userdata.toObject();
+    if (completer == null) {
+      return;
+    }
+    if (!completer.isCompleted) {
+      completer.complete(false);
```
desistefanova, line 875: I'm wandering whether we need this at all.

blagoev (approved):

lib/src/native/realm_core.dart
```diff
@@ -850,6 +850,36 @@ class _RealmCore {
...
+
+  Future<void> appEmailPasswordRegisterUser(Application application, String email, String password) {
+    final completer = Completer<void>();
+    using((arena) {
+      _realmLib.invokeGetBool(() => _realmLib.realm_app_email_password_provider_client_register_email(
+            application.handle._pointer,
+            email.toUtf8Ptr(arena),
+            password.toRealmString(arena).ref,
+            Pointer.fromFunction(void_completion_callback),
+            completer.toGCHandle(),
```
nielsenko, line 877: I would suggest using a hard handle here, as there are no guarantee that the client will ever await the returned future, and hence the local completer may become garbage before the void_completion_callback calls `userdata.toObject()`

nielsenko (changes requested):

lib/src/native/realm_core.dart
blagoev: yes I will be fixing all these calls to use "hard handles"

blagoev: btw, problem is that the future does not hold of the completer at all. So this is definitelly crashing.

blagoev: this should be fixed now

lib/src/native/realm_core.dart
```diff
@@ -888,6 +888,36 @@ class _RealmCore {
...
+
+  Future<void> appEmailPasswordRegisterUser(Application application, String email, String password) {
+    final completer = Completer<void>();
+    using((arena) {
+      _realmLib.invokeGetBool(() => _realmLib.realm_app_email_password_provider_client_register_email(
+            application.handle._pointer,
+            email.toUtf8Ptr(arena),
+            password.toRealmString(arena).ref,
+            Pointer.fromFunction(void_completion_callback),
+            completer.toPersistentHandle(),
```
nirinchev, line 915: That completer should be released if we use persistent handle. My suggestion would be to model methods like this like:

```dart
Future<T> doSomethinAsync(...) {
  final completer = Completer<T>();
  final completerHandle = completer.toPersistentHandle();
  try {
    _realmLib.do_the_async_action(completerHandle);
    final result = await completer.future;
    _realmLib.release_persistent_handle(completerHandle);
    return result;
  } catch (e) {
    _realmLib.release_persistent_handle(completerHandle);
    rethrow;
  }
}
```

nirinchev: Alternatively, we may be able to use a weak handle, as long as we await the future inside the method, thus keeping the completer alive:

```dart
Future<T> doSomethinAsync(...) {
  final completer = Completer<T>();
  _realmLib.do_the_async_action(completer.toWeakHandle());
  return await completer.future;
}
```

nielsenko: Brilliant 👍 .. I wish I had thought of that 🥇

nielsenko: .. actually I'm not sure completer is guaranteed to survive to the end of the scope .. but it looked good at first

blagoev: The completer is deleted in the current code. The _deletePersistentHandlePtr will do that.

blagoev: @nirinchev ^

blagoev: no the completer will not survive if we await there. The future does not hold onto the completer. But since this is an async method I will change it to await anyways

blagoev: I changed my mind and reverted to non async method

nirinchev: I see, I missed the delete userdata function that we're passing.

nielsenko (approved):

PR #454, opened 2022-03-31
Use a helper to create a realm instance

nirinchev:
This takes care of automatically closing and deleting it

### DIFF
--- a/test/test.dart
+++ b/test/test.dart
@@ -104,10 +104,7 @@ Future<void> setupTests(List<String>? args) async {
   await setupBaas();
 
   setUp(() {
-    String path = "${generateRandomString(10)}.realm";
-    if (Platform.isAndroid || Platform.isIOS) {
-      path = _path.join(Configuration.filesPath, path);
-    }
+    final path = generateRandomRealmPath();
     Configuration.defaultPath = path;
 
     addTearDown(() async {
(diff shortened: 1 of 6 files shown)


test/test.dart
```diff
@@ -101,33 +104,41 @@ Future<void> setupTests(List<String>? args) async {
   await setupBaas();
 
   setUp(() {
-    String path = "${generateRandomString(10)}.realm";
-    if (Platform.isAndroid || Platform.isIOS) {
-      path = _path.join(Configuration.filesPath, path);
-    }
+    final path = generateRandomRealmPath();
```
desistefanova, line 107: I think we need Map instead of queue. Having a Map with paths and realm objects will help us to find the realm by path key. Since setUp and TearDown are executed per each test we have the exact path for the current test inside addTearDown. We can `skip` while and `for` statements. Setting `Configuration.defaultPath = path;` bothers me because it is static and would be a problem in parallel execution.

nirinchev: There's no guarantee that in the test we'll be opening a Realm that is using the default path. We're doing that for most local tests, but with sync, we'll be opening random paths for each users, so we need to clean up the entire collection, not just the default path.

I'm not sure how parallel tests are run in dart - are they using the same isolate or different ones? If it's different, then we should be fine as they shouldn't share the `_openRealms` queue.

blagoev (approved):

PR #455, opened 2022-04-01
Delay constructing the config handle

nirinchev:
This is a POC that delays constructing the native config until we actually get to open the Realm instance.

I think this is a cleaner approach for passing more complex data to native. We can use it for all config types (realm, sync, app) as well as for credentials.

### DIFF
--- a/lib/src/configuration.dart
+++ b/lib/src/configuration.dart
@@ -71,19 +71,19 @@ class Configuration {
   /// an [RealmException] will be thrown.
   final int schemaVersion;
 
-  ///The path where the Realm should be stored.
+  /// The path where the Realm should be stored.
   ///
   /// If omitted the [defaultPath] for the platform will be used.
   final String path;
 
-  /// Gets or sets a value indicating whether a [Realm] is opened as readonly.
+  /// Specifies whether a [Realm] should be opened as read-only.
   /// This allows opening it from locked locations such as resources,
   /// bundled with an application.
   ///
   /// The realm file must already exists at [path]
   final bool isReadOnly;
 
-  /// Specifies if a [Realm] should be opened in-memory.
+  /// Specifies whether a [Realm] should be opened in-memory.
   ///
   /// This still requires a [path] (can be the default path) to identify the [Realm] so other processes can open the same [Realm].
   /// The file will also be used as swap space if the [Realm] becomes bigger than what fits in memory,
@@ -91,7 +91,7 @@ class Configuration {
   /// When all in-memory instance of [Realm] is closed all data in that [Realm] is deleted.
   final bool isInMemory;
 
-  /// Gets or sets a value of FIFO special files location.
+  /// Specifies the FIFO special files fallback location.
   /// Opening a [Realm] creates a number of FIFO special files in order to
   /// coordinate access to the [Realm] across threads and processes. If the [Realm] file is stored in a location
   /// that does not allow the creation of FIFO special files (e.g. FAT32 filesystems), then the [Realm] cannot be opened.
--- a/lib/src/configuration.dart
+++ b/lib/src/configuration.dart
@@ -27,7 +27,6 @@ import 'package:path/path.dart' as _path;
 /// Configuration used to create a [Realm] instance
 /// {@category Configuration}
 class Configuration {
-  final ConfigHandle _handle;
   final RealmSchema _schema;
   bool _isInUse = false;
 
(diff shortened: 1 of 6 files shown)


lib/src/native/realm_core.dart
```diff
@@ -205,12 +167,9 @@ class _RealmCore {
...
   }
 
-  void setScheduler(Configuration config, SchedulerHandle schedulerHandle) {
-    _realmLib.realm_config_set_scheduler(config.handle._pointer, schedulerHandle._pointer);
-  }
-
-  RealmHandle openRealm(Configuration config) {
-    final realmPtr = _realmLib.invokeGetPointer(() => _realmLib.realm_open(config.handle._pointer), "Error opening realm at path ${config.path}");
+  RealmHandle openRealm(Configuration config, Scheduler scheduler) {
+    final configHandle = _createConfig(config, scheduler.handle);
```
desistefanova, line 171: I agree. And also, creating a new `Config` each time we open a `Realm `would solve the issue with shared scheduler and we wont need anymore `config.isInUse` we wont need to throw `RealmStateError("A Realm instance for this configuraiton object already exists.");` But keep in mind that all the members in Configuration have to be `final` since setting them later will cause no update in the native config.

lib/src/configuration.dart
```diff
@@ -84,45 +65,34 @@ class Configuration {
...
   /// The schema version used to open the [Realm]
   ///
   /// If omitted the default value of `0` is used to open the [Realm]
   /// It is required to specify a schema version when initializing an existing
   /// Realm with a schema that contains objects that differ from their previous
   /// specification. If the schema was updated and the schemaVersion was not,
   /// an [RealmException] will be thrown.
-  int get schemaVersion => realmCore.getSchemaVersion(this);
-  set schemaVersion(int value) => realmCore.setSchemaVersion(this, value);
+  int schemaVersion;
```
nielsenko, line 75:
```suggestion
  final int schemaVersion;
```
and initialize in ctor .. and use the @immutable annotation from `meta` package.

lib/src/native/realm_core.dart
desistefanova: Also, can we also move the scheduler under the configuration class? In case we have only one realm opened per config it wont be a problem. On this way we will be consistent with C-API where the scheduler is part of the configuration.

blagoev: lets keep the scheduler in `realm_class`. now configuration is just a data bag.

lib/src/configuration.dart
```diff
@@ -84,45 +62,34 @@ class Configuration {
...
   ///
   /// This still requires a [path] (can be the default path) to identify the [Realm] so other processes can open the same [Realm].
   /// The file will also be used as swap space if the [Realm] becomes bigger than what fits in memory,
   /// but it is not persistent and will be removed when the last instance is closed.
   /// When all in-memory instance of [Realm] is closed all data in that [Realm] is deleted.
-  bool get isInMemory => realmCore.getConfigInMemory(this);
-  set isInMemory(bool value) => realmCore.setConfigInMemory(this, value);
+  final bool isInMemory;
 
   /// Gets or sets a value of FIFO special files location.
```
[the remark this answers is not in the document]
blagoev:
```suggestion
  /// Gets the FIFO special files location.
```

lib/src/configuration.dart
```diff
@@ -84,45 +62,34 @@ class Configuration {
...
+  final int schemaVersion;
 
   ///The path where the Realm should be stored.
   ///
   /// If omitted the [defaultPath] for the platform will be used.
-  String get path => realmCore.getConfigPath(this);
-  set path(String value) => realmCore.setConfigPath(this, value);
+  final String path;
 
   /// Gets or sets a value indicating whether a [Realm] is opened as readonly.
```
blagoev, line 79:
```suggestion
  /// Gets a value indicating whether a [Realm] is opened as read-only.
```

blagoev (changes requested):

lib/src/configuration.dart
```diff
@@ -84,45 +62,34 @@ class Configuration {
     return "";
   }
 
-  @override
-  bool operator ==(Object other) {
```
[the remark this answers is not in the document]
nirinchev: Since now configs are immutable, I'm not sure what a meaningful override of the equals operator would do beyond check for referential equality.

nirinchev:
@blagoev updated API docs, but I'm not sure I have an idea for how a meaningful equality comparer would look like on the config, especially once we introduce callbacks for migration, compaction, and initial data.

@desistefanova I tend to agree with Lubo that the scheduler should be separate from the config, but it's possible I'm not seeing some elegant solution there. Feel free to ping me on slack if you feel strongly about it and we can look at it together.

lib/src/configuration.dart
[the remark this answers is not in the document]
blagoev: I think keeping the same implementation will do. no?

desistefanova (approved):

lib/src/configuration.dart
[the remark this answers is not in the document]
blagoev: Oh I see what you mean. If we remove the realmCore call then we already have the implementation we need.

blagoev (approved):

PR #456, opened 2022-04-01
Gc handles

nielsenko:
Using gc-handles instead of per-function trampolines.


fealebenpae (approved):
I like PRs that delete more code than they add 😁 
This makes complete sense to me, and it's going to make writing new callback-heavy code much nicer. However, should we add an option to select between a weak or a strong reference?

lib/src/native/realm_core.dart
```diff
@@ -554,67 +553,79 @@ class _RealmCore {
...
+        return;
+      }
 
-    try {
-      final changesHandle = RealmObjectChangesHandle._(data);
-      controller.onChanges(changesHandle);
-    } catch (e) {
-      controller.onError(RealmError("Error handling collection change notifications. Error: $e"));
+      try {
+        final changesHandle = RealmObjectChangesHandle._(_realmLib.realm_clone(data.cast()).cast());
```
blagoev, line 583: realm_clone should be invoked with tryGetPointer and could return `null` which we should check

lib/src/native/realm_core.dart
```diff
@@ -554,67 +553,79 @@ class _RealmCore {
...
+        return;
+      }
 
-    try {
-      final changesHandle = RealmCollectionChangesHandle._(data);
-      controller.onChanges(changesHandle);
-    } catch (e) {
-      controller.onError(RealmError("Error handling collection change notifications. Error: $e"));
+      try {
+        final changesHandle = RealmCollectionChangesHandle._(_realmLib.realm_clone(data.cast()).cast());
```
blagoev, line 565: realm_clone result should be checked. See my other comment on that

lib/src/native/realm_core.dart
```diff
@@ -554,67 +553,79 @@ class _RealmCore {
...
-    final controller = object as NotificationsController;
-
-    if (data == nullptr) {
-      //realm_collection_changes data clone is done in native code before this callback is invoked. nullptr data means cloning failed.
-      controller.onError(RealmError("Invalid notifications data received"));
-      return;
-    }
+  static void collection_change_callback(Pointer<Void> userdata, Pointer<realm_collection_changes> data) {
+    final controller = _realmLib.gc_handle_deref(userdata);
+    if (controller is NotificationsController) {
```
blagoev, line 558: previously we were debug asserting this is a `NotificationController` instance

src/realm_dart.cpp
```diff
@@ -60,17 +60,84 @@ void realm_android_dummy();
...
+
+RLM_API void* gc_handle_hard_new(Dart_Handle handle) {
+    return new GCHandle(handle, true);
+}
+
+RLM_API void gc_handle_soften(void* handle) {
+    return reinterpret_cast<GCHandle*>(handle)->soften();
+}
+
+RLM_API Dart_Handle gc_handle_deref(void* handle) {
```
blagoev, line 141:
```suggestion
RLM_API Dart_Handle gc_handle_to_object(void* handle) {
```

blagoev: A small rename to better reflect the action's result and intent

src/realm_dart.cpp
```diff
@@ -60,17 +60,84 @@ void realm_android_dummy();
...
+    }
+    */
+
+    static void finalize_handle(void* isolate_callback_data, void* peer) {
+        delete reinterpret_cast<GCHandle*>(peer);
+    }
+
+    // TODO: HACK. Should be Dart_WeakPersistentHandle when hack removed
+    Dart_FinalizableHandle m_weakHandle;
+    Dart_PersistentHandle m_hardHandle; // used to pin dart object from native side, if needed
```
blagoev, line 126:
```suggestion
```

lib/src/native/realm_core.dart
```diff
@@ -554,67 +553,79 @@ class _RealmCore {
...
-    final controller = object as NotificationsController;
-
-    if (data == nullptr) {
-      //realm_collection_changes data clone is done in native code before this callback is invoked. nullptr data means cloning failed.
-      controller.onError(RealmError("Invalid notifications data received"));
-      return;
-    }
+  static void object_change_callback(Pointer<Void> userdata, Pointer<realm_object_changes> data) {
+    final controller = _realmLib.gc_handle_deref(userdata);
+    if (controller is NotificationsController) {
```
blagoev, line 575: We are removing the debug assertion for `NotificationsController` instance

blagoev (changes requested):

lib/src/native/realm_core.dart
```diff
@@ -554,67 +553,94 @@ class _RealmCore {
...
+      final clonedData = _realmLib.realm_clone(data.cast());
+      if (clonedData == nullptr) {
+        controller.onError(RealmError("Error while cloning notifications data"));
+        return;
+      }
+
+      final changesHandle = RealmObjectChangesHandle._(clonedData.cast());
       controller.onChanges(changesHandle);
     } catch (e) {
       controller.onError(RealmError("Error handling collection change notifications. Error: $e"));
```
desistefanova, line 602:
```suggestion
      controller.onError(RealmError("Error handling object change notifications. Error: $e"));
```

PR #459, opened 2022-04-06
Allow nullable PKs

### LINKED ISSUE #458: Primary keys should allow to be nullable
Right now we prevent primary keys from being nullable, but there's no reason for this. We should allow it.

nirinchev:
Fixes #458


nielsenko (approved):

blagoev (changes requested):
I would like to have a discussion about this offline.

nirinchev:
@blagoev did you get a chance to get buy-in from @ianpward on not supporting nullable PKs in the dart SDK? If so, I propose we close the PR and we can always reopen if user requests come in.

blagoev:
I converted to draft. If we close it someone might delete the branch. Let's keep it for a while. I know dangling PRs are not great but...

PR #460, opened 2022-04-06
[superseded] Add support for ObjectId

nirinchev:
Fixes https://github.com/realm/realm-dart/issues/461.

### DIFF
--- a/common/pubspec.yaml
+++ b/common/pubspec.yaml
@@ -1,5 +1,5 @@
 name: realm_common
-description: >- 
+description: >-
              Hosts the common code shared between realm, realm_dart and realm_generator packages.
              This package is part of the official Realm Flutter and Realm Dart SDKs.
 
@@ -8,6 +8,9 @@ version: 0.2.1+alpha
 homepage: https://www.realm.io
 repository: https://github.com/realm/realm-dart
 issue_tracker: https://github.com/realm/realm-dart/issues
+
+dependencies:
+  objectid: ^2.0.0
 
 environment:
   sdk: '>=2.15.0 <3.0.0'
--- a/generator/test/good_test_data/all_types.expected
+++ b/generator/test/good_test_data/all_types.expected
@@ -1,0 +1,232 @@
+// **************************************************************************
+// RealmObjectGenerator
+// **************************************************************************
+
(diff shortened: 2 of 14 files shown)


common/pubspec.yaml
```diff
@@ -9,6 +9,9 @@ homepage: https://www.realm.io
 repository: https://github.com/realm/realm-dart
 issue_tracker: https://github.com/realm/realm-dart/issues
 
+dependencies:
+  objectid: ^2.0.0
```
nielsenko, line 13: Nice. I thought we had to roll over own.

lib/src/native/realm_core.dart
```diff
@@ -854,7 +863,13 @@ extension on Pointer<realm_value_t> {
...
         throw Exception("Not implemented");
       case realm_value_type.RLM_TYPE_OBJECT_ID:
-        throw Exception("Not implemented");
+        final nativeBytes = ref.values.object_id.bytes;
+        final bytes = List<int>.filled(12, 0);
+        for (int i = 0; i < 12; i++) {
+          bytes[i] = nativeBytes[i];
+        }
+
+        return ObjectId.fromBytes(bytes);
```
nielsenko, line 872:
```suggestion
        return ObjectId.fromBytes(cast<Uint8>().asTypedList(12));
```
The `ObjectId.fromBytes` ctor already copies the bytes. Given the memory layout of the `realm_value_t` we can just read the first 12 bytes.

nielsenko (approved):

generator/test/good_test_data/all_types.expected
```diff
@@ -0,0 +1,232 @@
+// **************************************************************************
```
nirinchev, line 1: This is not strictly related to this PR, but I saw we don't validate the outputs of the `all_types.dart` test, which wasn't great as it would only validate that it doesn't error out, but not that it produces meaningful results. I wanted to verify the outputs for the new type, so I had to add the entire file.

blagoev:
This was superseded by #468

blagoev:
reopened since this has nullable objectid support

nirinchev:
Covered by https://github.com/realm/realm-dart/pull/851

PR #462, opened 2022-04-06
[superseded] Add support for uuids

nirinchev:
Fixes https://github.com/realm/realm-dart/issues/463.

### DIFF
--- a/common/lib/src/realm_types.dart
+++ b/common/lib/src/realm_types.dart
@@ -19,6 +19,7 @@
 import 'dart:ffi';
 import 'dart:typed_data';
 import 'package:objectid/objectid.dart';
+import 'package:uuid/uuid.dart';
 
 /// All supported `Realm` property types.
 /// {@category Configuration}
@@ -79,9 +80,6 @@ class RealmStateError extends StateError implements RealmError {
 }
 
 /// @nodoc
-class Uuid {} // TODO!
-
-/// @nodoc
 class Decimal128 {} // TODO!
 
 /// @nodoc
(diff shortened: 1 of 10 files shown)


common/lib/src/realm_types.dart
```diff
@@ -19,6 +19,7 @@
 import 'dart:ffi';
 import 'dart:typed_data';
 import 'package:objectid/objectid.dart';
+import 'package:uuid/uuid.dart';
```
nielsenko, line 22: I'm not too keen on this package. Try looking into the ctor `UuitValue.fromList` it basically converts the bytes to a `String` and the parses it back. I think that is unreasonably slow for our use.

lib/src/native/realm_core.dart
```diff
@@ -871,7 +880,13 @@ extension on Pointer<realm_value_t> {
...
         return ObjectId.fromBytes(bytes);
       case realm_value_type.RLM_TYPE_UUID:
-        throw Exception("Not implemented");
+        final nativeBytes = ref.values.uuid.bytes;
+        final bytes = List<int>.filled(16, 0);
+        for (int i = 0; i < 16; i++) {
+          bytes[i] = nativeBytes[i];
+        }
+
+        return UuidValue.fromList(bytes);
```
nielsenko, line 889:
```suggestion
        return UuidValue.fromList(cast<Uint8>().asTypedList(16));
```

nielsenko: Should be enough, given memory layout.. but the `UuidValue.fromList` ctor is gross

common/lib/src/realm_types.dart
nielsenko: We talked about this. There is no denying that the [uuid](https://pub.dev/packages/uuid) package is the more popular package. However its internal state is a string, which is pretty bad for us (really anybody), since we need to convert going back and forth.

However, we have options. One being simply implementing the interface of `UuidValue` in an efficient manner, at some point down the line.

nirinchev: Discussed offline - we agree that the package as it is is not great, but it's likely to be widely used and the functionality is there, even if the actual implementation is not super performant. We agreed to go with the existing package and try to submit a PR to the maintainer that optimizes the UuidValue construction.

blagoev:
closing this in favor of #470

PR #465, opened 2022-04-08
Move baas tests to QA

nirinchev:
Run baas tests against QA rather than dev


.github/workflows/pr.yml
```diff
@@ -40,9 +40,9 @@ jobs:
           dart run realm_dart deploy-apps \
             --baas-url https://realm-dev.mongodb.com \
```
nielsenko, line 41: Is the url not different for QA?

nirinchev: Ah... forgot that I had hardcoded it - will update

blagoev (approved):

PR #468, opened 2022-04-12
add support for objectId

### LINKED ISSUE #461: Add support for ObjectId
We should support ObjectId both as a regular property and as a primary key.

blagoev:
support objectId
use custom line by line matcher for generator tests
this supersedes https://github.com/realm/realm-dart/pull/460
fixes #461

### DIFF
--- a/generator/test/good_test_data/all_types.expected
+++ b/generator/test/good_test_data/all_types.expected
@@ -1,0 +1,221 @@
+// **************************************************************************
+// RealmObjectGenerator
+// **************************************************************************
+
+class Fooo extends _Foo with RealmEntity, RealmObject {
+  static var _defaultsSet = false;
+
+  Fooo({
+    int x = 0,
+  }) {
+    if (!_defaultsSet) {
+      _defaultsSet = RealmObject.setDefaults<Fooo>({
+        'x': 0,
+      });
+    }
+    RealmObject.set(this, 'x', x);
+  }
+
+  Fooo._();
+
+  @override
+  int get x => RealmObject.get<int>(this, 'x') as int;
+  @override
+  set x(int value) => RealmObject.set(this, 'x', value);
+
+  @override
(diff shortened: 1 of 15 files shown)


generator/test/test_util.dart
```diff
@@ -14,30 +15,84 @@ Map<String, String> getListOfTestFiles(String directory) {
...
+      if (expectedLines[i] != actualLines[i]) {
+        matchState["Error"] = "Difference at line ${i+1}. \nExpected: ${expectedLines[i]}.\n  Actual: ${actualLines[i]}";
+        return false;
+      }
+    }
+
+    return true;
+    }
+    catch (_) {
+      print("fali");
```
desistefanova, line 76:
```suggestion
      print("fail");
```

blagoev: this is debug left over. will remove

desistefanova (approved):

generator/test/good_test_data/all_types.expected
```diff
@@ -0,0 +1,221 @@
+// **************************************************************************
+// RealmObjectGenerator
+// **************************************************************************
+
+class Fooo extends _Foo with RealmEntity, RealmObject {
```
nirinchev, line 5: Why does this have 3 `o`-s? Shouldn't the generated class be called `Foo`?

blagoev: We support a map to directive on the model class. Which in  this case is like 
@RealmModel()
**@MapTo('Fooo')**
class _Foo {
  int x = 0;
}

blagoev: I agree this exact sample is not ideal. should have been named differently.

blagoev: I pushed a change on that name to avoid the confusion https://github.com/realm/realm-dart/commit/8d04e3f69d9392d9e090b5db9944f7a3d9ae172c

PR #469, opened 2022-04-12
application login user

### LINKED ISSUE #407: App: login user
realm_app_log_in_with_credentials

desistefanova:
Application login user. Fixes #407 
Fixes http transport headers enumeration


lib/src/application.dart
```diff
@@ -109,10 +111,26 @@ class ApplicationConfiguration {
...
 /// * Synchronize data between the local device and a remote Realm App with Synchronized Realms
 class Application {
-  final AppHandle _handle;
+  late final AppHandle _handle;
   final ApplicationConfiguration configuration;
-
-  Application(this.configuration) : _handle = realmCore.getApp(configuration);
+  late final RealmHttpTransportHandle _httpTransportHandle;
+  // ignore: unused_field
+  late final AppConfigHandle _appConfigHandle;
```
nielsenko, line 120: I prefer the previous approach, where only the `AppHandle` and `Configuration` is stored, and the `app_config` and `sync_client_config` is created on demand in `realmCore.getApp`.

lib/src/native/realm_core.dart
```diff
@@ -1123,10 +1165,10 @@ extension on Pointer<IntPtr> {
 }
 
 extension on Pointer<Void> {
-  T? toObject<T extends Object>() {
+  T? toObject<T extends Object>([bool isPersistent = false]) {
```
nielsenko, line 1168: I think an optional `bool` argument is a bit secret here.. Maybe make it named, or have to different methods for weak and persistent.

lib/src/native/realm_core.dart
```diff
@@ -1123,10 +1165,10 @@ extension on Pointer<IntPtr> {
...
     assert(this != nullptr, "Pointer<Void> is null");
 
-    final object = _realmLib.gc_handle_to_object(this);
+    Object object = isPersistent ? _realmLib.persistent_handle_to_object(this) : _realmLib.weak_handle_to_object(this);
```
nielsenko, line 1171: I prefer the previous approach, where you don't have to distinguish.

scripts/build-linux.sh
```diff
@@ -10,7 +10,7 @@ mkdir -p build-linux
 pushd build-linux
 
 cmake -GNinja \
-      -DCMAKE_BUILD_TYPE=MinSizeRel \
+      -DCMAKE_BUILD_TYPE=Debug \
```
nielsenko: This was probably not intended

test/test.dart
```diff
@@ -194,4 +196,34 @@ Future<void> setupBaas() async {
...
-}
\ No newline at end of file
+}
+
+@isTest
+Future<void> syncTest(
+  String name,
+  FutureOr<void> Function(ApplicationConfiguration configuration) testFunction, {
+  String appName = 'flexible',
+  dynamic skip,
```
nielsenko, line 206: I know `test` has `skip` as `dynamic`, but why repeat that mistake. Let us make it a `bool`.

test/test.dart
```diff
@@ -194,4 +196,34 @@ Future<void> setupBaas() async {
...
+}) async {
+  final url = Uri.tryParse(Platform.environment['BAAS_URL'] ?? 'https://realm-dev.mongodb.com');
+  final apiKey = Platform.environment['BAAS_API_KEY'];
+  final projectId = Platform.environment['BAAS_PROJECT_ID'];
```
nielsenko, line 210: These are not needed. There must have been some merge mistake along the way..

lib/src/native/realm_core.dart
```diff
@@ -843,12 +849,44 @@ class _RealmCore {
     });
   }
 
-  AppHandle getApp(ApplicationConfiguration configuration) {
-    final httpTransport = createHttpTransport(configuration.httpClient);
-    final appConfig = createAppConfig(configuration, httpTransport);
-    final syncClientConfig = createSyncClientConfig(configuration);
-    final realm_app = _realmLib.invokeGetPointer(() => _realmLib.realm_app_get(appConfig._pointer, syncClientConfig._pointer));
-    return AppHandle._(realm_app);
+  AppHandle getApp(AppConfigHandle appConfigHandle, SyncClientConfigHandle syncClientConfigHandle) {
```
nirinchev, line 852: Let's discuss this at the roundtable, but I think we should return the previous form of this method.

blagoev: I have added an agenda item

nielsenko (commented):
Let us discuss further at the round-table tomorrow

test/test.dart
blagoev: it can be a bool or a String that's why it is dynamic

test/test.dart
```diff
@@ -194,4 +196,34 @@ Future<void> setupBaas() async {
   final client = await (cluster == null ? BaasClient.docker(baasUrl) : BaasClient.atlas(baasUrl, cluster, apiKey!, privateApiKey!, projectId!));
 
   baasApps.addAll(await client.getOrCreateApps());
-}
\ No newline at end of file
+}
+
+@isTest
+Future<void> syncTest(
```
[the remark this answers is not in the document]
blagoev: that's an implementation detail in that name. This why I think syncTest is better

src/realm_dart.cpp
```diff
@@ -112,17 +112,32 @@ class GCHandle {
...
+
+RLM_API Dart_Handle persistent_handle_to_object(void* handle) {
+    Dart_PersistentHandle persistentHandle = reinterpret_cast<Dart_PersistentHandle>(handle);
+    return Dart_HandleFromPersistent_DL(persistentHandle);
+}
+
+RLM_API void delete_persistent_handle(void* handle) {
+    Dart_PersistentHandle persistentHandle = reinterpret_cast<Dart_PersistentHandle>(handle);
+    Dart_DeletePersistentHandle_DL(persistentHandle);
+}
```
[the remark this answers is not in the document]
blagoev: we don't have these functions exposed from the binary. Also this allows us to change the implementation if needed later.

scripts/build-linux.sh
blagoev: this should be fixed already

lib/src/native/realm_core.dart
blagoev: I refactored it since persistence handles are different than weak handles. Do you see an issue with the current approach?

blagoev: Also not sure what you mean by "don't have to distinguish". Previously we had the same distinction in another place

lib/src/native/realm_core.dart
blagoev: I was thinking of that myself. Still a single named argument seems too verbose and also this is not a public API but one for internal use. I think we should not be using named arguments by default. 
A different method will be better but extensions in the same file can not have private members as there aren't any so the same implementation would be repeated twice.

test/test.dart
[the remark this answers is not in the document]
nirinchev: baas is intentionally called like that to be a generic name unrelated to the marketing name of the product. Using `testWithBaas` or `baasTest` if you prefer a shorter name is better because it covers all the backend functionality, not just sync. Auth, function, and remote mongo tests will not use anything sync-related, but will need to run against a baas instance.

lib/src/native/realm_core.dart
nirinchev: I am a big proponent of using named arguments for bools.

lib/src/user.dart
```diff
@@ -0,0 +1,36 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////////
+
+import 'native/realm_core.dart';
+
+/// This class represents a user in a MongoDB Realm app.
+/// A user can log in to the server and, if access is granted, it is possible to synchronize the local and the remote Realm.
```
nirinchev, line 22: I know this is copied from the .NET docstring, but it is no longer correct - i.e. there are no remote Realms

```suggestion
/// A user can log in to the server and, if access is granted, it is possible to synchronize the local Realm to MongoDB.
```

test/test.dart
```diff
@@ -194,4 +196,34 @@ Future<void> setupBaas() async {
...
+  FutureOr<void> Function(ApplicationConfiguration configuration) testFunction, {
+  String appName = 'flexible',
+  dynamic skip,
+}) async {
+  final url = Uri.tryParse(Platform.environment['BAAS_URL'] ?? 'https://realm-dev.mongodb.com');
+  final apiKey = Platform.environment['BAAS_API_KEY'];
+  final projectId = Platform.environment['BAAS_PROJECT_ID'];
+
+  if (skip == null) {
+    skip = url == null || apiKey == null || projectId == null;
```
nirinchev, line 213: How can `url` be null here?

test/test.dart
```diff
@@ -194,4 +196,34 @@ Future<void> setupBaas() async {
...
+  }
+  else if (skip is bool) {
```
nirinchev, line 215: What is `skip` is a string?

test/test.dart
```diff
@@ -194,4 +196,34 @@ Future<void> setupBaas() async {
...
+  test(name, () async {
+      final app = baasApps[appName] ?? baasApps.values.first;
+      final temporary = await Directory.systemTemp.createTemp('realm_test_');
+      final configuration = ApplicationConfiguration(
+        app.clientAppId,
+        baseUrl: url,
+        baseFilePath: temporary,
+      );
+      return await testFunction(configuration);
+  }, skip: skip);
```
nirinchev, line 228: If baas is not setup, should we set `skip` to `"Baas is not setup"` or something similar to make it explicit why the test was skipped?

nirinchev (commented):
First pass - we probably need a lot more tests and also to expose the properties on the user, but I guess those can come as a follow-up PR.

lib/src/native/realm_core.dart
nielsenko: Previously `gc_handle_deref` worked for both weak and hard handles

test/test.dart
nielsenko: How would you call skip with a string?

blagoev: you pass a string in the skip arg. This is how the test package works

test/test.dart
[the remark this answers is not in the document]
blagoev: lets go with baasTest

test/test.dart
blagoev: its passed to the test package

test/test.dart
blagoev: not sure we should be overwritting that

test/test.dart
blagoev: Uri.tryParse returns nullable

test/test.dart
blagoev: its passed to the test package test method as is. Meaning skip is true (skip should happen) and no need to check for url

test/test.dart
blagoev: ok I will use a string skip in this case

test/test.dart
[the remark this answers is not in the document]
nirinchev: I'd strongly recommend keeping the baas terminology for consistency's sake. Both the actions that spin up a cluster are using baas as terminology, as well as the dart client that configures the app is called BaasClient. Additionally, the repo of the server is called baas and all SDKs are (should) be using baas when referring to the server. There's no reason to be inconsistent about it.

test/test.dart
nirinchev: Ah yeah - I got confused by the fact we provide a default value, but you're right that the developer can still set `BAAS_URL=foo` when running the tests.

nielsenko (approved):

PR #473, opened 2022-04-17
Get current user

### LINKED ISSUE #406: App: get current user
realm_app_get_current_user

blagoev:
fixes #406


lib/src/native/realm_core.dart
```diff
@@ -1010,6 +1010,15 @@ class _RealmCore {
     });
     return completer.future;
   }
+  
+  UserHandle? getCurrentUser(AppHandle appHandle) {
+    final userPtr = _realmLib.realm_app_get_current_user(appHandle._pointer);
```
nielsenko, line 1015: We can't use `_realmLib.invokeGetPointer` here, but should we check for error somehow?

nielsenko (approved):

lib/src/native/realm_core.dart
blagoev: Some functions that return pointers can't fail. This one and http_transport_new are such example. We just use the function directly in these cases.

PR #477, opened 2022-04-19
support app get all logged in users

### LINKED ISSUE #405: App: get all users
realm_app_get_all_users

blagoev:
fixes #405


nielsenko (approved):

lib/src/native/realm_core.dart
```diff
@@ -1055,6 +1055,29 @@ class _RealmCore {
...
     _realmLib.realm_clear_cached_apps();
   }
+
+  List<UserHandle> getUsers(App app) {
+    return using((arena) {
+      final usersCount = arena<IntPtr>();
+      _realmLib.invokeGetBool(() => _realmLib.realm_app_get_all_users(app.handle._pointer, nullptr, 0, usersCount));
+
+      final usersPtr = arena<Pointer<realm_user>>(usersCount.value);
+      _realmLib.invokeGetBool(() => _realmLib.realm_app_get_all_users(
```
nirinchev, line 1065: This should handle the case where a user is added/removed between the two calls. Right now, there's a race condition.

nirinchev: This is actually a bug in Core: https://github.com/realm/realm-core/issues/5430.

PR #480, opened 2022-04-19
Emailpassword reset password

### LINKED ISSUE #431: Auth Provider: reset password

blagoev:
closes #431


nielsenko (approved):

lib/src/credentials.dart
```diff
@@ -85,4 +85,9 @@ class EmailPasswordAuthProvider {
   Future<void> resendUserConfirmation(String email) {
     return realmCore.emailPasswordResendUserConfirmation(application, email);
   }
+
+  /// Completes the reset password procedure by providing the desired new password.
+  Future<void> resetPassword(String password, String token, String tokenId) {
```
nirinchev, line 90: I think this should have a little more detailed docs.

nirinchev (approved):

lib/src/credentials.dart
blagoev: That's a copy of what .NET has as api docs

blagoev: what else should we mention here?

nirinchev: The .NET API docs are more detailed: https://github.com/realm/realm-dotnet/blob/ce874c4127cacd575538dda7a62aeadf66265473/Realm/Realm/Sync/App.cs#L372-L386. 

Notably, we're missing descriptions for the parameters - it's probably not clear to a user what the token and tokenId mean in this case and how to get them.

PR #483, opened 2022-04-19
support calling custom reset password functions

### LINKED ISSUE #482: Auth Provider: support custom confirmation functions

blagoev:
closes #482


lib/src/credentials.dart
```diff
@@ -95,4 +95,9 @@ class EmailPasswordAuthProvider {
   Future<void> resetPassword(String email) {
     return realmCore.emailPasswordResetPassword(application, email);
   }
+
+  /// Calls the reset password function, configured on the server.
+  Future<void> callResetPasswordFunction(String email, String password, String functionArgsAsJSON) {
```
nirinchev, line 100: Just when I thought I'm starting to like dart 🤮

nirinchev (approved):

lib/src/credentials.dart
nielsenko: Normally Json objects are modelled as `Map<String, dynamic>` in dart. I know core takes a string, but perhaps we should hide that. Even better we could use a typed object that we serialize to Json (similar to how we do it with metrics).

nielsenko (changes requested):

PR #485, opened 2022-04-19
rename app

blagoev:
rename Application to App, rename ApplicationConfiguration to AppConfiguration


lib/src/application.dart
```diff
@@ -35,8 +35,8 @@ enum MetadataPersistenceMode {
 
 @immutable
 
-/// A class exposing configuration options for an [Application]
-class ApplicationConfiguration {
+/// A class exposing configuration options for an [App]
+class AppConfiguration {
```
nielsenko, line 39: Looks okay to me.. one thing I have been wondering about.. Why did you drop the `application_configuration.dart` file and stuff everything in `application.dart`?

blagoev: its common in dart to use one file for multiple classes. No need to split these into separate files.

nielsenko (approved):

PR #490, opened 2022-04-20
Modify MapTo on classes to work the same way as on properties

nirinchev:
This turned out to be fairly involved, but the current set of changes makes @MapTo on objects behave the same way as on properties - it changes the core names, but keeps the dart name. It also fixes a number of bugs that existed

* Fixes setting mapped properties in the ctor - previously we would use the old property names.
* Fixes links (both lists and single) to objects with remapped names.

Additionally, this upgrades to latest core because I needed https://github.com/realm/realm-core/pull/5415.


generator/lib/src/class_element_ex.dart
```diff
@@ -159,10 +129,27 @@ extension ClassElementEx on ClassElement {
         );
       }
 
+      final coreClassName = mapToInfo != null ? mapToInfo!.value.getField('name')!.toStringValue()! : name;
+
+      // Core has a limit of 57 characters for SDK names (technically 63, but SDKs names are always preffixed class_)
```
nielsenko, line 134:
```suggestion
      // Core has a limit of 57 characters for SDK names (technically 63, but SDKs names are always prefixed class_)
```
Nice check!

generator/lib/src/class_element_ex.dart
```diff
@@ -159,10 +129,27 @@ extension ClassElementEx on ClassElement {
         );
       }
 
+      final coreClassName = mapToInfo != null ? mapToInfo!.value.getField('name')!.toStringValue()! : name;
```
nielsenko, line 132: Maybe call it `realmName` to be consistent with `RealmModelInfo` or vice versa

nielsenko (approved):
I think you forgot to update `main.g.dart`, causing a test failure, but otherwise it looks kosher

nirinchev:
@nielsenko while adding more test cases, I uncovered another corner case that was previously unhandled and that turned out to be a fair number of changes, so please take another look to make sure I haven't missed something.

generator/lib/src/element.dart
```diff
@@ -104,7 +104,20 @@ extension ElementEx on Element {
...
+  String? get remappedRealmName {
+    AnnotationValue? mapTo;
+    try {
+      if (getDeclarationFromElement(this) != null) {
+        mapTo = annotationInfoOfExact(mapToChecker);
+      }
+    } on ArgumentError {
+      // This will happen for types defined in a different library
+      // In this case, fall back to not using the computed value
+      mapTo = mapToChecker.annotationsOfExact(this).singleOrNull as AnnotationValue?;
```
nielsenko, line 116: Could we add a test for cross-file links with mapped class name

nielsenko (approved):

blagoev (changes requested):
Seems good. Reviewing as red cause of the RealmObjectInternal usage in the tests.

test/realm_object_test.dart
```diff
@@ -314,4 +322,40 @@ Future<void> main([List<String>? args]) async {
...
+    final obj = realm.write(() {
+      final obj = realm.add(RemappedClass("some value"));
+      obj.listProperty.add(obj);
+      return obj;
+    });
+
+    final json = obj.toJson();
+
+    // remappedProperty is mapped as `__ other property __`
+    expect(json, contains('"__ other property __":"some value"'));
```
blagoev, line 339: I suggest we use better name for these. This code reads like a generated one.

nirinchev: I wanted to try using some crazy names, but can make them more sensible.

test/realm_object_test.dart
```diff
@@ -20,7 +20,8 @@
 
 import 'dart:io';
 import 'package:test/test.dart' hide test, throws;
-import '../lib/realm.dart';
+import '../lib/realm.dart' hide RealmObject;
+import '../lib/src/realm_object.dart' show RealmObject, RealmObjectInternal;
```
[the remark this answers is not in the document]
nirinchev: We're not exposing `toJson` publicly because it's not a sensible json representation of the object, but I need it for the tests because that's the only way to verify that what Core sees is the right thing.

lib/src/realm_object.dart
```diff
@@ -297,6 +299,8 @@ extension RealmObjectInternal on RealmObject {
 
   RealmObjectHandle get handle => _handle!;
   RealmAccessor get accessor => _accessor;
+
+  String toJson() => realmCore.objectToString(this);
```
blagoev, line 303: moving this to test.dart since this is the only place using it. Since every tests imports test.dart this method will be available in the other test files as well. test.dart will need to import realm_core.dart

blagoev: I mean code in `lib` should really be code used in the SDK. If tests need to call core we should do it there in such cases. Its not ideal since until now it was not needed but we need it now for `toJson` and it seems for clear app cache as well

test/realm_object_test.dart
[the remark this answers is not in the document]
blagoev: these are not needed now

blagoev (approved):
Just remove the RealmObjectInternal in object tests. It should not be needed now.

blagoev:
Also we should either ask core to create a release or we should add a task to update to released core version later. we should be careful on depending on some commit in master cause we might release with that. In general up until now we never point to not released commits

PR #495, opened 2022-04-25
Expose dynamic Realm API

nirinchev:
This exposes a set of string-based access API for the Realm. The current PR touches only the Realm entrypoint and a follow-up PR will be done to improve the experience when working with objects. The API is as follows

```dart
// Lookup all objects of type Car
final allCars = realm.dynamic.all('Car');

// The collection can be further filtered using the normal API
final carsFromJapan = allCars.query('manufacturer.domicile == "Japan"');

// Lookup an object by PK
final honda = realm.dynamic.find('Car', 'Honda');
```

Note: One can use the `RealmObject.get<...>(...)` API to then read properties on the object, but those will change in the future, so not showing them in the example.

This is a part of https://github.com/realm/realm-dart/issues/70.

Fixes https://github.com/realm/realm-dart/issues/690.


nielsenko (commented):
All tests pass, and we have a string based interface. However, I think the `DynamicRealm` should traffic in `dynamic`s with runtime type `DynamicRealmObject` and that these should overload `noSuchMethod` to access the right properties.

nirinchev:
There are three reasons why I didn't go for `dynamic`.
1. We've been burned by dynamics in the .NET SDK. I know that dart doesn't yet have anything Unity-like in its ecosystem, but I would imagine MS didn't envision Unity either when they implemented `dynamic` in C#.
2. The string-based API is slightly more discoverable as you get "intellisense" for it. So you have at least some pointers on what to do.
3. It is more flexible. You can for example read the property information from the schema and then access properties by their name - e.g.
    ```dart
    for (final prop in properties) {
        if (prop.type == RealmPropertyType.int) {
            final intValue = obj.dynamic.get<int>('prop.name');
            // Do something with intValue
        }
    }
    ```
    You can't achieve something like that (or at least I can't see how) with the dynamic type in dart. It's not often, but we have had a handful of requests for that in the .NET SDK and I think it's quite powerful.

nielsenko (approved):
It does expose `RealmObject.get<T>` which could use some polish (I don't like it returns Object?), but perhaps that can be addressed later.

nirinchev:
I believe RealmObject.get has always been exposed - that's what is used in the generated classes. We use it in tests, but I haven't changed the visibility of it (at least not intentionally).

desistefanova (approved):

lib/src/native/realm_core.dart
```diff
@@ -450,6 +453,53 @@ class _RealmCore {
...
+        (values, index) => values.elementAt(index).ref);
+
+    final properties = nativeProperties.map((e) => e.toSchemaProperty()).toList();
+
+    return SchemaObject(RealmObject, name, properties);
+  }
+
+  List<TReturn> _getValues<T extends NativeType, TReturn>(Arena arena, Pointer<T> Function(int size) createPtr,
+      bool Function(Pointer<T> valuesPtr, int size, Pointer<IntPtr> outSize) getValue, TReturn Function(Pointer<T> values, int index) getAtIndex) {
+    // TODO: this is a hack to get the size of the array and should not be needed once https://github.com/realm/realm-core/issues/5430 is addressed.
```
desistefanova, line 488: This core issue seems to be handled already. https://github.com/realm/realm-core/issues/5430

desistefanova (approved):
Is it possible to have flexible sync subscription by dynamic query?

lib/src/configuration.dart
```diff
@@ -48,11 +48,13 @@ typedef InitialDataCallback = void Function(Realm realm);
...
-
-  /// The default realm filename to be used. 
+  /// The default realm filename to be used.
   static String get defaultRealmName => _path.basename(defaultRealmPath);
   static set defaultRealmName(String name) => defaultRealmPath = _path.join(_path.dirname(defaultRealmPath), _path.basename(name));
-  
+
+  /// A collection of [SchemaObject] that will be used to construct the
+  /// [RealmSchema] once the Realm is opened.
+  final Iterable<SchemaObject> schemaObjects;
```
nirinchev, line 57: Note for reviewers: this is replacing the `RealmSchema` field that previously existed on the configuration. The reason is that it's now possible to provide an empty collection and have the schema be read from the file. We're now exposing the `schema` on the `Realm` instance directly as that can only be truly known after we've opened the file. I wanted to make that distinction more explicit, which is why I opted to change the field on the configuration (otherwise we'd have `realm.schema != realm.config.schema` and that felt very awkward).

lib/src/realm_class.dart
```diff
@@ -474,3 +464,74 @@ class RealmLogLevel {
   /// Same as [Level.OFF];
   static const off = Level.OFF;
 }
+
+/// @nodoc
+class RealmMetadata {
+  final Map<Type, RealmObjectMetadata> _typeMap = <Type, RealmObjectMetadata>{};
+  final Map<String, RealmObjectMetadata> _stringMap = <String, RealmObjectMetadata>{};
+
+  RealmMetadata._(Iterable<RealmObjectMetadata> objects) {
```
blagoev, line 473:
```suggestion
  RealmMetadata._(Iterable<RealmObjectMetadata> objectsMetadata) {
```

lib/src/realm_class.dart
```diff
@@ -474,3 +464,74 @@ class RealmLogLevel {
...
+    for (final meta in objects) {
```
blagoev, line 474:
```suggestion
    for (final objMetadata in objectsMetadata) {
```

lib/src/realm_class.dart
```diff
@@ -474,3 +464,74 @@ class RealmLogLevel {
...
+  RealmObjectMetadata getByType(Type type) {
+    final metadata = _typeMap[type];
+    if (metadata == null) {
+      throw RealmError("Object type $type not configured in the current Realm's schema. Add type $type to your config before opening the Realm");
+    }
+
+    return metadata;
+  }
+
+  RealmObjectMetadata getByName(String type) {
```
blagoev, line 492: How is this going to behave if there are same type names in different packages or libraries?

lib/src/realm_class.dart
```diff
@@ -474,3 +464,74 @@ class RealmLogLevel {
...
+
+    return metadata;
+  }
+}
+
+/// Exposes a set of dynamic methods on the Realm object. These don't use strongly typed
+/// classes and instead lookup objects by string name.
+///
+/// {@category Realm}
+class DynamicRealm {
```
blagoev, line 511: Should this class implement Realm class. Otherwise how are we going to keep up the interfaces in sync

lib/src/realm_class.dart
```diff
@@ -85,12 +86,20 @@ export 'session.dart' show Session, SessionState, ConnectionState, ProgressDirec
 ///
 /// {@category Realm}
 class Realm {
-  final Map<Type, RealmMetadata> _metadata = <Type, RealmMetadata>{};
-  final RealmHandle _handle;
+  late final RealmMetadata _metadata;
+  late final RealmHandle _handle;
+
+  /// An object encompassing this `Realm` instance's dynamic API.
+  late final DynamicRealm dynamic = DynamicRealm._(this);
```
blagoev, line 93: Why are we exposing this on the public interface. why not just have this as an argument to the migration function args. I think we should not make this public since there is no use case of this outside migrations. This is fully duplicating the typed interface so in a sense we are trying to make the SDK a dynamic one where we shouldn't. This SDK is for a typed language and should only provide typed language interface. At least we should try and in special cases like migrations we could provide a dynamic interface.

src/realm_dart_sync.cpp
```diff
@@ -109,7 +109,7 @@ RLM_API void realm_dart_sync_wait_for_completion_callback(realm_userdata_t userd
 
     auto ud = reinterpret_cast<realm_dart_userdata_async_t>(userdata);
     ud->scheduler->invoke([ud, error = std::move(error_copy)]() {
-        (reinterpret_cast<realm_sync_download_completion_func_t>(ud->dart_callback))(ud->handle, error.get());
+        (reinterpret_cast<realm_sync_wait_for_completion_func_t>(ud->dart_callback))(ud->handle, error.get());
```
blagoev, line 112: Not sure I followed the changes correctly to here. Was this a bug or something else?

lib/src/realm_object.dart
```diff
@@ -203,7 +204,9 @@ extension RealmEntityInternal on RealmEntity {
 mixin RealmObject on RealmEntity {
   RealmObjectHandle? _handle;
   RealmAccessor _accessor = RealmValuesAccessor();
-  static final Map<Type, RealmObject Function()> _factories = <Type, RealmObject Function()>{};
+  static final Map<Type, RealmObject Function()> _factories = <Type, RealmObject Function()>{
+    RealmObject: () => DynamicRealmObject._(),
```
blagoev, line 208: Not sure why would we need this.

blagoev (changes requested):
In general I think we should not be exposing the dynamic API except in migrations. We should be building a typed SDK instead of dynamic one and hence the SDK interface should be typed.
I tried to thoroughly review the changes since these are fundamental to the SDK internals

lib/src/list.dart
```diff
@@ -30,21 +30,26 @@ import 'results.dart';
...
+  factory RealmList._(RealmListHandle handle, Realm realm, RealmObjectMetadata? metadata) => ManagedRealmList._(handle, realm, metadata);
   factory RealmList(Iterable<T> items) => UnmanagedRealmList(items);
 }
 
 class ManagedRealmList<T extends Object> extends collection.ListBase<T> with RealmEntity implements RealmList<T> {
   final RealmListHandle _handle;
 
-  ManagedRealmList._(this._handle, Realm realm) {
+  @override
+  late final RealmObjectMetadata? _metadata;
```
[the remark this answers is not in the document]
nirinchev: Why? The `_metadata` field is in an abstract class and the compiler will complain about missing concrete implementations - is there an alternative approach here that I'm not seeing?

lib/src/list.dart
```diff
@@ -30,21 +30,26 @@ import 'results.dart';
 ///
 /// {@category Realm}
 abstract class RealmList<T extends Object> with RealmEntity implements List<T> {
+  late final RealmObjectMetadata? _metadata;
```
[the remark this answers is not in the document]
nirinchev: The Realm instance holds the metadata for all objects. The list needs to know what is the metadata of the objects it contains. This is not always possible to infer from the generic argument as the list can now be constructed as `RealmList<RealmObject>`.

lib/src/native/realm_core.dart
```diff
@@ -450,6 +453,63 @@ class _RealmCore {
     return RealmHandle._(realmPtr);
   }
 
+  RealmSchema readSchema(Realm realm) {
+    return using((Arena arena) {
+      final classKeys = _getValues<Uint32, int>(
```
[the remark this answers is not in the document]
nirinchev: This requires a lot of code duplication as we'll need to inline the _getValues function everywhere it's used. I don't see how this would be helpful or make the codebase more maintainable.

lib/src/native/realm_core.dart
```diff
@@ -450,6 +453,63 @@ class _RealmCore {
...
+    Pointer<T> valuesPtr;
+
+    while (true) {
+      valuesPtr = allocator(defaultSize);
+      _realmLib.invokeGetBool(() => getValue(valuesPtr, defaultSize, valuesCount));
+
+      if (defaultSize < valuesCount.value) {
+        // The supplied array was too small - resize it
+        defaultSize = valuesCount.value;
+        arena.free(valuesPtr);
```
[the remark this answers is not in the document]
nirinchev: An endless loop vs an endless recursion sounds about the same to me and a loop is generally cheaper. I don't see how the recursion would make the code easier to reason about.

lib/src/realm_class.dart
nirinchev: No, it provides a completely different API than the Realm class.

lib/src/realm_class.dart
nirinchev: I don't agree that there's a reason to limit the dynamic API to migrations. There are many use cases for the dynamic API outside of migrations:
1. Dart already exposes dynamic API despite being a statically typed language. So clearly there's some demand for it.
2. To write a Studio-like tool where you read the schema from disk.
3. To import dynamic data from an API (see [this forum post](https://www.mongodb.com/community/forums/t/how-to-add-dynamicrealmobject/119228) for such a request for the .NET SDK).
4. To handle unknown objects via sync. When we add support for Mixed/RealmValue, we can end up in a situation where an old version of the app is accessing objects that it doesn't have dart models for. Those need to be exposed as dynamic objects, otherwise the developer will have no way to interact with them.

src/realm_dart_sync.cpp
nirinchev: This is a change in the name from bumping the Core version.

lib/src/realm_object.dart
nirinchev: Because we need to make sure that `List<RealmObject>` will contain `DynamicRealmObject` elements.

lib/src/realm_class.dart
nirinchev: We don't support tables with the same name, so that would have been an error when opening the Realm file.

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
blagoev: I have strong opinion against writing such code. We can discuss how to improve it but this kind of code is totally degrading the codebase. I would suggest we avoid late generalizations like these.

lib/src/list.dart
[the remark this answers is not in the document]
blagoev: There should be a linter issue when you overwrite fields that suggest to not do this.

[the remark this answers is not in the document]
blagoev: https://dart-lang.github.io/linter/lints/overridden_fields.html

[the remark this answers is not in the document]
blagoev: oh, for abstract base class it is ok.

lib/src/list.dart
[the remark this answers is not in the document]
blagoev: It seems we need to discuss this offline.

lib/src/realm_class.dart
blagoev: Our task at hand is migrations. We don't have any need/request to support the scenarios you listed. Mixed and RealmValue might end up implemented in a way that does not need what .NET has. I think we are doing too much implementation of anticipation of future features implemented in certain ways. 
Dart has `dynamic` keyword and dynamic objects but never in this PR we have this keyword mentioned. So in a sense we are not following the language idiomaticity. Maybe it is just  me but I don't really understand the scenarios we are trying to cover maybe because they don't exists yet. Perhaps we can discuss offline to better understand what we are trying to achieve here.

test/configuration_test.dart
```diff
@@ -31,10 +31,6 @@ Future<void> main([List<String>? args]) async {
     Configuration.local([Car.schema]);
   });
 
-  test('Configuration exception if no schema', () {
```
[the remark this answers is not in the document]
nirinchev: A user might want to build a studio-like application with the dart SDK. I don't see a valid reason to prevent that.

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
nirinchev: This is not true - it's possible that between the invocations of `getValue` and `doWork` the value changed, which will lead to multiple invocations of the recursive function. By definition any recursive function that calls itself has the potential to lead to endless recursion if the conditions are not met. The above code is equivalent to

```dart
while (true) {
  value = getValue(size);
  if (size == value) {
    break;
  }

  size = value;
}
```

If the loop turns out to be endless, so will the recursion and vice-versa. A developer messing up the recursion base case is just as likely as a developer messing up the loop exit condition.

lib/src/realm_class.dart
blagoev: I guess my question should be why not merge these methods in Realm class or have these as extension methods. Why do we need a DynamicRealm concept?

lib/src/realm_object.dart
```diff
@@ -134,18 +133,20 @@ class RealmCoreAccessor implements RealmAccessor {
       final propertyMeta = metadata[name];
       if (propertyMeta.collectionType == RealmCollectionType.list) {
         final handle = realmCore.getListProperty(object, propertyMeta.key);
-        return object.realm.createList<T>(handle);
+        final listMeta = propertyMeta.objectType == null ? null : object.realm.metadata.getByName(propertyMeta.objectType!);
```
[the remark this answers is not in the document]
nirinchev: Yes - that's the case for a list of objects, where we expect the metadata to not be null. I can replace that with an assertion that instructs the user to report a bug if the metadata is null. Something like:

```dart
if (_metadata == null) {
  throw RealmError("A list of objects didn't have properly populated metadata. This is a bug in the Realm SDK - please report it at https://github.com/realm/realm-dart/issues/new");
}

return realm.createObject(T, value, _metadata!) as T;
```

blagoev:
I have been constantly thinking and revisiting this PR for the last weeks and I think we are doing a step in the wrong direction here. I have taken a look at how Realm Swift does this and am I really leaning towards a similar solution for Realm Flutter. I think Realm .NET made a mistake by enabling a generic dynamic API available at all times, which is only required while doing migrations. Instead I believe we should be providing this API only in the migration callback as Realm Swift does. We have this generic `dynamic` support in Dart which allows us to provide two dynamic objects, old and new version, in the migration callback from which we can allow even nicer than Swift API, cause `dynamic` objects in Dart allow access to any property even non existing ones. Let's discuss later when we regroup.

nirinchev:
String-based property setting/getting seems to be available via [subscripts](https://github.com/realm/realm-swift/blob/792c730c811c7c3e2da23e0d044c1584b0746b02/RealmSwift/Object.swift#L184-L192), like `foo["bar"] = 123`. As far as I can tell this functionality is always available, not just in migrations.

blagoev:
We have discussed internally and it seems we need a dynamic API in the Dart SDK to support certain use cases like app users building the Realm schema at runtime etc.

My comments about the `_getValues` function are still valid concerns. I suggest we refactor this code not to use such late function optimizations. This function is becoming an important internal SDK function while at the same time providing weird and unreadable signature. 
I guess I could help refactoring that code.

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
nirinchev: I have inlined the calls to _getValues. I still believe it's an unnecessary amount of code duplication, but don't care enough to hold this PR captive to resolve it.

nirinchev:
I have inlined `_getValues` - it's a fair amount of code duplication, but I don't suppose that code will change that often, so it's probably fine. @blagoev please take another look and let me know if something is unclear. Happy to also do a live review session if that'll help get this PR merged faster.

lib/src/realm_class.dart
```diff
@@ -107,12 +116,8 @@ class Realm {
   }
 
   void _populateMetadata() {
-    for (var realmClass in config.schema) {
-      final classMeta = realmCore.getClassMetadata(this, realmClass.name, realmClass.type);
-      final propertyMeta = realmCore.getPropertyMetadata(this, classMeta.key);
-      final metadata = RealmMetadata(classMeta, propertyMeta);
-      _metadata[realmClass.type] = metadata;
-    }
+    schema = config.schemaObjects.isNotEmpty ? RealmSchema(config.schemaObjects) : realmCore.readSchema(this);
```
desistefanova, line 119: What will happen if we create a Realm without schema we open the realm for the first time. The realm file will be created empty? I saw that this throw is removed `throw RealmError("No schema specified");`.

lib/src/configuration.dart
```diff
@@ -74,10 +78,10 @@ abstract class Configuration {
   static late String defaultRealmPath = _path.join(defaultStoragePath, 'default.realm');
 
   Configuration._(
-    List<SchemaObject> schemaObjects, {
+    this.schemaObjects, {
```
desistefanova, line 81: Is this still always required? Aren't we loading the schema from the realm file if the schema is not defined?

lib/src/native/realm_core.dart
```diff
@@ -1452,24 +1519,31 @@ class _RealmCore {
 
   List<UserIdentity> userGetIdentities(User user) {
     return using((arena) {
-      // TODO: Fix countIds in userGetIdentities once Core changes how count is retrieved. https://github.com/realm/realm-dart/issues/690
-      // This approach is prone to race conditions.
-      final idsCount = arena<Size>();
-      _realmLib.invokeGetBool(
-          () => _realmLib.realm_user_get_all_identities(user.handle._pointer, nullptr, 0, idsCount), "Error while getting user identities count");
+      final actualCount = arena<Size>();
```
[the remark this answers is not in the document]
nirinchev: Yes, I forgot we have an issue for this.

lib/src/realm_class.dart
nirinchev: Nothing - it will be an empty file.

lib/src/configuration.dart
nirinchev: I opted to keep it required since loading the schema from file is the more uncommon use case.

desistefanova (approved):
Added some questions?

lib/src/realm_class.dart
desistefanova: OK

lib/src/realm_class.dart
```diff
@@ -474,3 +464,74 @@ class RealmLogLevel {
   /// Same as [Level.OFF];
   static const off = Level.OFF;
 }
+
+/// @nodoc
+class RealmMetadata {
```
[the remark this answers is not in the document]
nirinchev: It's internal and not intended to be used by users (at least in this shape). We can probably expose it sometime in the future, but it's not the goal of this PR.

lib/src/native/realm_core.dart
```diff
@@ -448,6 +451,65 @@ class _RealmCore {
     return RealmHandle._(realmPtr);
   }
 
+  RealmSchema readSchema(Realm realm) {
+    return using((Arena arena) {
+      final actualCount = arena<Size>();
+      Pointer<Uint32> classesPtr;
+      var classesSize = 10;
+
+      while (true) {
```
blagoev, line 460: I feel bad we are introducing these endless loops in the code. Why not call this function again. I think I had a comment about this somewhere. No every other function is using an endless loop. This is error prone and could lead to problems if the code is refactored. Also its hard to read and reason about. Its much more clear if it calls itself with the correct size again.

lib/src/realm_object.dart
```diff
@@ -366,3 +367,8 @@ class RealmObjectNotificationsController<T extends RealmObject> extends Notifica
     streamController.addError(error);
   }
 }
+
+/// @nodoc
+class DynamicRealmObject with RealmEntity, RealmObject {
```
blagoev, line 372: I think we should overwrite `noSuchMethod` here to allow reading/writing any property on this object. Much better than using `RealmObject.get<XYZ>();`

pubspec.yaml
```diff
@@ -14,6 +14,7 @@ environment:
 dependencies:
   args: ^2.3.0
   build_cli_annotations: ^2.0.0
+  collection: ^1.16.0
```
blagoev, line 17: are we depending on that somewhere in this PR still? Or this was before some refactorings and is not needed?

test/dynamic_realm_test.dart
```diff
@@ -0,0 +1,197 @@
...
+      staticRealm.write(() {
+        staticRealm.add(Car('Honda'));
+        staticRealm.add(Car('Hyundai'));
+      });
+
+      final dynamicRealm = getDynamicRealm(staticRealm);
+
+      final car = dynamicRealm.dynamic.find(Car.schema.name, 'Honda');
+      expect(car, isNotNull);
+      expect(RealmObject.get<String>(car!, 'make'), 'Honda');
```
blagoev, line 181: this will be 
```
car.make
```
if we overwrite `noSuchMethod` in `DynamicRealmObject`

blagoev (changes requested):
I resolved most previous comments, still I have some here.

lib/src/realm_object.dart
blagoev: I see there is a previous discussion about this. I think @nielsenko is right and we could support dynamic with `noSuchMethod`. At the same time this object will have the ability to work with `.get<int>()` cause its part of the parent API

pubspec.yaml
blagoev: Feel free resolve if it is used.

lib/src/realm_class.dart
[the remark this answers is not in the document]
blagoev: I don't have a better suggestion so I am resolving this

lib/src/realm_object.dart
nirinchev: This is covered in a different PR - https://github.com/realm/realm-dart/pull/524. The current PR only exposes dynamic API on the Realm class, not the RealmObject. The introduction of `DynamicRealmObject` is necessary for some primitive testing, but I decided to split the two tasks as each is fairly large and hoped it will make reviews easier.

test/dynamic_realm_test.dart
nirinchev: As mentioned in the other comment, this is done in another PR: https://github.com/realm/realm-dart/pull/524/files#diff-c2ba1d881cf423a37576e2f5fd6b75b73bac7783f002a36eab9cb57815d4290dR289.

lib/src/native/realm_core.dart
nirinchev: I know that it's rare that a schema changes between two invocations of some code, but I wanted to strictly adhere to the Core API and make sure we don't introduce some very rare crashes where a user randomly gets an empty array because we provided a container that's too small.

lib/src/realm_object.dart
blagoev: ok I see

test/dynamic_realm_test.dart
blagoev: resolving

lib/src/native/realm_core.dart
blagoev: that's more readable code right there. It reads better, it is not hardcoded endless recursion and can not unintentionally become one if one refactors the code later on. The `while(true)` is just asking for trouble and is really hard to read imo. I hope we don't start using goto labels :)

Also there is no need to `arena.free(classesPtr);` in the `if` explicitly since this will get called automatically when the arena is torn down. Yes we will keep that allocation a bit longer but I think this is the idea of the arena to not bother freeing the memory. But freeing will also work.

blagoev: That code is perfect ^

blagoev: btw you can drop the `Internal`. `_readSchema` can co exists with `readSchema`

nirinchev:
@blagoev refactored the while loops to use recursion. While I don't see much of a difference, it's not a hill I'm willing to die on, so can you take another look and see if there's anything left or we can merge.

blagoev (approved):
Thanks @nirinchev. Lets merge it.

desistefanova:
> @blagoev refactored the while loops to use recursion. While I don't see much of a difference, it's not a hill I'm willing to die on, so can you take another look and see if there's anything left or we can merge.

It is not perfect in both ways. But it is accectable.

PR #498, opened 2022-04-26
Create a workflow that prepares a release

nirinchev:
Creates a workflow we can use to trigger a release build. Example PR being opened by the workflow: https://github.com/realm/realm-dart/pull/504.

It also replaces all the changelogs with symlinks to the main package changelog. The reason is that it simplifies workflows if we only have to update a single changelog and is offering too little value to end users since they're unlikely to ever interact with the common packages.

Fixes https://github.com/realm/realm-dart/issues/532
Related to https://github.com/realm/realm-dart/issues/487


.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,44 @@
...
+        id: update-changelog
+        uses: realm/ci-actions/update-changelog@dac063f8ebd86c201590d07536d6f56f86f66ba9
+        with:
+          changelog: ${{ github.workspace }}/CHANGELOG.md
+
+      - name: Update pubspec.yaml
+        uses: jacobtomlinson/gha-find-replace@b76729678e8d52dadb12e0e16454a93e301a919d #! 2.0.0
+        with:
+          find: 'version: .*'
+          replace: 'version: ${{ steps.update-changelog.outputs.new-version }}+alpha'
```
nielsenko, line 25: So `+alpha` is hardcoded in the action.. I guess that is okay

.github/workflows/prepare-release.yml
```diff
@@ -0,0 +1,44 @@
...
+          include: '**/pubspec.yaml'
+
+      - name: Update realm.podspec
+        uses: jacobtomlinson/gha-find-replace@b76729678e8d52dadb12e0e16454a93e301a919d #! 2.0.0
+        with:
+          find: "  s.version                   = '[^']*'"
```
nielsenko, line 31: Should we be a little less brittle wrt whitespace?

.github/workflows/prepare-release.yml
nirinchev: Yeah, that's only temporary until we move to 1.x at which point we can drop the +alpha.

.github/workflows/prepare-release.yml
nirinchev: The podspec file is arranged so that the `=` signs are aligned one under another so I don't suppose this will change often. I can change that to a more convoluted regex, but it seems like a bit of an overkill.

.github/workflows/prepare-release.yml
```diff
@@ -13,3 +13,34 @@ jobs:
...
+
+      - name: Update pubspec.yaml
+        uses: jacobtomlinson/gha-find-replace@b76729678e8d52dadb12e0e16454a93e301a919d #! 2.0.0
+        with:
+          find: 'version: .*'
+          replace: 'version: ${{ steps.update-changelog.outputs.new-version }}+alpha'
+          include: '**/pubspec.yaml'
+
+      - name: Update realm.podspec
+        uses: jacobtomlinson/gha-find-replace@b76729678e8d52dadb12e0e16454a93e301a919d #! 2.0.0
```
desistefanova, line 31: Why do we need this specific commit b767296 of jacobtomlinson/gha-find-replace?

.github/workflows/prepare-release.yml
```diff
@@ -13,3 +13,34 @@ jobs:
...
+        with:
+          find: "  s.version(\\s+)= '[^']*'"
+          replace: "  s.version$1= '${{ steps.update-changelog.outputs.new-version }}+alpha'"
+          include: '**/realm.podspec'
+
+      - name: Create Release PR
+        uses: peter-evans/create-pull-request@7380612b49221684fefa025244f2ef4008ae50ad #! 3.10.1
+        with:
+          branch: release/${{ steps.update-changelog.outputs.new-version }}
```
desistefanova, line 40: Where is this variable coming from `steps.update-changelog.outputs.new-version`?

nielsenko (approved):

desistefanova (approved):
I have only two questions while trying to understand how it works.

.github/workflows/prepare-release.yml
nirinchev: This is a best practice when using 3rd party actions. If we were to use `jacobtomlinson/gha-find-replace@v2`, an attacker that compromises the `jacobtomlinson/gha-find-replace` repo can push the v2 tag pointing to a malicious commit. It's part of the security hardening recommendations by Github: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions.

.github/workflows/prepare-release.yml
nirinchev: It's this step: https://github.com/realm/realm-dart/blob/984017ca461a692541abd068ad95af82ccdba368/.github/workflows/prepare-release.yml#L17-L21

The action is part of our common CI actions: https://github.com/realm/ci-actions/tree/main/update-changelog. Once it runs, it will set the `new-version` variable to the inferred new version from the changelog.

blagoev (approved):
I have the feeling we are doing it wrong by using single changelog in all packages. Hope this does not come back to bite us at some point

.github/workflows/prepare-release.yml
```diff
@@ -13,3 +13,42 @@ jobs:
...
+        uses: jacobtomlinson/gha-find-replace@b76729678e8d52dadb12e0e16454a93e301a919d #! 2.0.0
+        with:
+          find: "  s.version(\\s+)= '[^']*'"
+          replace: "  s.version$1= '${{ steps.update-changelog.outputs.new-version }}+alpha'"
+          include: '**/realm.podspec'
+
+      - name: Update realm_core.libraryVersion
+        uses: jacobtomlinson/gha-find-replace@b76729678e8d52dadb12e0e16454a93e301a919d #! 2.0.0
+        with:
+          find: "  String get libraryVersion => '[^']*';"
```
blagoev, line 40: Should we fail CI if we can not find this line. Just to be sure we don't refactor this code

nirinchev: Added a check for that

PR #505, opened 2022-04-26
tests: email/password register user

### LINKED ISSUE #453: Implement tests: Auth Provider: register email

desistefanova:
Testing register user with email/password authentication provider. Fixes #453


test/credentials_test.dart
```diff
@@ -36,4 +36,61 @@ Future<void> main([List<String>? args]) async {
...
+  baasTest('Email/Password - register user', (configuration) async {
+    final app = App(configuration);
+    final authProvider = EmailPasswordAuthProvider(app);
+    String emailPrefix = generateRandomString(5);
+    String username = "realm_tests_do_autoverify$emailPrefix@bar.com";
+    String password = "SWV23R#@T#VFQDV";
+    await authProvider.registerUser(username, password);
+    final user = await app.logIn(Credentials.emailPassword(username, password));
+    await app.logout(user);
+    await app.removeUser(user);
```
desistefanova, line 60: By some reason the user is not removed from the app. Should be investigated.

nielsenko (approved):

test/credentials_test.dart
nirinchev: We probably don't these here and we'll write separate tests for user.logout/app.removeUser.

nirinchev: We probably don't these here and we'll write separate tests for user.logout/app.removeUser.

nirinchev (approved):

test/credentials_test.dart
desistefanova: Yes sure.

nirinchev (changes requested):
These tests appear not to be running on CI. Can you hold off on merging them until we figure out what's going on? They are skipped but no reason is provided:

Screen Shot 2022-04-29 at 01 26 19

desistefanova:
> These tests appear not to be running on CI. Can you hold off on merging them until we figure out what's going on? They are skipped but no reason is provided:
> 
> Screen Shot 2022-04-29 at 01 26 19

I think the problem is in this row in `baasTest`
```
if (skip == null) {
    skip = url == null ? "BAAS URL not present" : true;
```
I will try to fix it.

nirinchev:
It looks like there's a bug in the skip logic: https://github.com/realm/realm-dart/blob/f3920a70bf48d82536225f2a1e352a7191d9e563/test/test.dart#L221-L223

The way I read it, if the url is null, the test will be skipped with reason `BAAS URL not present`, otherwise it will just be skipped because `skip` is set to `true`.

desistefanova:
> Edit: seems like you were faster than me 😅
> 
> It looks like there's a bug in the skip logic:
> 
> https://github.com/realm/realm-dart/blob/f3920a70bf48d82536225f2a1e352a7191d9e563/test/test.dart#L221-L223
> 
> The way I read it, if the url is null, the test will be skipped with reason `BAAS URL not present`, otherwise it will just be skipped because `skip` is set to `true`.

Yes, I was looking at this just right now.

.github/workflows/ci.yml
```diff
@@ -277,6 +277,41 @@ jobs:
         working-directory: ./flutter/realm_flutter/tests
 
 # Windows jobs
+  baas-windows:
+    runs-on: ubuntu-latest
```
blagoev, line 281: this is seems strange. is this intentional

test/test.dart
```diff
@@ -88,9 +88,9 @@ class $RemappedClass {
 }
 
 String? testName;
-Map<String, BaasApp> baasApps = <String, BaasApp>{};
-
+final baasApps = <String, BaasApp>{};
 final _openRealms = Queue<Realm>();
+late Uri? url;
```
blagoev, line 93: why a global var

blagoev (changes requested):

.github/workflows/ci.yml
nirinchev: What is strange - the fact we're adding baas tests for Windows or that the deploy step runs on ubuntu?

test/baas_test.dart
```diff
@@ -0,0 +1,7 @@
+import 'credentials_test.dart' as credentials_tests;
+import 'app_test.dart' as app_tests;
+
+Future<void> main([List<String>? args]) async {
```
nirinchev, line 4: Do we need this file?

test/test.dart
desistefanova: This will be removed and I will take the code from master.

test/baas_test.dart
desistefanova: No more

desistefanova: We don't need it anymore.

.github/workflows/ci.yml
blagoev: I misread this. Closing

blagoev (changes requested):

PR #506, opened 2022-04-27
tests: email/password confirm user

desistefanova:
Create tests for email/password confirm user tokens and auto confirm user.


blagoev (approved):

test/credentials_test.dart
```diff
@@ -61,36 +61,99 @@ Future<void> main([List<String>? args]) async {
...
   });
 
+  baasTest('Email/Password - register user auto confirm', (configuration) async {
+    final app = App(configuration);
+    final authProvider = EmailPasswordAuthProvider(app);
+    String username = "${generateRandomString(5)}@bar.com";
+    String password = "SWV23R#@T#VFQDV";
+    // For application with name 'auto' and with confirmationType = 'autoConfirm'
+    // all the usernames are automatically confirmed.
+    await authProvider.registerUser(username, password);
```
nirinchev, line 71: Can we add a test that we can now login with the credentials? FYI, there's [a bug in Cloud](https://jira.mongodb.org/browse/REALMC-12253) that means that you may have to retry the login since there's a small delay between registering and the autoconfirmation kicking in.

test/credentials_test.dart
```diff
@@ -61,36 +61,99 @@ Future<void> main([List<String>? args]) async {
...
+  baasTest('Email/Password - confirm user token expired', (configuration) async {
+    final app = App(configuration);
+    final authProvider = EmailPasswordAuthProvider(app);
+    String username = "${generateRandomString(5)}@hotmail.com";
+    String password = "SWV23R#@T#VFQDV";
+    await authProvider.registerUser(username, password);
+    expect(() async {
+      await authProvider.confirmUser(
+          "0e6340a446e68fe02a1af1b53c34d5f630b601ebf807d73d10a7fed5c2e996d87d04a683030377ac6058824d8555b24c1417de79019b40f1299aada7ef37fddc",
+          "6268f7dd73fafea76b730fc9");
```
nirinchev, line 114: How are these tokens generated?

test/test.dart
```diff
@@ -115,7 +115,7 @@ Future<void> setupTests(List<String>? args) async {
   parseTestNameFromArguments(args);
 
   await setupBaas();
-
+
```
nirinchev, line 118:
```suggestion

```

nirinchev (commented):
Looks good - I have a question and a suggestion

test/credentials_test.dart
desistefanova: They were sent to my email after registration. I just copied them. They are old tokens, but in a valid format. In this case the validation is throwing different exception.

test/credentials_test.dart
desistefanova: Done with retryLogin function

test/credentials_test.dart
nirinchev: Hm... I may be wrong, but I don't think we'll be able to use them like that. These are just strings and not JWTs that encode their expiration - I believe they're just stored in the database, so when you try to run this test against a different app, it'll fail with invalid token rather than expired token. Once we fix test runs on CI, I expect this test to fail, but let's see.

desistefanova: Yes, right, but they are running on the same app in this manual test. It would be good if I had the server code to check the validation, but I couldn't access it.

nielsenko (approved):

test/credentials_test.dart
```diff
@@ -52,55 +52,120 @@ Future<void> main([List<String>? args]) async {
...
+    String username = "${generateRandomString(5)}@hotmail.com";
+    String password = "SWV23R#@T#VFQDV";
+    await authProvider.registerUser(username, password);
+    expect(() async {
+      await authProvider.confirmUser("abc", "123");
+    }, throws<RealmException>("invalid token data"));
+  }, appName: "emailConfirm");
+
+  group("Email/Password - confirm user - manual tests", () {
+    // The tests in this group are for manual testing, since they require interaction with mail box.
```
blagoev, line 142: could we move the comment to be on the group call. Also could we rename all manual tests to include `Manual` in the name

blagoev (changes requested):
some small comments

test/credentials_test.dart
desistefanova: This test is not manual. We are passing expired tokens and the exception is thrown.

test/credentials_test.dart
desistefanova: OK

test/credentials_test.dart
blagoev: ok thanks

blagoev (approved):

PR #511, opened 2022-04-27
Cache complete binary builds

nirinchev:
This adds an extra cache step to the native builds that allows us to completely skip the build if we found the binary in the cache.

It speeds up builds noticeably when nothing in `./src` changes, otherwise it falls back to normal builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,7 +103,7 @@ jobs:
         uses: actions/download-artifact@v2
         with:
           name: librealm-linux
-          path: binary
+          path: binary/linux
 
       - name : Setup Dart SDK
         uses: dart-lang/setup-dart@main
@@ -161,7 +161,7 @@ jobs:
         uses: actions/download-artifact@v2
         with:
           name: librealm-linux
-          path: binary
+          path: binary/linux
 
       - name: Setup Flutter
         uses: subosito/flutter-action@v2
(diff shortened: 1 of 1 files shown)


fealebenpae (approved):
Do we still want to use ccache if we’re also caching the whole binary?

nirinchev:
I think you're being overly cautious here. We should not be wasting time waiting for the same build over and over again. There's no reason to believe this will silently fail unless we don't write any tests that exercise the new functionality. If we include the wrong binary, tests that attempt to call new functions will fail to find the correct symbol. Happy to chat about it during the roundtable, but we I feel strongly about reducing CI times and the possibilty that we accidentally change something about the way we build and don't realize we're no longer building correctly on CI is not realistic.

nielsenko (approved):
Personally I like the full caching of the binary folder. We had something similar previously, but it used the submodule commit, which didn't pick up changes to the dart specific c/c++ files. If I guess the semantics of `hashFiles` correctly, this PR should handle that issue correctly.

desistefanova (approved):

desistefanova (approved):

.github/workflows/ci.yml
```diff
@@ -58,20 +56,30 @@ jobs:
         uses: actions/checkout@v2
         with:
           submodules: 'recursive'
-      
+
+      - name: Check cache
+        id: check-cache
+        if: '!contains(github.head_ref, 'release/')'
+        uses: nirinchev/cache@d7c96a77c26ab70dd32b202c885cb4b34d95d8a8
```
blagoev, line 63: I would suggest we move this fork to the Realm org

blagoev (approved):
With a broken heart :broken_heart:.  I really think we should not be doing the compiler work here. Hope this does not cause any issues in the future.

PR #517, opened 2022-04-28
Add a toDartString extension that optionally frees native memory

nirinchev:


fealebenpae (approved):

nielsenko (approved):
This will work, but it puts the burden to remember on the caller. Down the line I would like to have an interface where we avoid first constructing a `std::string` the copying it to a `char*` only to convert it from `utf8` to `utf16`.

nirinchev:
I think for a more robust fix, we'll need to modify the C API, which I don't think is feasible right now.

lib/src/native/realm_core.dart
```diff
@@ -70,7 +70,7 @@ class _RealmCore {
     return _instance ??= _RealmCore._();
   }
 
-  String get libraryVersion => _realmLib.realm_get_library_version().cast<Utf8>().toDartString();
+  String get libraryVersion => _realmLib.realm_get_library_version().toDartString()!;
```
blagoev, line 73: I suggest we keep the `cast<Utf8>`. 
This change enables toDartString on every `Pointer<Int8>` which is not a string every time. Being more explicit here is intentional.

blagoev (changes requested):
It will work, but I would suggest keeping the extension on `Pointer<utf8>`. Lets keep the intentional interface.

lib/src/native/realm_core.dart
nirinchev: In any case, changed it to be an extension on Pointer<Utf8> - any time either approach would have saved is probably not worth debating this further.

desistefanova (approved):

lib/src/native/realm_core.dart
```diff
@@ -70,7 +70,7 @@ class _RealmCore {
     return _instance ??= _RealmCore._();
   }
 
-  String get libraryVersion => _realmLib.realm_get_library_version().cast<Utf8>().toDartString();
+  String get libraryVersion => _realmLib.realm_get_library_version().cast<Utf8>().toRealmDartString()!;
```
blagoev, line 73: why not use `toDartString` here

blagoev: and on all places where we use `!` in `toRealmDartString()!`

lib/src/native/realm_core.dart
```diff
@@ -81,7 +81,7 @@ class _RealmCore {
 
     String? message;
     if (error.ref.message != nullptr) {
-      message = error.ref.message.cast<Utf8>().toDartString();
+      message = error.ref.message.cast<Utf8>().toRealmDartString();
```
blagoev, line 84: we are not reaping the benefit of introducing `toRealmDartString` here. I think the previous version was removing the null check

lib/src/native/realm_core.dart
```diff
@@ -81,7 +81,7 @@ class _RealmCore {
 
     String? message;
     if (error.ref.message != nullptr) {
```
blagoev, line 83:
```suggestion
    
```

blagoev (changes requested):
some changes suggested

lib/src/native/realm_core.dart
nirinchev: My thinking was that it's better to have a single method for accessing strings. That way it's less cognitive load and we'll have an easier time refactoring stuff if necessary. Getting used to using `toRealmDartString` will also remind people to check whether they need to free the native string.

lib/src/native/realm_core.dart
nirinchev: yeah, I overlooked that

blagoev (approved):

PR #519, opened 2022-04-28
Add support for datetime properties

nirinchev:
This depends on https://github.com/realm/realm-dart/pull/495 because I was halfway through testing dynamic access when I realized dates are not supported.

Fixes https://github.com/realm/realm-dart/issues/520

UPDATE: Superseded by #569. Kept as a draft cause it has nullable datetime support as well

### DIFF
--- a/lib/src/native/realm_core.dart
+++ b/lib/src/native/realm_core.dart
@@ -70,7 +70,7 @@ class _RealmCore {
     return _instance ??= _RealmCore._();
   }
 
-  String get libraryVersion => _realmLib.realm_get_library_version().cast<Utf8>().toDartString();
+  String get libraryVersion => _realmLib.realm_get_library_version().toDartString()!;
 
   LastError? getLastError(Allocator allocator) {
     final error = allocator<realm_error_t>();
@@ -79,12 +79,7 @@ class _RealmCore {
       return null;
     }
 
-    String? message;
-    if (error.ref.message != nullptr) {
-      message = error.ref.message.cast<Utf8>().toDartString();
-    }
-
-    return LastError(error.ref.error, message);
+    return LastError(error.ref.error, error.ref.message.toDartString());
   }
 
   void throwLastError([String? errorMessage]) {
@@ -242,7 +237,7 @@ class _RealmCore {
     final classInfo = arena<realm_class_info>();
     _realmLib.invokeGetBool(() => _realmLib.realm_get_class(realm.handle._pointer, classKey, classInfo));
 
-    final name = classInfo.ref.name.cast<Utf8>().toDartString();
+    final name = classInfo.ref.name.toDartString()!;
 
     final nativeProperties = _getValues<realm_property_info, realm_property_info>(
         arena,
@@ -280,7 +275,7 @@ class _RealmCore {
   }
 
   String getFilesPath() {
-    return _realmLib.realm_dart_get_files_path().cast<Utf8>().toDartString();
+    return _realmLib.realm_dart_get_files_path().toDartString()!;
   }
 
   void closeRealm(Realm realm) {
@@ -322,13 +317,7 @@ class _RealmCore {
         throwLastError("Class $className not found in ${realm.config.path}");
       }
 
-      String? primaryKey;
-      if (classInfo.ref.primary_key != nullptr) {
-        primaryKey = classInfo.ref.primary_key.cast<Utf8>().toDartString();
-        if (primaryKey.isEmpty) {
-          primaryKey = null;
-        }
-      }
+      final primaryKey = classInfo.ref.primary_key.toDartString(treatEmptyAsNull: true);
       return RealmObjectMetadata(className, classType, primaryKey, classInfo.ref.key, _getPropertyMetadata(realm, classInfo.ref.key));
     });
   }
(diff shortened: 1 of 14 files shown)


lib/src/native/realm_core.dart
```diff
@@ -1338,6 +1338,12 @@ void _intoRealmValue(Object? value, Pointer<realm_value_t> realm_value, Allocato
         }
         realm_value.ref.type = realm_value_type.RLM_TYPE_UUID;
         break;
+      case DateTime:
+        final microseconds = (value as DateTime).toUtc().microsecondsSinceEpoch;
+        realm_value.ref.values.timestamp.seconds = microseconds ~/ 1000000;
+        realm_value.ref.values.timestamp.nanoseconds = (microseconds % 1000000) * 1000;
+        realm_value.ref.type = realm_value_type.RLM_TYPE_TIMESTAMP;
+        break;
```
nielsenko, line 1346: So no timezone :-/ .. Was about to suggest using https://pub.dev/packages/timezone, but since core doesn't support there is little point.

lib/src/native/realm_core.dart
```diff
@@ -1371,7 +1377,9 @@ extension on Pointer<realm_value_t> {
       case realm_value_type.RLM_TYPE_BINARY:
         throw Exception("Not implemented");
       case realm_value_type.RLM_TYPE_TIMESTAMP:
-        throw Exception("Not implemented");
+        final seconds = ref.values.timestamp.seconds;
+        final nanoseconds = ref.values.timestamp.nanoseconds;
+        return DateTime.fromMicrosecondsSinceEpoch(seconds * 1000000 + nanoseconds ~/ 1000).toUtc();
```
nielsenko, line 1382: Looks kosher, but I notice a lot of failing tests.

test/realm_object_test.dart
```diff
@@ -357,4 +358,133 @@ Future<void> main([List<String>? args]) async {
...
+
+    test('Nullable date roundtrips correctly: $date', () {
+      final config = Configuration([AllTypes.schema]);
+      final realm = getRealm(config);
+      final obj = realm.write(() {
+        return realm.add(AllTypes('', false, DateTime.utc(0), 0, ObjectId(), Uuid.v4(), 0, nullableDateProp: date));
+      });
+
+      final json = obj.toJson();
+      expect(json, contains('"dateProp":"${DateTime.utc(0).toRealmString()}"'));
```
nielsenko, line 391:
```suggestion
      expect(json['dateprop'], DateTime.utc(0).toRealmString()));
```

test/test.dart
```diff
@@ -265,3 +278,14 @@ Future<void> baasTest(
 extension RealmObjectTest on RealmObject {
   String toJson() => realmCore.objectToString(this);
```
nielsenko, line 279: This is surprising `toJson` by convention should return `dynamic` (runtime-type of objects will be `Map<String, dynamic>`). I would suggest running `jsonDecode` on the decoded `String` to follow convention.

test/test.dart
```diff
@@ -265,3 +278,14 @@ Future<void> baasTest(
...
 }
+
+extension DateFormat on DateTime {
+  String toRealmString() {
+    final utc = toUtc();
+    // This is kind of silly, but Core serializes negative dates as -003-01-01 12:34:56
+    final utcYear = utc.year < 0 ? '-${utc.year.abs().toString().padLeft(3, '0')}' : utc.year.toString().padLeft(4, '0');
```
nielsenko, line 286: Weird, and why pad 3? Just curious

nielsenko (approved):
Looks okay, but we should track down why the tests fails on some platforms

test/test.dart
nirinchev: I have no idea - I believe it was likely overlooked by the core team.

lib/src/native/realm_core.dart
nirinchev: Looks like a bug in date serialization in Core (or a bug in our code). Investigating.

lib/src/native/realm_core.dart
nirinchev: Yeah, that's something we've never supported and it comes up occasionally as an annoyance.

test/test.dart
nirinchev: I'm slightly worried about taking such approach because the json Core produces might not be the most valid json in the world - particularly, they'll serialize integers as strings, dates as non-roundtrippable strings, etc.

nirinchev:
I filed https://github.com/realm/realm-core/issues/5451 to track fixes for roundtripping zero/negative dates. I tried it on Windows and the values look correct in Studio, so I'm 99% confident it's a serialization bug in `to_json` and not an actual logic issue on our end.

nielsenko:
Should we disable the tests on windows for now? Or rework them not to use `toJson`. I think it would be nice to land this PR before release.

nirinchev:
This PR works around the issues with `to_json` on Windows, so the tests are not a blocker to merging it. It will need to be cherry picked on top of master though.

nirinchev:
Added in https://github.com/realm/realm-dart/pull/569.

blagoev:
reopening since this has nullable datetime support.

nirinchev:
We already merged nullable datetime support (there was nothing special about this). Did you mean to reopen https://github.com/realm/realm-dart/pull/462? Even so, we don't really need those PRs open - they'll still be accessible via the Github UI and we can't cherry pick them anyway since they've diverged too much from master. We'll probably need to update https://github.com/realm/realm-dart/pull/459 to pull in oid and uuid support.

PR #522, opened 2022-04-29
Fix a bug in bassTest, that url is never null, and skip is always true

nielsenko:


blagoev (approved):

test/test.dart
```diff
@@ -216,12 +216,13 @@ Future<void> baasTest(
   String appName = 'flexible',
   dynamic skip,
 }) async {
-  final url = Uri.tryParse(Platform.environment['BAAS_URL'] ?? 'https://realm-dev.mongodb.com');
+  final uriVariable = Platform.environment['BAAS_URL'];
+  final url = uriVariable != null ? Uri.tryParse(uriVariable) : null;
```
nirinchev, line 220: I think we should use `Uri.parse` here - if the developer explicitly set `BAAS_URL` to something, the tests should fail if it's not a valid uri rather than silently ignore it.

nirinchev (approved):

PR #524, opened 2022-04-29
Add string-based access API for RealmObject

nirinchev:
Adds a string-based API to read properties of a RealmObject. Right now we only support reading as that's required to get migrations working, but it can be extended in the future to also support setting.

The API is the following:

```dart
final obj = realm.dynamic.find("Foo", 123);

// Get simple primitive
final str = obj.dynamic.get<String>("stringProp");

// Get another object
final bar = obj.dynamic.get<RealmObject>("bar");
final barString = bar.dynamic.get<String>("stringProp");

// Get a list of objects
final bars = obj.dynamic.getList<RealmObject>("bars");

// Iterate the list as normal
for (final bar in bars) {
  print(bar.dynamic.get<DateTime>("dateProp");
}
```

Part of https://github.com/realm/realm-dart/issues/70.

### DIFF
--- a/lib/src/realm_object.dart
+++ b/lib/src/realm_object.dart
@@ -400,7 +400,11 @@ class DynamicRealmObject {
   }
 
   List<T> getList<T extends Object>(String name) {
-    _validatePropertyType<T>(name, RealmCollectionType.list, false);
+    final prop = _validatePropertyType<T>(name, RealmCollectionType.list, false);
+    if (T == Object && prop?.propertyType == RealmPropertyType.object) {
+      return RealmObject.get<RealmObject>(_obj, name) as List<T>;
+    }
+
     return RealmObject.get<T>(_obj, name) as List<T>;
   }
 
--- a/lib/src/realm_object.dart
+++ b/lib/src/realm_object.dart
@@ -205,7 +205,7 @@ mixin RealmObject on RealmEntity {
   RealmObjectHandle? _handle;
   RealmAccessor _accessor = RealmValuesAccessor();
   static final Map<Type, RealmObject Function()> _factories = <Type, RealmObject Function()>{
-    RealmObject: () => DynamicRealmObject._(),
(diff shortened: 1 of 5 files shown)


lib/src/realm_object.dart
```diff
@@ -267,6 +269,8 @@ mixin RealmObject on RealmEntity {
     final controller = RealmObjectNotificationsController<T>(object);
     return controller.createStream();
   }
+
+  late final DynamicRealmObject dynamic = DynamicRealmObject._(this);
```
fealebenpae, line 273: I'm not a super huge fan of hiding this API under a `dynamic` property like we did in .NET :/

nirinchev: The API is kind of large and niche, so I'd rather not pollute the main interface but happy to discuss alternatives.

lib/src/realm_object.dart
```diff
@@ -369,6 +373,91 @@ class RealmObjectNotificationsController<T extends RealmObject> extends Notifica
...
+    final prop = _validatePropertyType<T>(name, RealmCollectionType.none, true);
+
+    // If the user didn't provide a type argument, but the target is object, we should
+    // use RealmObject, otherwise we won't be able to find a Factory for the object.
+    if (T == Object && prop?.propertyType == RealmPropertyType.object) {
+      return RealmObject.get<RealmObject>(_obj, name) as T?;
+    }
+
+    return RealmObject.get<T>(_obj, name) as T?;
+  }
```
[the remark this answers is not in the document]
nirinchev: With dart links and strings are non-nullable. But I have reworked the API to get rid of the `getNullable` option as that was annoying me as well. Sadly, that required some changes I'm not overly happy with, but the dart type system is fairly limiting. LMK what you think of the new approach.

nirinchev:
Superseded by https://github.com/realm/realm-dart/pull/853.

PR #525, opened 2022-04-30
Expose customData and refreshCustomData on User

### LINKED ISSUE #410: App: refresh user custom data
realm_app_refresh_custom_data

### LINKED ISSUE #411: App: Link user to app
realm_app_link_user

### LINKED ISSUE #415: Users: Logout user
realm_user_log_out

### LINKED ISSUE #417: Users: Get identity
realm_user_get_identity

### LINKED ISSUE #418: Users: Get all identities
realm_user_get_all_identities

### LINKED ISSUE #420: Users: Get user state
realm_user_get_state

### LINKED ISSUE #423: Users: Get user custom data
realm_user_get_custom_data

blagoev:
Expose app property on User
Expose customData and refreshCustomData on User
switch to VS 2022 for Windows Builds (locally and on CI)

closes #423 
closes #410 
closes #411 
closes #420 
closes #417 
closes #418 
closes #415


Apple13ios:
> Expose app property on User Expose customData and refreshCustomData on User switch to VS 2022 for Windows Builds (locally and on CI)
> 
> closes #423 closes #410 closes #411 closes #420

lib/src/native/realm_core.dart
```diff
@@ -1104,6 +1104,82 @@ class _RealmCore {
...
+            ),
+        "Link credentials failed");
+    return completer.future;
+  }
+
+  int userGetState(User user) {
+    return _realmLib.realm_user_get_state(user.handle._pointer);
+  }
+
+  UserIdentity userGetIdentity(User user) {
```
nirinchev, line 1151: This should return just a string, no need to convert it to a `UserIdentity` object.

lib/src/native/realm_core.dart
```diff
@@ -1104,6 +1104,82 @@ class _RealmCore {
...
+
+  List<UserIdentity> userGetIdentities(User user) {
+    return using((arena) {
+      final idsCount = arena<IntPtr>();
+      _realmLib.invokeGetBool(
+          () => _realmLib.realm_user_get_all_identities(user.handle._pointer, nullptr, 0, idsCount), "Error while getting user identities count");
+
+      final idsPtr = arena<realm_user_identity_t>(idsCount.value);
+      _realmLib.invokeGetBool(
+          () => _realmLib.realm_user_get_all_identities(user.handle._pointer, idsPtr, idsCount.value, idsCount), "Error while getting user identities");
```
nirinchev, line 1167: This approach is prone to race conditions due to https://github.com/realm/realm-core/issues/5430. Nothing we can do about it until the Core bug has been fixed, but we can at least leave a comment to make sure we don't forget about it.

lib/src/native/realm_core.dart
```diff
@@ -1405,6 +1481,16 @@ extension on Object {
   }
 }
 
+extension on List<AuthProviderType> {
+  AuthProviderType fromIndex(int index) {
+    if (!AuthProviderType.values.any((value) => value.index == index)) {
```
nirinchev, line 1486: Why is this an extension if it never accesses the object it extends? Should we use `this` instead of `AuthProviderType.values`?

lib/src/user.dart
```diff
@@ -27,12 +32,96 @@ import 'native/realm_core.dart';
...
+  /// by linking it with e.g. Email/Password credentials.
+  /// Note: It is not possible to link two existing users of MongoDB Realm. The provided credentials must not have been used by another user.
+  Future<User> linkCredentials(Credentials credentials) async {
+    final userHandle = await realmCore.userLinkCredentials(app, this, credentials);
+    return UserInternal.create(app, userHandle);
+  }
+
+  /// The current state of this [User].
+  UserState get state {
+    final nativeState = realmCore.userGetState(this);
```
nirinchev, line 64: Should we change `realmCore.userGetState` to return `UserState` rather than an int?

nirinchev (commented):
Looks good - mostly minor comments from me.

desistefanova (approved):

lib/src/native/realm_core.dart
blagoev: I know .NET has this as a string. My thinking here it is more on par with the other API. So we have a getter returning a single `UserIdentity` and one that returns multiple ones as List<UserIdentity>. The properties are named 'identity' and `identities` which kind of implies more aligned return type.

lib/src/native/realm_core.dart
blagoev: indeed. We can't do much about it. I will add a TODO comment

lib/src/native/realm_core.dart
blagoev: yes that's an extraordinary one. I wanted to extend `List<AuthProviderType>` but still enumerate all enum values. Otherwise the exception will not make sense. But I still wanted this method to be available on `AuthProviderType.values`.  
I guess I can change it to use the `this` but than the exception will mean that `the index was not found in this list only`. 
The current implementation means that every `List<AuthProviderType>` has a method that can convert an `int` to an `AuthProviderType`. To me it's reasonable to have extension methods which do not access `this`.

If you feel strongly about it I guess I could rework this to use `this` instead.

lib/src/user.dart
blagoev: I moved the conversion to realm_core

lib/src/native/realm_core.dart
nirinchev: Those are different though. The User has an Id which is unique to that user and then they have one or more identities each with their unique ids. If I have a user that logged in with email/password auth, I could have something like:

```
{
  id: "123",
  identities: [
    { id: "abc", provider: "password" }
  ]
}
```

I think it would be weird if we return an identity here and `user.identity != user.identitites.single`.

lib/src/native/realm_core.dart
nirinchev: No, I don't feel strongly about it, just found it a little awkward as I was reading it.

lib/src/native/realm_core.dart
blagoev: got it will fix

PR #528, opened 2022-05-02
Introduce handle ownership semantics

nirinchev:
This is a refactoring of our native handle hierarchy. A summary of the changes:

* Introduces a method to release the native pointer to `Handle<T>`
  * The unowned version of the handle will not attach a native finalizer and the `release` method is a no-op.
* Adds `OwnableHandle<T>` that is owned by a Realm and it's lifetime is at most as long as the lifetime of the parent Realm.
* Modifies all the realm types (list, results, object, query) to inherit from `OwnableHandle` and modifies their ctors to optionally pass in their parent Realm as root in case it's unowned.
* `RealmHandle` cleans up all its children when released.
* Modifies the scheduler not to close its parent realm when freed - I'm not sure why this was the behavior in the first place, but I don't think it was intentional/useful.

There are some deficiencies with the current implementation:
* The `RealmHandle` keeps strong references to all its children. This is less than ideal, but there's no way to avoid it until we get weak references with dart 2.17.
  * This should not be a huge problem because the Realm is explicitly closed as soon as the callback is over.

### DIFF
--- a/lib/src/realm_class.dart
+++ b/lib/src/realm_class.dart
@@ -208,13 +208,18 @@ class Realm {
   /// All [RealmObject]s and `Realm ` collections are invalidated and can not be used.
   /// This method will not throw if called multiple times.
   void close() {
+    if (isClosed) {
+      return;
+    }
+
     realmCore.closeRealm(this);
     handle.release();
-    scheduler.stop();
+
+    _scheduler.stop();
   }
 
   /// Checks whether the `Realm` is closed.
-  bool get isClosed => realmCore.isRealmClosed(this);
+  bool get isClosed => _handle.isReleased || realmCore.isRealmClosed(this);
 
   /// Fast lookup for a [RealmObject] with the specified [primaryKey].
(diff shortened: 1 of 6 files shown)


lib/src/realm_class.dart
```diff
@@ -67,7 +67,7 @@ class Realm {
 
   /// Opens a `Realm` using a [Configuration] object.
   Realm(Configuration config) : _config = config {
-    _scheduler = Scheduler(close);
```
nirinchev, line 70: Not sure why the scheduler was closing the Realm before - seems unnecessary.

blagoev (changes requested):
I will really need a discussion about this. Seems a bit overhaul our current architecture quite a bit which really bothers me. Why are we making such drastic changes to the internals. Lets have a discussion about this.

nielsenko (approved):
I don't think this PR is that intrusive, and it doesn't touch the public api. I think we should include this before MDB world.

PR #542, opened 2022-05-04
Prepare a workflow for publishing releases

nirinchev:
This implements a workflow that uploads binaries to S3 and optionally to pub.dev. It roughly does the following:

* Prepare packages:
  1. Downloads binaries from PR build
  1. Infers the version
    a. [Prod] version is equal to whatever is in pubspec.yaml
    a. [Staging] version is equal to whatever is in pubspec.yaml + commit sha
  1. Runs the `archive` command to produce the platform-specific archives
  1. Updates all references to `realm_common` and `realm_generator`
    a. [Prod] those point to the upcoming version
    a. [Staging] those point to `../realm_common` and `../realm_generator`
  1. Removes `publish_to: none ` from all packages
  1. Prepares the various packages in their respective folders and runs `dart pub publish --dry run` on those
  1. Uploads the release folder containing all the packages as well as all the binaries to S3 (`static.realm.io/downloads/dart/version`)
* Publish packages (only prod):
  1. Downloads the release folder
  1. Uploads all packages to pub.dev
  1. Merges the existing release PR
  1. Publishes a github release
  1. Posts to #realm-releases

This introduces some changes to `install_command.dart` and `init.dart`:
1. Reworked `install_command` to remove most of the code duplication
2. Removed some unused arguments from init.dart to make it more obvious where locations are
3. `install_command` now works with file references. It will still attempt to download binaries for those unless `publish_to: none` is specified in the reference pubspec.yaml
4. `install_command` now downloads packages from S3 rather than github releases.

Example S3 url for the packages: https://s3.amazonaws.com/static.realm.io/downloads/dart/0.3.0-37c7a6092191f97dcff467dca4bfe8be9405d77c/packages.tar.gz

Example S3 url for the binaries: https://s3.amazonaws.com/static.realm.io/downloads/dart/0.3.0-37c7a6092191f97dcff467dca4bfe8be9405d77c/macos.tar.gz

Example Github release: https://github.com/realm/realm-dart/releases/tag/untagged-a796952aac036395c582

Example slack post: https://mongodb.slack.com/archives/CMMLKL5SB/p1651685930973989

TODO:
- [x] In ci.yml: update `todo-update-me-to-release` to `release`
- [x] In prepare-release.yml: update `test-master` and `test-release` to `master` and `release`.
- [x] In publish-release.yml: Remove --dry-run from `Publish packages to pub.dev` step
- [x] In publish-release.yml: Update `Publish Github Release.commit` to `master`
- [x] In publish-release.yml: Update `Publish Github Release.draft` to `false`

Follow-up tasks:
- [ ] Update ci.yml to parallelize all architectures once https://github.com/realm/realm-dart/pull/514 is merged
- [ ] Prepare a workflow to update the standard text of changelog.md after merging (it's a bit of an annoyance so not worth delaying this PR).
- [ ] Update realm/realm_dart references in pubspec.ymls to also point to the released version (these are the two examples).

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -20,7 +20,7 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
       with:
-        submodules: recursive
+        submodules: false
 
     - name: Read version
       id: get-version
@@ -103,6 +103,7 @@ jobs:
     - name: Cleanup binary symlinks
       run: |
         rm -rf flutter/realm_flutter/android/src/main/cpp/lib
+        rm flutter/realm_flutter/ios/src
         rm flutter/realm_flutter/ios/realm_dart.xcframework
         rm flutter/realm_flutter/linux/binary
         rm flutter/realm_flutter/windows/binary
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -20,7 +20,7 @@ jobs:
     - name: Checkout code
       uses: actions/checkout@v2
       with:
-        submodules: false
+        submodules: recursive
 
     - name: Read version
       id: get-version
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -27,8 +27,6 @@ jobs:
       run: |
         pkgVersion=$(sed -ne "s/^version: \(.*\)/\1/p" pubspec.yaml)
         pkgSuffix="${{ github.event.inputs.environment != 'Production' && format('-{0}', github.sha) || '' }}"
-        echo $pkgVersion
-        echo $pkgSuffix
         echo "::set-output name=version::$pkgVersion$pkgSuffix"
       shell: bash
 
@@ -197,21 +195,6 @@ jobs:
         dart pub publish --dry-run --directory realm || true
       working-directory: release/${{ needs.prepare-packages.outputs.version }}
 
-    # - name: Configure Docs Credentials
-    #   uses: aws-actions/configure-aws-credentials@v1
-    #   with:
-    #     aws-access-key-id: ${{ secrets.DOCS_S3_ACCESS_KEY }}
-    #     aws-secret-access-key: ${{ secrets.DOCS_S3_SECRET_KEY }}
-    #     aws-region: us-east-2
-    # - name: Upload docs
-    #   run: |
-    #     Expand-Archive -Path Realm/packages/Docs.zip/Docs.zip -DestinationPath Realm/packages
-    #     $versions = "${{ steps.get-version.outputs.version }}", "latest"
-    #     Foreach ($ver in $versions)
-    #     {
-    #       aws s3 sync --acl public-read "${{ github.workspace }}\Realm\packages\_site" s3://realm-sdks/realm-sdks/dotnet/$ver/
-    #     }
-
     - name: Find Release PR
       uses: juliangruber/find-pull-request-action@f9f7484f8237cf8485e5ab826e542ba5dd9e9c6e
       id: find-pull-request
@@ -236,49 +219,6 @@ jobs:
         tag: ${{ needs.prepare-packages.outputs.version }}
         token: ${{ secrets.REALM_CI_PAT }}
         draft: true
-
-    # - name: Update Changelog
-    #   run: |
-    #     echo "## vNext (TBD)
-
-    #     ### Enhancements
-    #     * None
-
-    #     ### Fixed
-    #     * None
-
-    #     ### Compatibility
-    #     * Realm Studio: 11.0.0 or later.
-
-    #     ### Internal
-    #     * Using Core x.y.z.
-    #     " | cat - CHANGELOG.md >> temp
-    #     mv temp CHANGELOG.md
-
-    # - name: git status
-    #   run: |
-    #     git status
-
-    # - name: Create vNext PR
-    #   id: vnext-pr
-    #   uses: peter-evans/create-pull-request@7380612b49221684fefa025244f2ef4008ae50ad
-    #   with:
-    #     branch: prepare-vnext
-    #     # TODO: update to master
-    #     base: test-master
-    #     title: Prepare for vNext
-    #     draft: false
-    #     body: Update Changelog for vNext
-    #     commit-message: Prepare for vNext
-    #     token: ${{ secrets.REALM_CI_PAT }}
-    #     delete-branch: true
-
-    # - name: Merge Pull Request
-    #   uses: juliangruber/merge-pull-request-action@8a13f2645ad8b6ada32f829b2fae9c0955a5265d
-    #   with:
-    #     github-token: ${{ secrets.REALM_CI_PAT }}
-    #     number: ${{ steps.vnext-pr.outputs.pull-request-number }}
-    #     method: squash
 
     - name: 'Post to #realm-releases'
       uses: realm/ci-actions/release-to-slack@v3
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -51,7 +51,7 @@ jobs:
 
         for directory in binary/*; do
           targetFile="release/${directory#*-}.tar.gz"
-          dart run realm_dart archive --source-dir "${{ github.workspace }}/binary/$directory" --output-file "${{ github.workspace }}/$targetFile"
+          dart run realm_dart archive --source-dir "${{ github.workspace }}/$directory" --output-file "${{ github.workspace }}/$targetFile"
         done
 
     - name: Update realm_* path dependencies (Production)
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -82,7 +82,6 @@ jobs:
         replace: 'version: ${{ steps.get-version.outputs.version }}'
         include: '**pubspec.yaml'
 
-
     - name: Remove pubspec.yaml/publish_to:none
       uses: jacobtomlinson/gha-find-replace@b76729678e8d52dadb12e0e16454a93e301a919d #! 2.0.0
       with:
@@ -168,7 +167,7 @@ jobs:
     if: ${{ github.event.inputs.environment == 'Production' }}
     environment:
       name: 'Production'
-      url: ${{ steps.upload-to-s3.outputs.url }}
+      url: https://pub.dev/packages/realm/versions/${{ needs.prepare-packages.outputs.version }}
     name: Publish release
     needs:
     - prepare-packages
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -47,14 +47,12 @@ jobs:
     # as the artifact name - i.e. librealm-linux, etc.
     - name: Archive binaries
       run: |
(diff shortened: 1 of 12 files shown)


nielsenko (approved):
LGTM

desistefanova (approved):

.github/workflows/publish-release.yml
```diff
@@ -3,22 +3,227 @@ on:
...
+        dart pub publish --dry-run --directory realm || true
+      working-directory: release/${{ needs.prepare-packages.outputs.version }}
+
+    - name: Find Release PR
+      uses: juliangruber/find-pull-request-action@f9f7484f8237cf8485e5ab826e542ba5dd9e9c6e
+      id: find-pull-request
+      with:
+        branch: ${{ github.ref }}
+
+    - name: Merge Pull Request
```
blagoev, line 204: Should these steps be checking a success status of something before it. Unsure what will happen if we fail to upload binaries to S3 and we merge to master this the release PR

.github/workflows/publish-release.yml
```diff
@@ -3,22 +3,227 @@ on:
...
+      uses: juliangruber/merge-pull-request-action@8a13f2645ad8b6ada32f829b2fae9c0955a5265d
+      with:
+        github-token: ${{ secrets.REALM_CI_PAT }}
+        number: ${{ steps.find-pull-request.outputs.number }}
+        method: squash
+
+    - name: Publish Github Release
```
blagoev, line 211: The same here. Should we be executing these steps only if upload binaries succeeded.

lib/src/cli/install/install_command.dart
```diff
@@ -237,195 +125,82 @@ class InstallCommand extends Command<void> {
...
-  }
-
-  Future<void> saveVersionFile(Directory destinationDir, Pubspec realmPubspec) async {
-    var versionFile = File(path.join(destinationDir.absolute.path, versionFileName));
+    final versionFile = File(path.join(destinationDir.absolute.path, _versionFileName));
     await versionFile.writeAsString("${realmPubspec.version}");
   }
 
-  Future<String> getRealmPackagePath() async {
+  Future<String> _getRealmPackagePath() async {
```
blagoev, line 133: Making all these methods and fields private only hurts readability imo. There no reason to hide these since we are not using this class anywhere else so we need to care of its public interface.

lib/src/cli/install/install_command.dart
```diff
@@ -237,195 +125,82 @@ class InstallCommand extends Command<void> {
...
-  FutureOr<void>? runInternal() async {
-    options = parseOptionsResult(argResults!);
-    validateOptions();
+    final realmPackagePath = await _getRealmPackagePath();
+    final realmPubspec = await _parseRealmPubspec(realmPackagePath);
 
-    if (await skipInstall()) {
-      print("Realm install command started from within the realm-dart repo or one of the realm packages is a path dependency or has a path dependency override."
-          " Skipping install.");
+    if (realmPubspec.publishTo == 'none') {
```
blagoev, line 175: The previous implementation checked the package names in skipInstall as well. This one only checks publish to for skiping install. This makes this susceptible in the future the check to fail if someone forgets and commits in any realm package the `pubspec.yaml` file without `publish_to: none`. 
I suggest we keep the `skipInstall` implementation which also checked the package name

lib/src/cli/archive/archive_command.dart
```diff
@@ -47,12 +47,10 @@ class ArchiveCommand extends Command<void> {
       abort("source-dir option not specified");
     }
 
-    if (options.outputFile == null) {
-      abort("output-file option not specified");
-    }
+    final outputFile = options.outputFile ?? '${options.sourceDir}.tar.gz';
```
blagoev, line 50: Output file is marked as required. So we should be validating it here as before. Or if we really want to make it optional than we need to change `options.dart`. Imo being explicit here only helps, cause this command is an internal one and not something users will be using. So being explicit here only helps us - where goes what when reading the various cmake/gh actions scripts.

lib/src/cli/install/install_command.dart
```diff
@@ -237,195 +125,82 @@ class InstallCommand extends Command<void> {
...
+      print("Referencing $_packageName@${realmPubspec.version} which hasn't been published. Skipping download.");
```
blagoev, line 176: This error message is misleading. When we publish a version we are going to keep the `publish_to: none` in the repo.

lib/src/cli/install/install_command.dart
blagoev: Also I am not sure why are we allowing path dependencies for `realm` and `realm_dart` packages. What is the benefit here. Imo we should be checking for that as the previous version since this is most certainly a bug or we are the ones executing the install command. In such cases install should not happen

blagoev (changes requested):

.github/workflows/publish-release.yml
nirinchev: The job fails it any of its steps fails, so we won't get this far if something earlier fails.

lib/src/cli/install/install_command.dart
nirinchev: The idea is to simplify using prerelease binaries here. Since we're not uploading those to pub.dev, the only other way for people to reference it is as a path dependency. I'd rather have us support that scenario, even if it's for internal use. It's a lot simpler to tell docs/product to download and unzip a package and then run install than to tell them where to unzip the binaries for each platform.

lib/src/cli/install/install_command.dart
nirinchev: This is intentional - we don't want to download binaries if we're working with the dart repo. When we upload the packages to S3, we remove the `publish_to: none`, so if you download a package from S3, you can reference it by path and run the install command to download the binaries.

lib/src/cli/install/install_command.dart
nirinchev: I agree - I was wondering what the best practice was around this. The main reason why I did it was that there was some unused method that wasn't flagged by the compiler because it was public. But now that they're all used, I can make them public again.

lib/src/cli/install/install_command.dart
nirinchev: I can bring back the package name check - not sure it brings a whole lot value, but it doesn't hurt either.

lib/src/cli/install/install_command.dart
blagoev: Not sure what do you mean. How are we uploading the packages to S3. `dart pub publish` does not do that.

flutter/realm_flutter/linux/CMakeLists.txt
```diff
@@ -37,7 +37,7 @@ include(${EPHEMERAL_DIR}/generated_config.cmake)
 execute_process(COMMAND "${FLUTTER_ROOT}/bin/flutter" "pub" "run" "realm" "install" "--target-os-type" "linux" "--package-name" "realm" #"--debug"
   OUTPUT_VARIABLE output
   RESULT_VARIABLE result
-  COMMAND_ERROR_IS_FATAL ANY
+  # COMMAND_ERROR_IS_FATAL ANY
```
nirinchev, line 40: @blagoev and @fealebenpae - I commented this out to allow me to reduce the minimum required cmake version to 3.10 which is in line with the `CMakeLists` generated by `flutter create`. I think it's a bit annoying to require such a high version of cmake because it means developers have to jump through hoops to update it - e.g. on ubuntu 20.04, the highest you can get _easily_ is 3.16. Let me know if you're opposed to this change or if there are better ways to do it.

nirinchev:
@blagoev I've addressed your comments and tested the S3 package as a path dependency on all supported platforms and it seems to work. This should be ready for a final review.

flutter/realm_flutter/linux/CMakeLists.txt
blagoev: I think we should keep the required newer cmake version. We shouldn't be sticking to what `flutter create` does since these templates are not updated regularly but we indeed require newer cmake functionality. This is the same with cocoapods. 
If we need to make it even easier for developers we could update our readme with more details how to install these.

nirinchev: Is that parameter that important though? Can we replace it with something and keep the functionality with the lower cmake version?

nirinchev: As far as I can tell, this only matters if we're running multiple commands which we're not.

lib/src/cli/install/install_command.dart
blagoev: I added a discussion point for this. I think we can handle this use case better

flutter/realm_flutter/linux/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.19)
+cmake_minimum_required(VERSION 3.10)
```
blagoev, line 1: we require Cmake 3.21 now. https://github.com/realm/realm-dart#environment-setup-for-realm-flutter
I think we should keep the requirement.

blagoev (changes requested):

flutter/realm_flutter/linux/CMakeLists.txt
blagoev: I have added a discussion point for this.

flutter/realm_flutter/linux/CMakeLists.txt
nirinchev: This is different - we're not building Core here, so there's no reason to use 3.21.

flutter/realm_flutter/linux/CMakeLists.txt
fealebenpae: Can we have a flutter plugin without this native code that compiles at a user's machine? Our native code is pre-compiled as a shared library, so these dummy plugins do nothing of value.

flutter/realm_flutter/linux/CMakeLists.txt
fealebenpae: Yeah, that's just the dummy flutter plugin - it doesn't build anything except the "hello world" plugin code. it doesn't need the newer CMake version.

flutter/realm_flutter/linux/CMakeLists.txt
nirinchev: This downloads the native binary though, so we definitely need it.

blagoev:
@nirinchev the discussion could fix something in a folloupPR, so lets merge this one

blagoev (approved):

flutter/realm_flutter/linux/CMakeLists.txt
blagoev: yes indeed

flutter/realm_flutter/linux/CMakeLists.txt
fealebenpae: I can't believe I'm saying this, but npm is better. In that case I agree with @nirinchev - it makes sense to stick to a lower CMake version, ideally the one that Flutter itself depends on, since we aren't actually compiling anything more complicated than the default plugin package template here.

PR #544, opened 2022-05-05
tests: reset password

desistefanova:
Create tests for email/password reset password and call reset password function.
Enable Flutter tests to be run on the CI for Windows and Linux.
Enable calling reset password function without passing additional arguments.


test/test.dart
```diff
@@ -112,12 +112,15 @@ void xtest(String? name, dynamic Function() testFunction) {
...
 Future<void> setupTests(List<String>? args) async {
   parseTestNameFromArguments(args);
+  
+  setUpAll(() async {
+    await setupBaas();
+  });
 
-  await setupBaas();
-  setUp(() {
+  setUp(() async {
```
nielsenko, line 120:
```suggestion
  setUp(() {
```

test/credentials_test.dart
```diff
@@ -228,4 +227,56 @@ group("Manual test: Email/Password - confirm user", () {
...
+      final app = App(configuration);
+      final authProvider = EmailPasswordAuthProvider(app);
+      String newPassword = "RWE@#EDE";
+      await authProvider.completeResetPassword(newPassword, token, tokenId);
+      final user = await loginWithRetry(app, Credentials.emailPassword(validUsername, strongPassword), 3);
+      expect(user, isNotNull);
+    }, appName: "emailConfirm", skip: "Run this test manually after test 1 and after setting token and tokenId");
+  });
+
+  // baasTest('Email/Password - call reset password function', (configuration) async {
```
nielsenko, line 273: Perhaps remove this out-commented code

nielsenko (approved):

test/test.dart
desistefanova: Yes, I have added it probably while I was trying to call `await setupBaas()` in setUp

test/credentials_test.dart
desistefanova: This PR is draft. I'm going to continue implementing the commented test.

.github/workflows/ci.yml
```diff
@@ -118,7 +118,14 @@ jobs:
   flutter-linux:
     runs-on: ubuntu-latest
     name: Flutter Tests Linux
+    env:
+      BAAS_URL: ${{ secrets.REALM_QA_URL }}
```
nirinchev, line 122: These should not be reusing clusters setup for other test runs. Every time we want a test setup, we should create a separate cluster, otherwise we risk tests interfering with one another.

test/credentials_test.dart
```diff
@@ -226,4 +226,119 @@ Future<void> main([List<String>? args]) async {
...
+      // Make sure you have recieved an emails with hyperlink ResetPassword.
+      // Find the token and tokenId in the query parameters of the link received in the email and enter them in the following variables.
+      String token = "fb485146b15497209a9d1b67128ae29199cdff2c26f389e0ee5d52ae6bc4228e5738b29256eade976e24767804bfb4dc68198075e3a461cd4f73864901fb09be";
+      String tokenId = "6272619334f5b3a770a7dc2c";
+
+      final app = App(configuration);
+      final authProvider = EmailPasswordAuthProvider(app);
+      String newPassword = "RWE@#EDE";
+      await authProvider.completeResetPassword(newPassword, token, tokenId);
+      final user = await loginWithRetry(app, Credentials.emailPassword(validUsername, strongPassword));
```
nirinchev, line 286: Does this require `loginWithRetry`? My expectation is that the password change propagates immediately.

test/credentials_test.dart
```diff
@@ -226,4 +226,119 @@ Future<void> main([List<String>? args]) async {
...
+
+  // The tests in this group are for manual testing, since they require interaction with mail box.
+  // Please enter a valid data in the variables under comments.
+  // The email should be a valid and existing one in order you to be able to receive automatic emails.
+  // Run first two steps (test 1 and test 2) to create and confirm the user.
+  // Before running test 2 set the variables with token details received as link query parameters in the confirmation email.
+  // Then run test 3 to call reset password function configured in the cloud.
+  // Test 3 will check also whether the new password is applied and the old password is rejected.
+  // Run the last test 4 to call resetFunc without passing additional arguments.
+  group("Manual test: Email/Password - call reset password function with new password", () {
```
nirinchev, line 299: Why do this need to be manual? As far as I can tell we only have a manual step for confirming the user which can be achieved either by using automatic user confirmation or by using function confirmation.

nirinchev (commented):
Tests seem fine to me, but I think we can enable the function reset tests to run on CI and we also need to setup baas for more platforms. One thing you can do (or I can do it in a separate PR) is to move this to a matrix job so that it's not that much copy-paste.

test/credentials_test.dart
desistefanova: I'm not sure. Where can I check it?

test/credentials_test.dart
desistefanova: You are right. I will change it.

test/credentials_test.dart
nirinchev: I don't think we have docs/way to check that - my suggestion would be to just use `app.logIn` directly rather than the version that retries.

desistefanova: Is it only confirmation process that takes more time to be able to login?

nirinchev: In my experience, it's confirmation when autoconfirm is set to `true` or when using a confirmation function. I guess the reason there is that there's no explicit "confirm" call that you invoke, so the `register` call does not wait for confirmation to run. Whenever you have a "completeXXX" method, that typically completes when the change has been persisted and propagated.

test/credentials_test.dart
desistefanova: Done

blagoev (approved):

.github/workflows/ci.yml
desistefanova: Created issue RDART-367.

PR #557, opened 2022-05-08
Fix boolean

### LINKED ISSUE #474: type bool bug
# 20220420 Flutter Test

Mac mini (M1)

```
flutter doctor                                                                                             
Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 2.10.4, on macOS 12.3.1 21E258 darwin-arm, locale zh-Hans-CN)
[✓] Android toolchain - develop for Android devices (Android SDK version 32.1.0-rc1)
[✓] Xcode - develop for iOS and macOS (Xcode 13.3.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2021.1)
[✓] VS Code (version 1.66.2)
[!] Proxy Configuration
    ! NO_PROXY is not set
[✓] Connected device (3 available)
[!] HTTP Host Availability
    ✗ HTTP host https://cocoapods.org/ is not reachable. Reason: An error occurred while checking the HTTP host: Connection reset by
      peer

! Doctor found issues in 2 categories.
```

[test_demo](https://github.com/FlutterTrip/issues/tree/main/test_flutter)

The same bug as dart

gif

---

Hi
ver ^0.2.1+alpha

``` shell
$ dart --version
Dart SDK version: 2.16.2 (stable) (Tue Mar 22 13:15:13 2022 +0100) on "macos_x64"
```

## Step

- `dart pub add realm_dart`
- `dart run realm_dart install`

then run

## bug

img

code:

```
realm.write(() {
    /*
      @RealmModel()
class _TestModel {
  @PrimaryKey()
  late int primaryKey = 0;
  bool test1 = true;
  bool test2 = false;
}
    */
    realm.add(TestModel()); // default model
    realm.add(TestModel(primaryKey: 1));
    realm.add(TestModel(test1: false, test2: true, primaryKey: 2));
    realm.add(TestModel(test1: true, test2: false, primaryKey: 3));
  });
```
type  `bool`

添加进去的数据，和最终取出来的相反
The added data is the opposite of what is finally taken out

> code: test1 = false, realm file test1 = true 

[test_demo](https://github.com/FlutterTrip/issues/tree/main/test)

blagoev:
Fixes boolean value persistence
fixes #474

### DIFF
--- a/test/realm_object_test.dart
+++ b/test/realm_object_test.dart
@@ -62,6 +62,9 @@ class _RemappedFromAnotherFile {
 
 @RealmModel()
 class _BoolValue {
+  @PrimaryKey()
+  late int key;
+
   late bool value;
 }
 
@@ -368,11 +371,11 @@ Future<void> main([List<String>? args]) async {
     var realm = getRealm(config);
 
     realm.write(() {
-      realm.add(BoolValue(true));
-      realm.add(BoolValue(false));
+      realm.add(BoolValue(1, true));
+      realm.add(BoolValue(2, false));
     });
 
-    expect(realm.all<BoolValue>()[0].value, true);
-    expect(realm.all<BoolValue>()[1].value, false);
+    expect(realm.find<BoolValue>(1)!.toJson().replaceAll('"', '').contains("value:true"), isTrue);
+    expect(realm.find<BoolValue>(2)!.toJson().replaceAll('"', '').contains("value:false"), isTrue);
   });
 }
--- a/test/realm_object_test.dart
+++ b/test/realm_object_test.dart
@@ -60,6 +60,11 @@ class _RemappedFromAnotherFile {
   late $RemappedClass? linkToAnotherClass;
 }
 
+@RealmModel()
+class _BoolValue {
+  late bool value;
+}
+
 Future<void> main([List<String>? args]) async {
   print("Current PID $pid");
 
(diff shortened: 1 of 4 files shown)


test/realm_object_test.dart
```diff
@@ -357,4 +362,17 @@ Future<void> main([List<String>? args]) async {
...
+    var config = Configuration([BoolValue.schema]);
+    var realm = getRealm(config);
+
+    realm.write(() {
+      realm.add(BoolValue(true));
+      realm.add(BoolValue(false));
+    });
+
+    expect(realm.all<BoolValue>()[0].value, true);
+    expect(realm.all<BoolValue>()[1].value, false);
```
nirinchev, line 376: Realm does not guarantee ordering of tables. In most cases, the order will be the insertion order, but we can't depend on this behavior not changing.

nirinchev (commented):
The changes seem fine to me, but I don't think the test is actually validating the fix - i.e. I expect the same test to have passed with the previous implementation due to the symmetric nature of the bug. A more robust test can be written using the `to_json` method on RealmObject and validating that the serialized json value matches expectations.

desistefanova (approved):
Agree with @nirinchev about items order could not be the same as the insertion order.

test/realm_object_test.dart
blagoev: ok I will redo using primary keys. 
also I have verified the fix manually.

blagoev:
Insertion order is pretty much set in stone now. I can show you examples of tests depending on this behavior in every SDK.

PR #566, opened 2022-05-11
App config log level and request timeout

### LINKED ISSUE #372: Sync Client Config: set log level

### LINKED ISSUE #376: Sync Client Config: set metadata mode

desistefanova:
Sync Client Config: set log level. Fixes #372 
Sync Client Config: support connection timeout. Fixes #376

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,6 @@
 * Support user state. ([#525](https://github.com/realm/realm-dart/pull/525))
 * Support getting user id and identities. ([#525](https://github.com/realm/realm-dart/pull/525))
 * Support user logout. ([#525](https://github.com/realm/realm-dart/pull/525))
-.
 ### Fixed
 * Fixed an issue that would result in the wrong transaction being rolled back if you start a write transaction inside a write transaction. ([#442](https://github.com/realm/realm-dart/issues/442))
 * Fixed boolean value persistence ([#474](https://github.com/realm/realm-dart/issues/474))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 * Support ObjectId type. ([#468](https://github.com/realm/realm-dart/pull/468))
 * Support Uuid type. ([#470](https://github.com/realm/realm-dart/pull/470))
 * Support application login. ([#469](https://github.com/realm/realm-dart/pull/469))
+* Support app configuration log level and request timeout.
 * Support EmailPassword register user. ([#452](https://github.com/realm/realm-dart/pull/452))
 * Support EmailPassword confirm user. ([#478](https://github.com/realm/realm-dart/pull/478))
 * Support EmailPassword resend user confirmation email. ([#479](https://github.com/realm/realm-dart/pull/479))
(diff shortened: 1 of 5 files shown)


test/app_test.dart
```diff
@@ -43,15 +46,47 @@ Future<void> main([List<String>? args]) async {
...
     expect(b.baseFilePath.path, Directory.systemTemp.path);
     expect(b.baseUrl, Uri.parse('https://not_re.al'));
     expect(b.defaultRequestTimeout, const Duration(seconds: 2));
+    expect(b.logLevel, LogLevel.info);
+    expect(b.metadataPersistenceMode, MetadataPersistenceMode.disabled);
+    expect(b.requestTimeout, 1000);
     expect(b.httpClient, httpClient);
   });
 
+  test('AppConfiguration can be created and an App could be created', () {
```
blagoev, line 64: we have a test `app can be created`. I suggest we name this `AppConfiguration can be created` only since this tests all properties of the config.

CHANGELOG.md
```diff
@@ -32,7 +33,7 @@
 * Support user state. ([#525](https://github.com/realm/realm-dart/pull/525))
 * Support getting user id and identities. ([#525](https://github.com/realm/realm-dart/pull/525))
 * Support user logout. ([#525](https://github.com/realm/realm-dart/pull/525))
-
+.
```
blagoev, line 36:
```suggestion

```

lib/src/app.dart
```diff
@@ -101,6 +143,8 @@ class AppConfiguration {
     this.localAppVersion,
     this.metadataEncryptionKey,
     this.metadataPersistenceMode = MetadataPersistenceMode.plaintext,
+    this.logLevel = LogLevel.error,
+    this.requestTimeout = 0,
```
blagoev, line 147: Is this intentional? Should we be setting `requestTimeout` to zero ?

blagoev: I saw the core invocation below. I think this will be misleading if one reads the values when the config is created without one. I think we should either make this nullable or copy the default value Core uses.

blagoev (changes requested):

test/app_test.dart
desistefanova: We already have a test with name "AppConfiguration can be created" but it initializes only the appConfig in Dart. I renamed it to 'AppConfiguration can be initialized'. Then renamed this test to 'AppConfiguration can be created', since it really creates the native app configuration.

lib/src/app.dart
desistefanova: There was a check in realm_core to be left by default in case it is set to 0.
But you are right, it is confusing. So that I set it to the default value which is 2 min.

test/app_test.dart
```diff
@@ -43,14 +46,58 @@ Future<void> main([List<String>? args]) async {
...
+      metadataPersistenceMode: MetadataPersistenceMode.disabled,
+      logLevel: LogLevel.info,
+      maxConnectionTimeout: const Duration(minutes: 1),
       httpClient: httpClient,
     );
     expect(b.appId, 'myapp1');
     expect(b.baseFilePath.path, Directory.systemTemp.path);
     expect(b.baseUrl, Uri.parse('https://not_re.al'));
     expect(b.defaultRequestTimeout, const Duration(seconds: 2));
+    expect(b.logLevel, LogLevel.info);
```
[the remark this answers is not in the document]
desistefanova: OK

nielsenko (approved):

PR #568, opened 2022-05-12
Temporarely disable coverage reports

nielsenko:


blagoev (approved):
why not disable the whole step

PR #569, opened 2022-05-15
add datetime support

blagoev:
supersedes  #519

### DIFF
--- a/lib/src/native/realm_core.dart
+++ b/lib/src/native/realm_core.dart
@@ -1351,6 +1351,9 @@ Pointer<realm_value_t> _toRealmValue(Object? value, Allocator allocator) {
   return realm_value;
 }
 
+const int _microsecondsPerSecond = 1000 * 1000;
+const int _nanosecondsPerMicrosecond = 1000;
+
 void _intoRealmValue(Object? value, Pointer<realm_value_t> realm_value, Allocator allocator) {
   if (value == null) {
     realm_value.ref.type = realm_value_type.RLM_TYPE_NULL;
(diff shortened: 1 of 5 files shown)


nielsenko:
Why not just continue with #519?

nirinchev:
We could have just rebased on #519 on master but I guess it doesn't matter that much.

lib/src/native/realm_core.dart
```diff
@@ -1431,7 +1445,9 @@ extension on Pointer<realm_value_t> {
       case realm_value_type.RLM_TYPE_BINARY:
         throw Exception("Not implemented");
       case realm_value_type.RLM_TYPE_TIMESTAMP:
-        throw Exception("Not implemented");
+        final seconds = ref.values.timestamp.seconds;
+        final nanoseconds = ref.values.timestamp.nanoseconds;
+        return DateTime.fromMicrosecondsSinceEpoch(seconds * _microsecondsPerSecond + nanoseconds ~/ _nanosecondsPerMicrosecond, isUtc: true);
```
desistefanova, line 1450: I think there should be `.toLocal()` at the end. When we pass the value to core we convert it toUtc. Here we are receiving utc from core and we have to convert it back to local before to return.

nirinchev: I don't think we should be converting dates to local because Core doesn't store timezone info. It'll be misleading for developers if we returned local date here. Also, there's no guarantee that the original date the developer passed was in local time - it could be UTC or some completely different timezone.

test/realm_object_test.dart
```diff
@@ -68,6 +69,20 @@ class _BoolValue {
   late bool value;
 }
 
+extension on DateTime {
+  String toNormalizedDateString() {
+    final utc = toUtc();
```
desistefanova, line 74: Why this formatting function changes the date? If the date is local and you want to format it will return utc date.

lib/src/native/realm_core.dart
desistefanova: It is not needed the core to keep time zone. if you pass a local date from one device, then when you convert toUtc the method will use the current time zone of the device. And when they receive it on another device it will be converted to the device local time. 
In case we want the user to manage their dates then we shouldn't convert their dates to UTC in `_intoRealmValue` method without they to know. Probably have to throw if `isUtc==false` before to change their date.

nirinchev: I think it's a little annoying if we force people to always convert dates to utc although I agree it's more correct as we explicitly tell them that Realm will lose the timezone information.

desistefanova: When they create a DateTime in Dart it is local by default unless they have used DateTime.utc constructor. It will be good to have at least some warning or api doc.

test/realm_object_test.dart
nirinchev: This is used for tests only - to verify that the string output by Core matches our expectations. Since Core dates are always utc, it makes sense to always convert it.

blagoev:
@desistefanova @nirinchev Could you have your say on this so we could merge it. I see some comments but I am unsure if I need to take some action on the code. If there are changes needed please write a clear comment on that ?

nirinchev:
I wrote the code so I'm not sure it's me who should comment on whether to merge it or not.

desistefanova (approved):
Just a reminder for documenting DateTime UTC convertion.

nielsenko (approved):

nielsenko:
Needs a rebase though..

PR #570, opened 2022-05-15
User get device id,

blagoev:
User get auth provider type
User get profile data

### DIFF
--- a/lib/src/user.dart
+++ b/lib/src/user.dart
@@ -23,7 +23,7 @@ import 'app.dart';
 import 'credentials.dart';
 import 'realm_class.dart';
 
-/// This class represents a user in a MongoDB Realm app.
+/// This class represents a `user` in a MongoDB Realm app.
 /// A user can log in to the server and, if access is granted, it is possible to synchronize the local Realm to MongoDB.
 /// Moreover, synchronization is halted when the user is logged out. It is possible to persist a user. By retrieving a user, there is no need to log in again.
 /// Persisting a user between sessions, the user's credentials are stored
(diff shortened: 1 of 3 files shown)


nielsenko (approved):

lib/src/user.dart
```diff
@@ -118,15 +133,29 @@ enum UserState {
...
 }
 
+class UserProfile {
+  final String? name;
+  final String? firstName;
+  final String? lastName;
+  final String? email;
+  final String? gender;
+  final String? birthDay;
+  final String? minAge;
```
desistefanova, line 152: What about trying to parse min and max Age to int?

blagoev: .net has these as strings. So I left them the same

test/user_test.dart
```diff
@@ -91,4 +91,29 @@ Future<void> main([List<String>? args]) async {
     final user2 = await app.logIn(Credentials.emailPassword(username, password));
     expect(user1, equals(user2));
   }, appName: AppNames.autoConfirm, skip: "Blocked on https://github.com/realm/realm-core/issues/5467");
+
```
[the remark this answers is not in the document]
blagoev: not sure. I guess it is created on sign in

desistefanova (approved):

lib/src/user.dart
```diff
@@ -118,15 +133,29 @@ enum UserState {
...
   final String id;
 
-  /// The authentication provider defining this identity
+  /// The [AuthProviderType] defining this identity.
   final AuthProviderType provider;
 
   const UserIdentity._(this.id, this.provider);
 }
 
+class UserProfile {
```
nirinchev, line 145: The user profile may hold other properties - not only these. My suggestion would be to change these to be getter-only properties and to expose an indexer on the profile for accessing other properties. Something like:

```dart
class UserProfile {
    final Map<String, dynamic> _data;

    String? get name => this['name'];

    // Rest of named properties

    dynamic operator[](String property) => _data[property];

    UserProfile._(this._data);
}
```

blagoev: I just modeled it after .NET. There are not such methods. But I think it wont hurt adding them. Will do

blagoev:
fixes #421 #422 #424

PR #573, opened 2022-05-16
Run ffigen on CI and fail build if there are changes

nirinchev:


nielsenko (approved):

blagoev (approved):

blagoev:
It will be better if we run this only if there are changes to the Core submodule

nielsenko:
I think we need to install `libclang-11-dev` apt package on the ubuntu host. As I recall ffigen is dependent on this slightly old version of clang.

nirinchev:
The failures are due to a cloud deployment and unrelated to this PR.

@blagoev I agree there's probably room for optimization, but this runs as part of the "generator tests" job and adds 40 seconds to the runtime. Considering the generator job is still fairly fast, it's likely to complete roughly at the same time as the linux test job, which is the fastest test job.

PR #574, opened 2022-05-16
Populate correct path for flexible sync configurations

nirinchev:
This change makes it so we generate a path to the realm based on the app id and the user id. It generates path of the format `*appConfig.basePath*/*app.id*/*user.id*/flx_sync_default.realm`.

I've also removed two previously public properties: `defaultPath` and `filesPath`. The reason is that `defaultPath` no longer makes sense since that'll be different between different configs and `filesPath` is also misleading since the base path of the app is where we're storing realm files for flx sync realms. The correct way to access the path of the configuration is to just construct it and evaluate the `path` property.


blagoev:
@nirinchev I have updated to latest master and removed the Core invoke TODO.

I could not make this green cause I there is this change that hides the app.configuration here https://github.com/realm/realm-dart/commit/eaaf73480fd88946026ad99c69a8dcb9673aee45#diff-3705042816f3bd9d24304192811bed9d5104fa569a36f063a4166cfea97b845a

I am unsure why we are hiding the `AppConfiguration`. Imo we should be still exposing it as before. It contains useful info for the app which can be beneficial to the user at runtime. Otherwise they will need to store this themselves. 

Will postpone this PR until we had a chance to discuss this here or on the meeting

blagoev:
@nirinchev I have updated to latest master and removed the Core invoke TODO.

I could not make this green cause I there is this change that hides the app.configuration here https://github.com/realm/realm-dart/commit/eaaf73480fd88946026ad99c69a8dcb9673aee45#diff-3705042816f3bd9d24304192811bed9d5104fa569a36f063a4166cfea97b845a

I am unsure why we are hiding the `AppConfiguration`. Imo we should be still exposing it as before. It contains useful info for the app which can be beneficial to the user at runtime. Otherwise they will need to store this themselves. 

Will postpone this PR until we had a chance to discuss this here or on the meeting

blagoev (approved):

nirinchev:
CI failures are unrelated and will be addressed separately.

PR #577, opened 2022-05-16
Sync config error handler

### LINKED ISSUE #361: Sync Client Config: set error handler

desistefanova:
FlexibleSyncConfig ErrorHandle implementation. Fixes #361


common/lib/src/realm_types.dart
```diff
@@ -109,6 +97,37 @@ enum SyncErrorCategory {
...
+  SyncErrorCode(String message, this.category, this.code) : super(message);
+}
+
+class SyncError extends RealmError {
+  final SyncErrorCode errorCode;
+  final bool isFatal;
+  final bool isUnrecognizedByClient;
+  final bool isClientResetRequested;
+  final Map<String, String> userInfoMap;
+  final int userInfoLength;
```
nirinchev, line 118: We probably don't need those on the `SyncError` class.

* `isUnrecognizedByClient` does not bring much value to end users.
* If `isClientResetRequested` is `true`, we should throw a special error type that is different from the general-purpose `SyncError` and we should report it via the client reset handler, not the session error callback.
* `userInfoMap` and `userInfoLength` are internal and serve the purpose of propagating some client reset information - they should be translated into strongly typed properties.

desistefanova: It is done as we decided - SessionError extends SyncError extends RealmError. Redundant properties are removed.

src/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ set(HEADERS
 )
 
 add_library(realm_dart SHARED ${SOURCES} ${HEADERS})
-target_compile_definitions(RealmFFIStatic PUBLIC -DRealm_EXPORTS)
+target_compile_definitions(RealmFFIStatic PUBLIC -DRealm_EXPORTS -DREALM_ENABLE_SYNC=1)
```
nielsenko, line 26: I know I did this, but perhaps @fealebenpae can weigh in, if this is the correct way?

nielsenko (approved):

src/CMakeLists.txt
fealebenpae: No, we shouldn't need to do this. What problem does it solve?

desistefanova: OK, I remove -DREALM_ENABLE_SYNC=1 since it is handled in #588

common/lib/src/realm_types.dart
```diff
@@ -79,6 +79,20 @@ class RealmStateError extends StateError implements RealmError {
...
 
+/// An error type that describes a session-level error condition.
+/// /// {@category Sync}
+class SessionError extends SyncError {
+  /// If true the received error is fatal.
+  final bool isFatal;
+
+  SessionError(
+    String message,
+    this.isFatal, {
```
blagoev, line 90: isFatal could have a default value to false, but category should be required imo.

lib/src/native/realm_core.dart
```diff
@@ -190,7 +190,9 @@ class _RealmCore {
         final syncConfigPtr = _realmLib.invokeGetPointer(() => _realmLib.realm_flx_sync_config_new(config.user.handle._pointer));
         try {
           _realmLib.realm_sync_config_set_session_stop_policy(syncConfigPtr, config.sessionStopPolicy.index);
-          _realmLib.realm_config_set_sync_config(configPtr, syncConfigPtr);
+          _realmLib.realm_sync_config_set_error_handler(
+              syncConfigPtr, Pointer.fromFunction(_syncErrorHandlerCallback), config.toPersistentHandle(), nullptr);
```
blagoev, line 194: I believe this should be a weak handle to the handler directly. Why try to persist the config?

lib/src/native/realm_core.dart
```diff
@@ -393,6 +395,17 @@ class _RealmCore {
     return config.shouldCompactCallback!(totalSize, usedSize) ? TRUE : FALSE;
   }
 
+  static void _syncErrorHandlerCallback(Pointer<Void> userdata, Pointer<realm_sync_session> user, realm_sync_error error) {
+    final FlexibleSyncConfiguration? syncConfig = userdata.toObject(isPersistent: true);
```
blagoev, line 399: we should be getting the error handler directly here or nullptr if there was no handler registered

blagoev (changes requested):

common/lib/src/realm_types.dart
desistefanova: Yes, sure.

lib/src/native/realm_core.dart
desistefanova: OK.

lib/src/native/realm_core.dart
desistefanova: I made it persistent, because if it is released we will not be able to receive the callback. This config is created when the app is created and it is expected to be used during the whole lifecycle of the Session.

test/session_test.dart
```diff
@@ -281,6 +282,20 @@ Future<void> main([List<String>? args]) async {
     await uploadData.subscription.cancel();
     await downloadData.subscription.cancel();
   });
+
+  baasTest('SyncSession test error handles', (configuration) async {
```
desistefanova, line 286:
```suggestion
  baasTest('SyncSession test error handler', (configuration) async {
```

test/session_test.dart
```diff
@@ -281,6 +282,20 @@ Future<void> main([List<String>? args]) async {
...
   });
+
+  baasTest('SyncSession test error handler', (configuration) async {
+    final app = App(configuration);
+    final user = await getIntegrationUser(app);
+    final config = Configuration.flexibleSync(user, [Task.schema], sessionErrorHandler: (sessionError) {
+      expect(sessionError.category, SyncErrorCategory.session);
+      expect(sessionError.isFatal, false);
+      expect(sessionError.code, 100);
+      expect(sessionError.message, "Error");
```
desistefanova, line 293:
```suggestion
      expect(sessionError.message, "Simulated sync session error");
```

lib/src/native/realm_core.dart
desistefanova: I need syncConfig, because depends on is_client_reset_requested we caldifferent event handlers. See #608

lib/src/native/realm_core.dart
```diff
@@ -399,6 +401,19 @@ class _RealmCore {
...
+    if (syncConfig == null) {
+      return;
+    }
+    final sessionError = error.toSessionError();
+    syncConfig.sessionErrorHandler!(sessionError);
```
desistefanova, line 410:
```suggestion
    if(syncConfig.sessionErrorHandler!=null) {
        syncConfig.sessionErrorHandler!(sessionError);
    }
```

lib/src/native/realm_core.dart
blagoev: I see

blagoev (approved):

PR #586, opened 2022-05-18
Fix passing environment variable for baas

desistefanova:
- [x] CI.yaml environment variables were not passed correctly.

- [x] IOS and Android don't have access to host environment variables and their Baas tests are not executed: "Skip: BAAS URL not present".

CI.yaml environment variables are moved to the top. 
This PR implements passing the environment variables values as arguments from the host OS to Flutter test driver.


flutter/realm_flutter/tests/test_driver/app_test.dart
```diff
@@ -21,10 +21,21 @@ void main(List<String> args) {
 
     // This single tests runs all Realm tests and reports test run failure if any Realm test fails. Contains all failed tests names.
     test('run all', () async {
-      String result = await driver!.requestData(testCommand, timeout: const Duration(minutes: 30));
+      String testCommandWithArgs = testCommand;
+      testCommandWithArgs += getEnvVariable("BAAS_URL");
```
blagoev, line 25:
```suggestion
      testCommandWithArgs += getArgFromEnvVariable("BAAS_URL");
```
```suggestion
      testCommandWithArgs += getEnvVariable("BAAS_URL");
```

test/test.dart
```diff
@@ -234,27 +240,43 @@ Future<void> tryDeleteFile(FileSystemEntity fileEntity, {bool recursive = false}
   }
 }
 
-void parseTestNameFromArguments(List<String>? arguments) {
-  if (testName != null || arguments == null || arguments.isEmpty) {
-    return;
-  }
+Map<String, String?> testArguments(List<String>? arguments) {
```
blagoev, line 243:
```suggestion
Map<String, String?> parseTestArguments(List<String>? arguments) {
```
```suggestion
Map<String, String?> testArguments(List<String>? arguments) {
```

blagoev, line 243:
```suggestion
Map<String, String?> parseTestArguments(List<String>? arguments) {
```
```suggestion
Map<String, String?> testArguments(List<String>? arguments) {
```

test/test.dart
```diff
@@ -234,27 +240,43 @@ Future<void> tryDeleteFile(FileSystemEntity fileEntity, {bool recursive = false}
...
+  Map<String, String?> testArgs = {};
+    final parser = ArgParser()
```
blagoev, line 245: seems not formatted

blagoev, line 245: seems not formatted

test/test.dart
```diff
@@ -234,27 +240,43 @@ Future<void> tryDeleteFile(FileSystemEntity fileEntity, {bool recursive = false}
...
+      ..addArgument(result, argBaasApiKey)
+      ..addArgument(result, argBaasPrivateApiKey)
+    ..addArgument(result, argBaasProjectId);
+  return testArgs;
+}
 
-  int nameArgIndex = arguments.indexOf("--name");
-  if (nameArgIndex >= 0 && arguments.length > nameArgIndex) {
-    testName = arguments[nameArgIndex + 1];
+extension on Map<String, String?> {
```
blagoev, line 264: what's the benefit of this extension. why not just do regular method
```
addArgument(testArgs, result, "name");
addArgument(testArgs, result, argBaasUrl);
```

blagoev, line 264: What's the benefit of this extension. Why not just use regular method.
```
_addArgument(testArgs, result, "name");
_addArgument(testArgs, result, argBaasUrl);
```
as it is it is exposed in all test files since all import `test.dart`

flutter/realm_flutter/tests/test_driver/app.dart
```diff
@@ -1,32 +1,31 @@
...
       }
-
-      List<String> testArgs = List<String>.empty();
+      List<String> testArgs = [];
       if (testName.isNotEmpty) {
         // Build correct test arguments using the dart test arg name '--name'
-        testArgs = ['--name', testName];
+        testArgs.addAll(['--name', testName]);
       }
+      testArgs.addAll(command.substring(command.indexOf("--")).split(" "));
```
blagoev, line 23: Did you test this with environment where `BASS_` env vars are not defined. Here `indexOf` can return -1. 

We should be able to run the tests locally with no `BAAS_` env vars.

blagoev, line 23: indexOf can return -1. I think we should be guarding against that here. Can we use the ArgParser here as well?

blagoev, line 23: indexOf can return -1. I think we should be guarding against that here. Can we use the ArgParser here as well?

blagoev (changes requested):

PR #587, opened 2022-05-18
Test session error handler with raising fake error

desistefanova:
This PR is only for code review. Probably will not be merged.
Depends on the fix in #588


src/sync_session.cpp
```diff
@@ -125,4 +128,18 @@ RLM_API uint64_t realm_dart_sync_session_register_connection_state_change_callba
 
 } // anonymous namespace
 
+RLM_API void realm_syncsession_report_error_for_testing(realm_sync_session_t* session, uint32_t category, int errorCode, bool isFatal) noexcept
```
blagoev, line 131: we prefix our native methods with `realm_dart_`

blagoev, line 131:
```suggestion
RLM_API void realm_dart_syncsession_report_error_for_testing(realm_sync_session_t* session, uint32_t category, int errorCode, bool isFatal) noexcept
```

lib/src/session.dart
```diff
@@ -71,6 +71,10 @@ class Session {
     final controller = SessionProgressNotificationsController(this, direction, mode);
     return controller.createStream();
   }
+
+  void raiseSessionError(SyncErrorCategory category, int errorCode, bool isFatal) {
```
blagoev, line 75: this should be an internal method.

blagoev (changes requested):

desistefanova:
I'm closing this PR, because it is already merged into [Sync config error handler #577](https://github.com/realm/realm-dart/pull/577)

src/sync_session.cpp
desistefanova: Done in https://github.com/realm/realm-dart/pull/577

lib/src/session.dart
desistefanova: Done in https://github.com/realm/realm-dart/pull/577

PR #588, opened 2022-05-18
Headers cleanup and ffigen symbol addresses

fealebenpae:
Several big things:

- Fold ffigen's `config.yaml` into its `pubspec.yaml` so we can just run `dart run ffigen` to generate bindings.
- Expose symbol addresses all functions starting with `realm_dart_` in the generated bindings.
  - I renamed all realm-dart functions to begin with `realm_dart`, which should be our convention.
- Remove the header symlinks from the ffigen folder and use relative paths to the real headers.
- Got rid of the separate `realm_dart_get_files_path()` definition for iOS and Android and declared it into `realm_dart.h`.
- Cleaned up header includes and changed some quote includes to angle bracket includes.
- Remove the `DllMain` entry point for Windows - we don't really need it.


src/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ set(HEADERS
 add_library(realm_dart SHARED ${SOURCES} ${HEADERS})
 target_compile_definitions(RealmFFIStatic PUBLIC -DRealm_EXPORTS)
 
-target_link_libraries(realm_dart dart-dl RealmFFIStatic Realm::Storage)
+target_link_libraries(realm_dart dart-dl RealmFFIStatic Realm::ObjectStore)
```
nielsenko, line 26: Curious about this change. What is the purpose?

nielsenko (approved):

src/CMakeLists.txt
fealebenpae: Ah, that's related to the discussion in https://github.com/realm/realm-dart/pull/577#discussion_r875994306.

blagoev (approved):

desistefanova (approved):

PR #591, opened 2022-05-19
fix generator coverage error on CI

blagoev:
This fixes a small error in the generator coverage on CI. We forgot to disable the upload step while we disabled the generator


nielsenko (approved):
https://github.com/CocoaPods/CocoaPods/issues/11355 is still open, but they marked it as resolved in the comments. They are waiting to ensure all regions have picked up the new cloud-flare DDoS heuristics before closing.

So we should be able to revert this (and the previous work-around) soon.

PR #606, opened 2022-05-23
Fix a sporadic session test failure

nirinchev:
Sometimes `session.state` would be `connecting` rather than `connected` when we first check it, leading to a sporadic test failure. This adds some leeway by retrying the check a couple of times.

### DIFF
--- a/test/test.dart
+++ b/test/test.dart
@@ -324,7 +324,7 @@ Future<void> baasTest(
 
 Future<AppConfiguration> getAppConfig({AppNames appName = AppNames.flexible}) async {
   final baasUrl = arguments[argBaasUrl];
-  
+
   final app = baasApps[appName.name] ??
       baasApps.values.firstWhere((element) => element.name == BaasClient.defaultAppName, orElse: () => throw RealmError("No BAAS apps"));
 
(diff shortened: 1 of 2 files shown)


blagoev (approved):

test/test.dart
```diff
@@ -374,6 +374,18 @@ Future<User> loginWithRetry(App app, Credentials credentials, {int retryCount =
   }
 }
 
+Future<void> waitForCondition(bool Function() condition,
+    {Duration timeout = const Duration(seconds: 10), Duration retryDelay = const Duration(milliseconds: 100), String? message}) async {
```
blagoev, line 378: I would rather choose a small default and if more is needed (the rare case) than its required to specify one.

blagoev: Just so the frequent usages with 1 sec duration do not require to specify the timeout everytime

test/test.dart
```diff
@@ -206,16 +207,7 @@ Future<void> setupTests(List<String>? args) async {
...
 
       for (final path in paths) {
-        try {
-          Realm.deleteRealm(path);
-        } catch (e) {
-          print("Can not delete realm at path: $path. Did you forget to close it?");
-        }
-        String pathKey = _path.basenameWithoutExtension(path);
-        String realmDir = _path.dirname(path);
-        await Directory(realmDir).list().forEach((f) {
```
nielsenko, line 216: I prefer `await for` over `await x.forEach((y) {})` https://api.dart.dev/stable/2.16.2/dart-io/Directory-class.html

test/test.dart
```diff
@@ -241,20 +233,28 @@ String generateRandomString(int len) {
...
+  for (var i = 0; i < 100; i++) {
     try {
-      await fileEntity.delete(recursive: recursive);
-      break;
+      Realm.deleteRealm(path);
+      await File('$path.lock').delete();
+      return;
     } catch (e) {
+      print('Failed to delete realm at path $path. Trying again in 50ms');
       await Future<void>.delayed(const Duration(milliseconds: 50));
```
nielsenko, line 253: Maybe complain a bit. Otherwise this will just hide problems I think?

test/test.dart
nirinchev: That's deleted code though?

test/test.dart
nirinchev: How can we complain?

PR #608, opened 2022-05-23
Sync config client reset handler

### LINKED ISSUE #628: Support SyncErrorHandler and SyncClientResetHandler

desistefanova:
Support SyncClientResetHandler 

Fixup SyncErrors support

fixes #628


src/sync_session.cpp
```diff
@@ -132,16 +132,19 @@ RLM_API void realm_dart_sync_session_report_error_for_testing(realm_sync_session
 {
     std::error_code error_code;
     std::string msg;
-    if (category == 0) {
+    std::uint8_t throwError = 0;
```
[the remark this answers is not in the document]
desistefanova: right

common/lib/src/realm_types.dart
```diff
@@ -97,7 +107,7 @@ class SessionError extends SyncError {
 /// {@category Sync}
 class SyncError extends RealmError {
   /// The code of the error
-  final int code; // TODO: this should be an enum
+  final int code; // TODO: this should be an enum. There are two error codes enums in C-API realm::sync::ProtocolError and realm::sync::ClientError
```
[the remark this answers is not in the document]
desistefanova: I don't think

common/lib/src/realm_types.dart
```diff
@@ -79,8 +79,18 @@ class RealmStateError extends StateError implements RealmError {
   RealmStateError(String message) : super(message);
 }
 
+/// An error type that describes a client reset error condition.
+/// {@category Sync}
+class ClientResetError extends SessionError {
+  ClientResetError(
+    String message,
+    SyncErrorCategory category, {
+    bool isFatal = false,
```
nirinchev, line 88: Isn't a client reset error always fatal?

lib/src/configuration.dart
```diff
@@ -49,6 +49,11 @@ typedef InitialDataCallback = void Function(Realm realm);
 /// Client reset errors will not be reported through this callback as they are handled by [ClientResetHandler].
 typedef SessionErrorHandler = void Function(SessionError error);
 
+/// The signature of a callback that will be invoked if a client reset error occurs for this [Realm].
+///
+/// Currently, Flexible sync only supports the Manual Recovery.
```
nirinchev, line 54: Even though we only support manual reset, we should design the API in a way that is extensible. My suggestion is to model it the same way as the .NET API where we have an abstract handler and concrete implementations (dart will only have one to start with, but more will be added later).

common/lib/src/realm_types.dart
desistefanova: How can we check it?

lib/src/configuration.dart
desistefanova: You are right. We can also only rename the handler to ManualyClientResetHandler in order to avoid breaking changes later.

blagoev (approved):

flutter/realm_flutter/tests/test_driver/app_test.dart
```diff
@@ -28,11 +28,11 @@ void main(List<String> args) {
       testCommandWithArgs += getArgFromEnvVariable("BAAS_PRIVATE_API_KEY");
       testCommandWithArgs += getArgFromEnvVariable("BAAS_PROJECT_ID");
 
-      String result = await driver!.requestData(testCommandWithArgs, timeout: const Duration(minutes: 30));
+      String result = await driver!.requestData(testCommandWithArgs, timeout: const Duration(hours: 1));
```
nielsenko, line 31: Seems like a long time..

lib/src/configuration.dart
```diff
@@ -307,3 +304,31 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+  void Function(SyncError code) callback;
+
+  /// Invokes the [SyncError] handling [callback].
+  void call(SyncError error) => callback(error);
+
+  /// Initializes a new instance of of [SyncErrorHandler].
+  SyncErrorHandler(this.callback);
+}
+
+/// The signature of a callback that will be invoked if a client reset error occurs for this [Realm].
```
nielsenko, line 322: Again, why not just a `typedef SyncClientResetErrorHandler = void Function(SyncClientResetError code)`

nielsenko (changes requested):

flutter/realm_flutter/tests/test_driver/app_test.dart
```diff
@@ -28,11 +28,11 @@ void main(List<String> args) {
...
       if (result.isNotEmpty) {
         fail("Failed tests: \n $result");
       }
-    }, timeout: const Timeout(Duration(minutes: 30)));
+    }, timeout: const Timeout(Duration(hours: 1)));
```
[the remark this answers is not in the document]
blagoev: Yes it seems it is. We have much more tests now that take more than 30mins to run

lib/src/configuration.dart
```diff
@@ -64,11 +59,7 @@ abstract class Configuration {
 
   static String? _defaultPath;
 
-  Configuration._(
-    List<SchemaObject> schemaObjects, {
-    String? path,
-    this.fifoFilesFallbackPath
-  }) : schema = RealmSchema(schemaObjects) {
+  Configuration._(List<SchemaObject> schemaObjects, {String? path, this.fifoFilesFallbackPath}) : schema = RealmSchema(schemaObjects) {
```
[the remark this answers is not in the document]
blagoev: I don't understand why the suggested change

lib/src/native/realm_core.dart
```diff
@@ -277,7 +279,7 @@ class _RealmCore {
     return result == nullptr ? null : SubscriptionHandle._(result);
   }
 
-  static void _stateChangeCallback(Pointer<Void> userdata, int state) {
+  static void stateChangeCallback(Pointer<Void> userdata, int state) {
```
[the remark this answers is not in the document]
blagoev: Is this really a problem?

lib/src/configuration.dart
```diff
@@ -307,3 +304,31 @@ class RealmSchema extends Iterable<SchemaObject> {
   @override
   SchemaObject elementAt(int index) => _schema.elementAt(index);
 }
+
+///The signature of a callback that will be invoked whenever a [SyncError] occurs for the synchronized Realm.
```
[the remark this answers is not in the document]
blagoev: This change is to handle feedback from the .NET team about how extensible our handlers are. We are going to have multiple kinds of handlers in the future. 
The comment is from here https://github.com/realm/realm-dart/pull/608#discussion_r882205004

lib/src/configuration.dart
blagoev: see comment above

lib/src/configuration.dart
```diff
@@ -307,3 +304,31 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+///
+/// Client reset errors will not be reported through this callback as they are handled by [ClientResetHandler].
```
desistefanova, line 310:
```suggestion
/// Client reset errors will not be reported through this callback as they are handled by [SyncClientResetErrorHandler].
```

lib/src/native/realm_core.dart
```diff
@@ -1097,7 +1104,7 @@ class _RealmCore {
     return _realmLib.realm_app_get_app_id(app.handle._pointer).cast<Utf8>().toRealmDartString()!;
   }
 
-  static void _app_user_completion_callback(Pointer<Void> userdata, Pointer<realm_user> user, Pointer<realm_app_error> error) {
+  static void app_user_completion_callback(Pointer<Void> userdata, Pointer<realm_user> user, Pointer<realm_app_error> error) {
```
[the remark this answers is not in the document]
blagoev: We should be careful when making statements like this. This is not polluting the PR. This change brings consistency across the callback names. Which I noticed while working on the changes. I don't think it matters much if we use underscore or not and since we had callbacks that don't use it I decided to rename these as well. There should be no real problem here.

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
desistefanova: Yes, why do we need to expose the callbacks? It will make realmCore API too complex and it will be hard to find a method when using it.

flutter/realm_flutter/tests/test_driver/app_test.dart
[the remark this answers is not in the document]
nirinchev: That sounds very surprising. We have a lot more tests in the .NET SDK and the slowest runs we've seen take less than 30 minutes.

common/lib/src/realm_types.dart
```diff
@@ -79,30 +79,140 @@ class RealmStateError extends StateError implements RealmError {
...
+
+  SyncClientError(
     String message,
-    SyncErrorCategory category, {
+    SyncErrorCategory category,
+    SyncClientErrorCode errorCode, {
     this.isFatal = false,
-    int code = 0,
-  }) : super(message, category, code);
+  }) : super(message, category, errorCode.index);
```
desistefanova, line 161:
```
}) : super(message, category, errorCode.code);
```

common/lib/src/realm_types.dart
```diff
@@ -79,30 +79,140 @@ class RealmStateError extends StateError implements RealmError {
...
+
+  /// The [SyncConnectionErrorCode] value indicating the type of the sync error.
+  SyncConnectionErrorCode get code => SyncConnectionErrorCode.values[codeValue];
+
+  SyncConnectionError(
+    String message,
+    SyncErrorCategory category,
+    SyncConnectionErrorCode errorCode, {
+    this.isFatal = false,
+  }) : super(message, category, errorCode.index);
```
desistefanova, line 190:
```suggestion
  }) : super(message, category, errorCode.code);
```

common/lib/src/realm_types.dart
```diff
@@ -79,30 +79,140 @@ class RealmStateError extends StateError implements RealmError {
...
+
+  /// The [SyncSessionErrorCode] value indicating the type of the sync error.
+  SyncSessionErrorCode get code => SyncSessionErrorCode.values[codeValue];
+
+  SyncSessionError(
+    String message,
+    SyncErrorCategory category,
+    SyncSessionErrorCode errorCode, {
+    this.isFatal = false,
+  }) : super(message, category, errorCode.index);
```
desistefanova, line 207:
```suggestion
  }) : super(message, category, errorCode.code);
```

common/lib/src/realm_types.dart
```diff
@@ -79,30 +79,140 @@ class RealmStateError extends StateError implements RealmError {
...
+class SyncClientResetError extends SyncError {
+  /// If true the received error is fatal.
+  final bool isFatal = true;
 
-  /// The category of the error
-  final SyncErrorCategory category;
+  /// The [ClientResetError] has error code of [SyncClientErrorCode.autoClientResetFailure]
+  SyncClientErrorCode get code => SyncClientErrorCode.autoClientResetFailure;
+
+  SyncClientResetError(String message) : super(message, SyncErrorCategory.client, SyncClientErrorCode.autoClientResetFailure.index);
```
desistefanova, line 173:
```suggestion
  SyncClientResetError(String message) : super(message, SyncErrorCategory.client, SyncClientErrorCode.autoClientResetFailure.code);
```

lib/src/configuration.dart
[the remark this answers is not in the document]
nirinchev: The feedback was about the client reset handler. The error handler can be a callback and doesn't need a dedicated class.

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
nirinchev: I agree - even for internal classes, let's not have public members that don't need to be public.

nirinchev (changes requested):

common/lib/src/realm_types.dart
```diff
@@ -79,30 +79,140 @@ class RealmStateError extends StateError implements RealmError {
...
+  /// The category of the sync error
+  final SyncErrorCategory category;
+
+  SyncError(String message, this.category, this.codeValue) : super(message);
+
+  /// Creates a specific type of [SyncError] instance based on the [category] and the [code] supplied.
+  static SyncError create(String message, SyncErrorCategory category, int code, {bool isFatal = false}) {
+    switch (category) {
+      case SyncErrorCategory.client:
+        final errorCode = SyncClientErrorCode.values[code];
```
desistefanova, line 127: values[code] will return "Index out of range" error. We have to find a way to convert codes to enum. Values are working with indexes that are actually the positions in the enum.

flutter/realm_flutter/tests/test_driver/app_test.dart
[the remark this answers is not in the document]
blagoev: well It seems occasionally we reach that limit.

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
blagoev: I will move them all to underscore then. I was trying to make them consistent.

common/lib/src/realm_types.dart
blagoev: good catch. Will fix

common/lib/src/realm_types.dart
```diff
@@ -123,32 +233,266 @@ enum SyncErrorCategory {
...
+  objectAlreadyExists(227),
+
+  /// Server permissions for this file ident have changed since the last time it (used) (IDENT)
+  serverPermissionsChanged(228),
+
+  /// Client tried to open a session before initial sync is complete (BIND)
+  initialSyncNotCompleted(229),
+
+  /// Client attempted a write that is disallowed by permissions, or modifies an object outside the current query - requires client reset (UPLOAD)
+  writeNotAllowed(230);
```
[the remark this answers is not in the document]
blagoev: unsure, just copied the C api values

flutter/realm_flutter/tests/test_driver/app_test.dart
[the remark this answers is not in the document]
nirinchev: I imagine that may be related to the windows slowness that we haven't figured out yet. Once we fix that, we should hopefully see tests complete a lot faster (I believe on Linux they complete in a few minutes).

lib/src/configuration.dart
```diff
@@ -307,3 +304,31 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+
+/// The signature of a callback that will be invoked if a client reset error occurs for this [Realm].
+///
+/// Currently, Flexible sync only supports the Manual Recovery.
+class SyncClientResetErrorHandler {
+  /// The callback that handles the [SyncClientResetError].
+  void Function(SyncClientResetError code) callback;
+
+  /// Invokes the [SyncClientResetError] handling [callback].
+  void call(SyncClientResetError error) => callback(error);
```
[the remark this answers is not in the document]
blagoev: This is a callable class it should be public

common/lib/src/realm_types.dart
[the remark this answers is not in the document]
nirinchev: Yeah, those are asking for a review, but let's not do it here - it's just annoying that we'll add something enum members only to immediately deprecate them.

lib/src/configuration.dart
```diff
@@ -222,16 +215,20 @@ class FlexibleSyncConfiguration extends Configuration {
...
   SessionStopPolicy _sessionStopPolicy = SessionStopPolicy.afterChangesUploaded;
-  
-  /// Called when a [SessionError] occurs for the synchronized Realm.
-  final SessionErrorHandler? sessionErrorHandler;
+
+  /// Called when a [SyncError] occurs for this synchronized [Realm].
+  final SyncErrorHandler? syncErrorHandler;
+
+  /// Called when a [SyncClientResetError] occurs for this synchronized [Realm]
+  final SyncClientResetErrorHandler? syncClientResetErrorHandler;
```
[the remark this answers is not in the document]
blagoev: This is possible only if we move it to required in the flexible sync config? Are you suggesting that?

lib/src/configuration.dart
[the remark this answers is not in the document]
nirinchev: What does a callable class mean? It's not supposed to be called by users, only by us.

lib/src/configuration.dart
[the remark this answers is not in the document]
blagoev: we have manual reset only currently no? how do you suggest to supply a default one?

lib/src/configuration.dart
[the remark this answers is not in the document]
blagoev: A callable class in Dart is a class you can call as a function

lib/src/configuration.dart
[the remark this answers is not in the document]
nirinchev: We can construct one that just logs and tells people to do something themselves:

```dart
final defaultHandler = ManualResetHandler((_) {
  Realm.logger.log(LogLevel.Error, "A client reset error occurred but no handler was supplied. "
    "Synchronization is now paused and will resume automatically once the app is restarted and "
    "the server data is redownloaded. Any unsynchronized changes the client has made or will "
    "make will be lost. To handle that scenario, pass in a non-null value to 
    "`syncClientResetErrorHandler` when constructing `Configuration.flexibleSync`");
});
```

It doesn't do anything to handle the error but instructs the user what they need to do since the Core error message is going to be overly generic and not geared toward the Dart SDK API.

lib/src/configuration.dart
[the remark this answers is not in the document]
nirinchev: I see - then this should probably not be callable class. The manual reset handler does indeed have a single method on it, but the discard local handler has 3 and the recovery handler will have at least 3.

lib/src/configuration.dart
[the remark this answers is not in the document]
blagoev: I think this is not ideal. Why not make it to require. I am for moving it as a required field right now later when we have more types we can relax the constraint.  
Also this kind of message should be an exception. This is really hiding the issue until later when the person might have shipped the app. So we better be clear we require this field. ClientResets can happen regularly. 
Also our SyncErrorHandler should be a required field. Thus we have a chance to really raise any sync issue to the developer. They can choose to handle it as printing or something else.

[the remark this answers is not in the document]
blagoev: since there is no time for better I will implement what you suggested.

lib/src/configuration.dart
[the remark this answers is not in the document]
blagoev: removed

lib/src/configuration.dart
```diff
@@ -211,6 +208,23 @@ enum SessionStopPolicy {
...
+
+void defaultSyncErrorHandler(SyncError e) {
+   Realm.logger.log(RealmLogLevel.error, e);
+}
+
+void _defaultSyncClientResetHandler(SyncError e) {
+  Realm.logger.log(RealmLogLevel.error, "A client reset error occurred but no handler was supplied. "
+    "Synchronization is now paused and will resume automatically once the app is restarted and "
+    "the server data is redownloaded. Any unsynchronized changes the client has made or will "
+    "make will be lost. To handle that scenario, pass in a non-null value to"
```
nirinchev, line 224:
```suggestion
    "make will be lost. To handle that scenario, pass in a non-null value to " 
```

nielsenko (approved):

lib/src/configuration.dart
```diff
@@ -295,3 +317,16 @@ class RealmSchema extends Iterable<SchemaObject> {
   @override
   SchemaObject elementAt(int index) => _schema.elementAt(index);
 }
+
+/// The signature of a callback that will be invoked if a client reset error occurs for this [Realm].
+///
+/// Currently, Flexible sync only supports the Manual Recovery.
+class SyncClientResetErrorHandler {
```
nirinchev, line 324: Can we make that class abstract and its constructor private? That way people will be forced to use `ManualSyncClientResetHandler` instead. The reasoning here is to have a clear distinction between the interface and the concrete implementors so that when we add more options, people won't default to the wrong one.

lib/src/configuration.dart
```diff
@@ -295,3 +317,16 @@ class RealmSchema extends Iterable<SchemaObject> {
   @override
   SchemaObject elementAt(int index) => _schema.elementAt(index);
 }
+
+/// The signature of a callback that will be invoked if a client reset error occurs for this [Realm].
+///
+/// Currently, Flexible sync only supports the Manual Recovery.
```
nirinchev, line 323: The docs should probably mention `[ManualSyncClientResetHandler]` so that it's explicit what is the type users are supposed to use.

lib/src/configuration.dart
```diff
@@ -295,3 +317,16 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+class SyncClientResetErrorHandler {
+  /// The callback that handles the [SyncClientResetError].
+  final void Function(SyncClientResetError code) callback;
+
+  /// Initializes a new instance of of [SyncClientResetErrorHandler].
+  const SyncClientResetErrorHandler(this.callback);
+}
+
+typedef ManualSyncClientResetHandler = SyncClientResetErrorHandler;
```
nirinchev, line 332: This is missing docs.

lib/src/native/realm_core.dart
```diff
@@ -404,15 +406,19 @@ class _RealmCore {
...
     }
-    final sessionError = error.toSessionError();
-    if(syncConfig.sessionErrorHandler != null) {
-        syncConfig.sessionErrorHandler!(sessionError);
-    }
+
+    final syncError = error.toSyncError();
+
+    if (syncError is SyncClientResetError) {
+        syncConfig.syncClientResetErrorHandler!.callback(syncError);
```
nirinchev: Probably doesn't need the `!` anymore.

nirinchev (commented):
Some final suggestions for restructuring the type hierarchy, otherwise looks good.

lib/src/native/realm_core.dart
blagoev: fixed

lib/src/configuration.dart
blagoev: It's not a wrong per se. If someone wants to use this implementation or implement or extends it they can do it. No need to be too restrictive here imo

lib/src/configuration.dart
blagoev: fixed

test/session_test.dart
```diff
@@ -279,15 +279,50 @@ Future<void> main([List<String>? args]) async {
...
+      final syncClientResetError = syncError.as<SyncClientResetError>();
+      expect(syncClientResetError.category, SyncErrorCategory.session);
+      expect(syncClientResetError.isFatal, true);
+      expect(syncClientResetError.code, SyncClientErrorCode.autoClientResetFailure);
+      expect(syncClientResetError.message, "Simulated session error");
     });
     final realm = getRealm(config);
 
-    realm.syncSession.raiseSessionError(SyncErrorCategory.session, 100, false);
+    realm.syncSession.raiseError(SyncErrorCategory.session, SyncClientErrorCode.autoClientResetFailure.index, true);
```
desistefanova, line 325:
```suggestion
    realm.syncSession.raiseError(SyncErrorCategory.session, SyncClientErrorCode.autoClientResetFailure.code, true);
```

test/session_test.dart
```diff
@@ -279,15 +279,50 @@ Future<void> main([List<String>? args]) async {
...
+      expect(syncError, isA<SyncClientError>());
+      final syncClientError = syncError.as<SyncClientError>();
+      expect(syncClientError.category, SyncErrorCategory.client);
+      expect(syncClientError.isFatal, true);
+      expect(syncClientError.code, SyncClientErrorCode.badChangeset);
+      expect(syncClientError.message, "Simulated client error");
+    });
+    final realm = getRealm(config);
+
+    realm.syncSession.raiseError(SyncErrorCategory.client, SyncClientErrorCode.badChangeset.index, true);
```
desistefanova, line 309:
```suggestion
    realm.syncSession.raiseError(SyncErrorCategory.client, SyncClientErrorCode.badChangeset.code, true);
```

test/session_test.dart
```diff
@@ -279,15 +279,50 @@ Future<void> main([List<String>? args]) async {
...
       expect(sessionError.isFatal, false);
-      expect(sessionError.code, 100);
-      expect(sessionError.message, "Simulated sync session error");
+      expect(sessionError.code, SyncSessionErrorCode.badAuthentication);
+      expect(sessionError.message, "Simulated session error");
+    });
+
+    final realm = getRealm(config);
+
+    realm.syncSession.raiseError(SyncErrorCategory.session, SyncSessionErrorCode.badAuthentication.index, false);
```
desistefanova, line 293:
```suggestion
    realm.syncSession.raiseError(SyncErrorCategory.session, SyncSessionErrorCode.badAuthentication.code, false);
```

test/session_test.dart
```diff
@@ -279,15 +279,50 @@ Future<void> main([List<String>? args]) async {
...
     final user = await getIntegrationUser(app);
-    final config = Configuration.flexibleSync(user, [Task.schema], sessionErrorHandler: (sessionError) {
+    final config = Configuration.flexibleSync(user, [Task.schema], syncErrorHandler: (syncError) {
+      expect(syncError, isA<SyncSessionError>());
+      final sessionError = syncError.as<SyncSessionError>();
       expect(sessionError.category, SyncErrorCategory.session);
       expect(sessionError.isFatal, false);
-      expect(sessionError.code, 100);
-      expect(sessionError.message, "Simulated sync session error");
+      expect(sessionError.code, SyncSessionErrorCode.badAuthentication);
```
desistefanova, line 287:
```suggestion
      expect(sessionError.code, SyncSessionErrorCode.badAuthentication.code);
```

common/lib/src/realm_types.dart
```diff
@@ -79,30 +79,170 @@ class RealmStateError extends StateError implements RealmError {
...
+}
+
+/// An error type that describes a session-level error condition.
+/// {@category Sync}
+class SyncSessionError extends SyncError {
+  /// If true the received error is fatal.
+  final bool isFatal;
+
+  /// The [SyncSessionErrorCode] value indicating the type of the sync error.
+  SyncSessionErrorCode get code => SyncSessionErrorCode.values[codeValue];
```
desistefanova, line 220:
```suggestion
  SyncSessionErrorCode get code => SyncSessionErrorCode.fromInt(codeValue);
```

common/lib/src/realm_types.dart
```diff
@@ -79,30 +79,170 @@ class RealmStateError extends StateError implements RealmError {
...
-/// /// {@category Sync}
-class SessionError extends SyncError {
+/// {@category Sync}
+class SyncClientError extends SyncError {
   /// If true the received error is fatal.
   final bool isFatal;
 
-  SessionError(
+  /// The [SyncClientErrorCode] value indicating the type of the sync error.
+  SyncClientErrorCode get code => SyncClientErrorCode.values[codeValue];
```
desistefanova, line 159:
```suggestion
  SyncClientErrorCode get code => SyncClientErrorCode.fromInf(codeValue);
```

common/lib/src/realm_types.dart
```diff
@@ -79,30 +79,170 @@ class RealmStateError extends StateError implements RealmError {
...
-class SyncError extends RealmError {
-  /// The code of the error
-  final int code; // TODO: this should be an enum
+class SyncClientResetError extends SyncError {
+  /// If true the received error is fatal.
+  final bool isFatal = true;
 
-  /// The category of the error
-  final SyncErrorCategory category;
+  /// The [ClientResetError] has error code of [SyncClientErrorCode.autoClientResetFailure]
```
desistefanova, line 180:
```suggestion
  /// The [SyncClientResetError] has error code of [SyncClientErrorCode.autoClientResetFailure]
```

common/lib/src/realm_types.dart
```diff
@@ -79,30 +79,170 @@ class RealmStateError extends StateError implements RealmError {
...
 
-/// Thrown or reporeted if an error occurs during synchronization
+/// An error type that describes a client reset error condition.
 /// {@category Sync}
-class SyncError extends RealmError {
-  /// The code of the error
-  final int code; // TODO: this should be an enum
+class SyncClientResetError extends SyncError {
+  /// If true the received error is fatal.
+  final bool isFatal = true;
```
desistefanova, line 178: If it is always fatal error, do we need this member isFatal at all?

lib/src/configuration.dart
nirinchev: We should not allow extending this class. The reason is because we're going to use the inheritor type to determine the resync mode. Something like:

```dart
ResyncMode getMode() {
  if (clientResetHandler is ManualResetHandler) {
    return ResyncMode.Manual;	  
  }

  // these will be added soon-ish
  if (clientResetHandler is DiscardLocalHandler) {
    return ResyncMode.DiscardLocal;
  }

  if (clientResetHandler is RecoverLocalHandler) {
    return ResyncMode.RecoverLocal;
  }

  throw RealmException('unsupported client reset handler');
}
```

blagoev: We can not stop people from implementing a class in Dart. 

I don't see are reason for these restriction. We could still do what you say and still allow custom implementations. In such cases we will call the callback of their implementation only. In this case it will be a `CustomClientResetHandler` implementation. 
Actually in the case above I would rather have a getter that returns the type of `ResyncMode` on the  ClientResetClass. Thus we don't throw `unsupported errors` but can work with any implementation of ClientResetClass.

I think we are talking to much in anticipation of the future. Which can go only so far. Lets not put the carriage before the horse.
I will resolve this and we can handle it later when the times comes.

PR #613, opened 2022-05-25
Update realm-core to v12.0.0

nielsenko:


blagoev:
Do we need something from v12. If not I suggest we don't update until there is a proper Core release.

PR #614, opened 2022-05-25
Upgrade to ffigen 5 and fix header paths (again)

fealebenpae:
This tries to resolve the glob issue introduced by #588. Turns out that the [glob implementation](https://github.com/google/quiver-dart/blob/524f33ddaa416af6e616a2f86edac44a5647ceed/lib/pattern.dart#L93) used by ffigen doesn't take into account different directory separators, unlike [glob](https://github.com/isaacs/node-glob#windows) on Node.js. In order to work around this I had to not use directory separator characters in the glob string, so `**/src/*.h` became `**src**.h`. The trouble with this is that it matched the `dart-dl` headers.

The problems are compounded by ffigen 5 - it seems to consider `realm.h` multiple times - once for each entry point header that includes it, like `realm_dart.h` and `subscription_set.h` and others. This causes it to produce several copies of each `realm.h` symbol binding, unlike ffigen 4. The workaround I came up with to combat this is `entrypoint.h` - ensuring that ffigen only processes a single translation unit altogether works around the binding duplication issue. This regrettably means that we still need to make sure `entrypoint.h` contains the explicit list of headers we want ffigen to process. Perhaps we can open an issue against ffigen about the symbol duplication, and if we move `dart-dl` out from the `src` folder we'd remove the need for `entrypoint.h`.

In all fairness I didn't have to upgrade to ffigen 5 now, but seeing as the problems it caused are specifically related to how we discover headers I think it's best to solve this once for the foreseeable future. Side-effects include its support for ABI-specific integers in Dart 2.17, with types like `ffi.Char` and `ffi.Size`. I modified `realm_core.dart` to use the new types and changed the way we pass strings and integer arrays to native, because our old extensions were based around types ffigen doesn't generate anymore.


ffigen/entrypoint.h
```diff
@@ -0,0 +1,22 @@
...
+typedef struct _Dart_WeakPersistentHandle* Dart_WeakPersistentHandle;
+typedef struct _Dart_FinalizableHandle* Dart_FinalizableHandle;
+typedef int64_t Dart_Port;
+
+// The C API header
+#include <realm.h>
+
+// List all the headers that should be processed by ffigen, relative to the src folder
+#include <realm_dart_scheduler.h>
+#include <realm_dart.h>
```
blagoev, line 19: What good is this for if we do the same as the `config.yaml`

blagoev: I suggest we continue with the `config.yaml` much simpler than fiddling with a workaround headers.

blagoev (changes requested):

ffigen/entrypoint.h
blagoev: we have them regardless in other places. I suggest we continue using it.

blagoev:
the ffigen 5 duplicating seems like a bug in ffigen

ffigen/entrypoint.h
fealebenpae: We should stop using symlinks in other places as well 😛
But please go over the PR again when you have more time - the single entry point header is necessary both to get rid of symlinks, but to also work around a breaking change in ffigen 5 when it comes to encountering the same declaration multiple times.

fealebenpae:
> the ffigen 5 duplicating seems like a bug in ffigen

Yeah, I'm inclined to agree - I just haven't had the time to research whether it's actual intended change in behavior they need for their Objective-C support, or an unintended side-effect of something else.

blagoev:
we moved the discussion offline. Will be parking this PR for when ffigen 5 bug is fixed.

PR #617, opened 2022-05-29
fixup shutdown api,

blagoev:
fixup shutdown api,
add changelog,
bring back scheduler arguments

### DIFF
--- a/lib/src/app.dart
+++ b/lib/src/app.dart
@@ -19,6 +19,7 @@ import 'dart:io';
 
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
+import 'package:realm_dart/src/scheduler.dart';
 
 import '../realm.dart';
 import 'configuration.dart';
@@ -117,7 +118,7 @@ class App {
   String get id => realmCore.appGetId(this);
 
   /// Create an app with a particular [AppConfiguration]
-  App(AppConfiguration configuration) : this._(realmCore.getApp(configuration));
+  App(AppConfiguration configuration) : this._(realmCore.getApp(configuration, scheduler.handle));
 
   App._(this._handle);
 
--- a/lib/src/realm_object.dart
+++ b/lib/src/realm_object.dart
@@ -18,6 +18,8 @@
 
 import 'dart:async';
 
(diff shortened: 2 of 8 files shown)


lib/src/app.dart
```diff
@@ -117,7 +118,7 @@ class App {
   String get id => realmCore.appGetId(this);
 
   /// Create an app with a particular [AppConfiguration]
-  App(AppConfiguration configuration) : this._(realmCore.getApp(configuration));
+  App(AppConfiguration configuration) : this._(realmCore.getApp(configuration, scheduler.handle));
```
nielsenko, line 121: First, why do you pass a handle here. We previously agreed that we should not pass handles to `_RealmCore`. Secondly, why pass a handle to a singleton?

nielsenko (changes requested):

lib/src/app.dart
blagoev: The reason is realm_core should not know about the rest of the SDK and should use an explicit interface. This is how it designed from the start. Thus methods in realm_core are an extension to `realm.h` and not part of the SDK. Then we can redesign the SDK without touching the realm_core interface and we can replace the realm_core implementation in the future while keeping the interface intact. 

The reason to pass handle here is since I did not want to include scheduler.dart in realm_core which gives very easy access to the singleton value and someone might use that in the future.

nielsenko: `_RealmCore` is internal. It already knows how to read configs, handles, etc. We are not bound by its interface at all. I don't buy this argument.

blagoev: we have discussed this offline and I will be fixing it reverting these changes

lib/src/realm_object.dart
```diff
@@ -21,6 +21,7 @@ import 'dart:async';
 import 'list.dart';
 import 'native/realm_core.dart';
 import 'realm_class.dart';
+import 'scheduler.dart';
```
nielsenko, line 24: The scheduler imports should not be needed

nielsenko (changes requested):

PR #621, opened 2022-05-30
Support opening a synchronized realm without synchronizing

nielsenko:
Use the new `SyncFollowerConfiguration` to open a synchronized realm,
without actually starting a sync session.


lib/src/configuration.dart
```diff
@@ -241,6 +242,14 @@ extension FlexibleSyncConfigurationInternal on FlexibleSyncConfiguration {
   set sessionStopPolicy(SessionStopPolicy value) => _sessionStopPolicy = value;
 }
 
+class SyncFollowerConfiguration extends Configuration {
```
desistefanova, line 245: We can come up with a more meaningful name for SyncFollowerConfiguration. But I have no ideas.

lib/src/configuration.dart
```diff
@@ -241,6 +242,14 @@ extension FlexibleSyncConfigurationInternal on FlexibleSyncConfiguration {
...
+  SyncFollowerConfiguration._(
+    super.schemaObjects, {
+    super.fifoFilesFallbackPath,
+    super.path,
```
desistefanova, line 249: Isn't it dangerous to change the paths after the realm already exists and it is synced?

desistefanova: Do you have some test sample that display this scenario with opening local realm?

desistefanova (commented):
I need to see some test that demonstrates the use case of opening a realm with SyncFollowerConfiguration.

lib/src/configuration.dart
nielsenko: The path needs to match the one used, when opening the same realm with a `FlexibleSyncConfiguration`

lib/src/configuration.dart
nielsenko: Yes me too.. I was hoping you could help me out 😏

lib/src/configuration.dart
nielsenko: I will add a test.

lib/src/configuration.dart
nielsenko: Any of these you like:
- `SecondarySyncConfiguration`,
- `SecondaryProcessConfiguration`,
- `SyncSlaveConfiguration`,
- `SyncSecondaryConfiguration`,
- `SyncReplicaConfiguration`,
- `ReplicaConfiguration`,
- `SyncStandbyConfiguration`,
-  `StandbyConfiguration`

nirinchev: `SessionlessSyncConfiguration`? 😄

nielsenko:
It is tricky to write a test that can run in a single process for this. I have a sample app that is running fine with this, and the code is really simple. I will set ready-for-review.

lib/src/configuration.dart
nielsenko: No I won't .. it will require some refactoring of the test harnish

nirinchev:
Another test you can write is to create a Realm with a proper sync config, write some data, then close it. Then create a `SessionlessSyncConfig` pointing to the same file and open the Realm with that. It should succeed and you should be able to read the data. If you try to do it with a regular `RealmConfig`, that should fail.

test/configuration_test.dart
```diff
@@ -20,6 +20,7 @@
 import 'dart:io';
 import 'dart:math';
 import 'package:test/test.dart' hide test, throws;
+import 'package:path/path.dart' as path_;
```
blagoev, line 23:
```suggestion
import 'package:path/path.dart' as path;
```

test/configuration_test.dart
```diff
@@ -414,4 +415,26 @@ Future<void> main([List<String>? args]) async {
 
     expect(config.path, 'my-custom-path.realm');
   });
+
+  baasTest('Configuration.sessionlessSync', (appConfig) async {
+    final app = App(appConfig);
+    final user = await app.logIn(Credentials.emailPassword(testUsername, testPassword));
+
+    final dir = await Directory.systemTemp.createTemp();
+    final path = path_.join(dir.path, 'test.realm');
```
blagoev, line 424:
```suggestion
    final testRealmPath = path_.join(dir.path, 'test.realm');
```

lib/src/configuration.dart
```diff
@@ -147,6 +147,18 @@ abstract class Configuration {
         syncErrorHandler: syncErrorHandler,
         syncClientResetErrorHandler: syncClientResetErrorHandler,
       );
+
+  /// Constructs a [SessionlessSyncConfiguration]
+  static SessionlessSyncConfiguration sessionlessSync(
```
blagoev, line 152: Do we need another ctor for this. Why not expose a flag on the `local` ctor instead. This is really is useful in other scenarios as well. I believe this type of open should be exposed on the local ctor so we support all local open realm settings. 
I will add a discussion point about this since it seems we need to discuss other items related to this as well.

blagoev (changes requested):

blagoev:
The change is needed, but lets discuss how to expose it.

lib/src/configuration.dart
nielsenko: I already had a discussion with @nirinchev about this. Since we are opening a realm with history, it is really closer in nature to a flexible sync realm than a local realm. There are a number of features that are not exposed on synchronizable realms, such as 

- `initialDataCallback`,
- `schemaVersion`,
- `disableFormatUpgrade`, and
- `shouldCompactCallback`

That would be odd, if not dangerous to expose, if the realm is opened simultaneously in another process with a flexible  sync configuration. It might make sense to expose `isReadOnly`, though.

lib/src/configuration.dart
```diff
@@ -263,6 +278,20 @@ extension FlexibleSyncConfigurationInternal on FlexibleSyncConfiguration {
   set sessionStopPolicy(SessionStopPolicy value) => _sessionStopPolicy = value;
 }
 
+/// [DisconnectedSyncConfiguration] is used to open [Realm] instances that are synchronized
+/// with MongoDB Atlas, without establishing a connection to Atlas App Services. This allows
+/// for the synchronized realm to be opened in multiple processes concurrently, as long as 
+/// at most one of them uses a a [FlexibleSyncConfiguration].
```
blagoev, line 284:
```suggestion
/// only one of them uses a [FlexibleSyncConfiguration] to sync changes.
```

lib/src/realm_class.dart
```diff
@@ -324,6 +324,8 @@ class Realm {
   /// Disclaimer: This method is mostly needed on Dart standalone and if not called the Dart probram will hang and not exit. 
   /// This is a workaround of a Dart VM bug and will be removed in a future version of the SDK.
   static void shutdown() => scheduler.stop();
+
+  static void startup() => realmCore.clearCachedApps(); // Workaround
```
blagoev, line 328: Why do we need this?

blagoev (changes requested):

lib/src/configuration.dart
blagoev: we settled on `disconnectedSync` constructor and type

blagoev (approved):

PR #623, opened 2022-05-30
Add support for remaining session API

nirinchev:
This adds support for connection state change notifications on the sync session and extends the session user test.

Fixes https://github.com/realm/realm-dart/issues/596
Fixes https://github.com/realm/realm-dart/issues/595

### DIFF
--- a/lib/src/native/realm_core.dart
+++ b/lib/src/native/realm_core.dart
@@ -410,8 +410,8 @@ class _RealmCore {
       return;
     }
     final sessionError = error.toSessionError();
-    if(syncConfig.sessionErrorHandler != null) {
-        syncConfig.sessionErrorHandler!(sessionError);
+    if (syncConfig.sessionErrorHandler != null) {
+      syncConfig.sessionErrorHandler!(sessionError);
     }
   }
 
@@ -1066,7 +1066,7 @@ class _RealmCore {
 
       _realmLib.realm_sync_client_config_set_base_file_path(handle._pointer, configuration.baseFilePath.path.toUtf8Ptr(arena));
       _realmLib.realm_sync_client_config_set_metadata_mode(handle._pointer, configuration.metadataPersistenceMode.index);
-      
+
       _realmLib.realm_sync_client_config_set_log_level(handle._pointer, _LogLevel.fromLevel(Realm.logger.level).index);
       _realmLib.realm_dart_sync_client_config_set_log_callback(
         handle._pointer,
@@ -1075,7 +1075,7 @@ class _RealmCore {
         nullptr,
         scheduler.handle._pointer,
       );
-      
+
       _realmLib.realm_sync_client_config_set_connect_timeout(handle._pointer, configuration.maxConnectionTimeout.inMicroseconds);
       if (configuration.metadataEncryptionKey != null && configuration.metadataPersistenceMode == MetadataPersistenceMode.encrypted) {
         _realmLib.realm_sync_client_config_set_metadata_encryption_key(handle._pointer, configuration.metadataEncryptionKey!.toUint8Ptr(arena));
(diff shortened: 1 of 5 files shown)


lib/src/session.dart
```diff
@@ -68,6 +68,13 @@ class Session {
     return controller.createStream();
   }
 
+  /// Gets a [Stream] of [ConnectionState] that can be used to be notified whenever the
+  /// connection state changes.
+  Stream<ConnectionStateChange> getConnectionStateStream() {
```
nielsenko, line 73: I suggest this is just a getter property called `connectionStateChanges`. We don't pass any parameters here, so it doesn't need to be a method.

lib/src/session.dart
```diff
@@ -184,3 +202,38 @@ class SessionProgressNotificationsController {
     _token = null;
   }
 }
+
+/// @nodoc
+class SessionConnectionStateController {
+  final Session _session;
+  late final StreamController<ConnectionStateChange> _streamController;
+  int? _token;
```
nielsenko, line 210: My usual reservations.

lib/src/native/realm_core.dart
```diff
@@ -1502,6 +1502,28 @@ class _RealmCore {
     _realmLib.realm_sync_session_unregister_progress_notifier(session.handle._pointer, token);
   }
 
+  int sessionRegisterConnectionStateNotifier(Session session, SessionConnectionStateController controller) {
+    return _realmLib.realm_dart_sync_session_register_connection_state_change_callback(
+        session.handle._pointer,
+        Pointer.fromFunction(on_connection_state_change),
+        controller.toPersistentHandle(),
```
nielsenko, line 1509: My usual reservations

test/session_test.dart
```diff
@@ -289,6 +291,39 @@ Future<void> main([List<String>? args]) async {
...
     realm.syncSession.raiseSessionError(SyncErrorCategory.session, 100, false);
   });
+
+  baasTest('SyncSession.getConnectionStateStream', (configuration) async {
+    final realm = await getIntegrationRealm();
+
+    await waitForCondition(() => realm.syncSession.connectionState == ConnectionState.connected);
+
+    final states = <ConnectionStateChange>[];
+    final stream = realm.syncSession.getConnectionStateStream();
```
nielsenko, line 301: There are specific support for testing streams, such as `emitInOrder` etc. that can make these tests more elegant.

nielsenko (changes requested):
I would like `getConnectionStateStream()` to become a getter, and renamed to `connectionStateChanges`. The other comments are suggestions.

lib/src/session.dart
nirinchev: Yeah, I looked at 2.17 finalizers and they don't seem to work the way I was hoping they will. I think we can still make it work by using weak references but I feel it's a bit of a risky change that I won't have the time to properly test and think through.

test/session_test.dart
nirinchev: I'm a noob when it comes to streams and tests, so couldn't figure out how to do it - my main issue is that I need to do something to trigger emitting the events, which I don't see an obvious way to do.

lib/src/session.dart
blagoev: This is related to what @nielsenko is working on. Should we drop both tasks for until after the release. Or should we fix both with the same fix

blagoev (approved):
minor comments

lib/src/session.dart
blagoev: +1

lib/src/native/realm_core.dart
blagoev: Yes this should be a weakhandle. Since we not doing that for the other callback lets have a //TODO here to fix it

lib/src/session.dart
nirinchev: I think we should merge this one as is since it's not inherently wrong - if the user cancels the subscription, the handle will be cleaned up and similarly to the progress notifiers, these are not expensive subscriptions to keep around.

When @nielsenko has a fix for the progress notifier, he can apply the same fix for the connection state notifiers, but that's not critical for the release.

blagoev: its ok by me

nirinchev:
The test failures all seem to be related to https://github.com/realm/realm-dart/pull/639, so I'll merge this as is and hope failures will be fixed by the http trampoline PR.

PR #624, opened 2022-05-30
build all platforms in a matrix

### LINKED ISSUE #560: (Parallelize) building for all architectures on CI

fealebenpae:
Fixes #560.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -190,6 +190,7 @@ jobs:
       - name: Build .xcframework
         run: |
           rm -rf ./binary/ios/realm_dart.xcframework
+          find .
           xcodebuild -create-xcframework \
             -framework ./binary/ios/Release-iphoneos/realm_dart.framework \
             -framework ./binary/ios/Release-iphonesimulator/realm_dart.framework \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,7 +189,6 @@ jobs:
 
       - name: Build .xcframework
         run: |
-          rm -rf ./binary/ios/realm_dart.xcframework
           xcodebuild -create-xcframework \
             -framework ./binary/ios/Release-iphoneos/realm_dart.framework \
             -framework ./binary/ios/Release-iphonesimulator/realm_dart.framework \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,11 +148,12 @@ jobs:
           cache-key: ${{ matrix.preset }}
           cache-restore-keys: ${{ matrix.preset }}
 
+      - name: Setup Android NDK
+        if: startsWith(matrix.preset, 'android-')
+        run: echo "ANDROID_NDK=$ANDROID_NDK_LATEST_HOME" >> $GITHUB_ENV
+
       - name: Build
         if: ${{ contains(github.head_ref, 'release/') || steps.check-cache.outputs.cache-hit != 'true' }}
-        env:
-          # needed by the android preset
-          ANDROID_NDK: ${{ env.ANDROID_NDK_LATEST_HOME }}
         run: |
           cmake --preset ${{ matrix.preset }}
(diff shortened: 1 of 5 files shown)


nielsenko (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -117,13 +103,25 @@ jobs:
...
-    runs-on: ubuntu-latest
-    name: Build Linux
+  build-native:
+    runs-on: ${{ matrix.runner }}-latest
+    name: Build native
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - { preset: windows-x64, runner: windows }
```
nirinchev, line 115: This will produce an artifact called `librealm-windows-x64`. This is problematic because the release workflow expect the artifact names to match the target platform name (i.e. we need to upload the binary as `windows`, not `windows-x64`).

nirinchev (changes requested):
With these changes, we'll now create an extra folder in the artifact. This means that we need to update the release workflow to take that into account - currently it expects the artifact to contain the binary or the architecture/binary (e.g. `windows.zip/realm_dart.dll` rather than `windows-x64.zip/windows/realm_dart.dll`.

The `Archive binaries` step determines the name of the archive, so that will need to ignore the name of the top folder (which matches the artifact name) and instead look at the child folder. It'll also need to be updated to put all the android architectures in the same android folder rather than produce 4 different archives.

.github/workflows/ci.yml
```diff
@@ -135,32 +133,95 @@ jobs:
...
+          key: binaries-${{ matrix.build || matrix.preset }}-${{hashFiles('./src/**')}}
 
       - name: Setup Ninja
-        if: ${{ contains(github.head_ref, 'release/') || steps.check-cache.outputs.cache-hit != 'true' }}
+        if: contains(github.head_ref, 'release/') || steps.check-cache.outputs.cache-hit != 'true'
         uses: seanmiddleditch/gha-setup-ninja@master
 
-      - name: Build Realm for Linux
+      - name: Setup Vcpkg
+        if: contains(github.head_ref, 'release/') || steps.check-cache.outputs.cache-hit != 'true'
```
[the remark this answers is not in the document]
fealebenpae: Same as above. Plus eventually it'll be nice to use vcpkg for OpenSSL on Linux and Android too.

src/CMakeLists.txt
```diff
@@ -60,8 +60,6 @@ if(ANDROID)
     # Core requires these options for the final *.so file. See more realm-core/CMakeLists.txt
     target_link_options(realm_dart PRIVATE LINKER:-gc-sections)
 
-    set_target_properties(realm_dart PROPERTIES LIBRARY_OUTPUT_DIRECTORY "${PROJECT_SOURCE_DIR}/binary/android/${ANDROID_ABI}")
```
blagoev, line 63: are we sure we are outputting the files to the correct locations

blagoev (approved):

src/CMakeLists.txt
fealebenpae: This is already being taken care of by the logic at the end of the file and it was redundant to also have it here.

.github/workflows/ci.yml
```diff
@@ -534,49 +516,14 @@ jobs:
...
-          ./scripts/build-android.sh
-
-      - name: Store artifacts
-        uses: actions/upload-artifact@v2
-        with:
-          name: librealm-android
-          path: binary/android/**
-          retention-days: 1
-
   flutter-android:
```
desistefanova, line 519: Just an idea: I think, in order to be sure that the build is correct for all the architectures we will have to `Create Android Emulator` and `Run tests on Android Emulator` for all the architectures. It also could be done in a matrix.
The same for IOS. We could do it for the master only.

desistefanova (approved):
We have to think also how to clear the baas cluster as soon as the tests for a specific platform is passed. Currently we holds 8 clusters for almost 30 min and the other CI workflow runs may fail.

PR #625, opened 2022-05-30
report error on setup tests exceptions

blagoev:


flutter/realm_flutter/tests/test_driver/realm_test.dart
```diff
@@ -19,32 +19,37 @@ import '../test/session_test.dart' as session_test;
...
+
+    tearDownAll(() {
+      if (failedTests.isNotEmpty) {
+        completer.complete(failedTests.join('\n'));
+      } else {
+        completer.complete('');
+      }
+    });
+  } catch (e) {
+    completer.complete(e.toString());
```
nielsenko, line 51: Should it be:
```suggestion
    completer.completeError(e.toString());
```

blagoev: no we don't need to do that, we need the string result

nielsenko (approved):

PR #627, opened 2022-05-31
don't delete realms on CI

blagoev:


nielsenko (approved):
We should try to get to the bottom of this later

PR #632, opened 2022-05-31
Use new callback trampolines

fealebenpae:


lib/src/native/realm_core.dart
```diff
@@ -279,24 +286,18 @@ class _RealmCore {
     return result == nullptr ? null : SubscriptionHandle._(result);
   }
 
-  static void _stateChangeCallback(Pointer<Void> userdata, int state) {
-    final completer = userdata.toObject<Completer<SubscriptionSetState>>(isPersistent: true);
-    if (completer == null) {
-      return;
-    }
+  static void _stateChangeCallback(Object userdata, int state) {
+    final completer = userdata as Completer<SubscriptionSetState>;
```
nielsenko, line 290: Nice

lib/src/native/realm_core.dart
```diff
@@ -279,24 +286,18 @@ class _RealmCore {
...
-    _realmLib.realm_dart_sync_on_subscription_set_state_change_async(
-      subscriptions.handle._pointer,
-      notifyWhen.index,
-      Pointer.fromFunction(_stateChangeCallback),
-      completer.toPersistentHandle(),
-      _realmLib.addresses.realm_dart_delete_persistent_handle,
-      scheduler.handle._pointer,
-    );
+    final callback = Pointer.fromFunction<Void Function(Handle, Int32)>(_stateChangeCallback);
+    final userdata = _realmLib.realm_dart_userdata_async_new(completer, callback.cast(), scheduler.handle._pointer);
```
nielsenko, line 298: We could introduce a extension method, so that we could write something like:
```dart
final userdata = completer.asUserdataAsync(callback.cast());
```

src/CMakeLists.txt
```diff
@@ -10,18 +10,15 @@ add_subdirectory(dart-dl)
...
 
 set(HEADERS
     realm_dart.h
+    realm_dart.hpp
     realm_dart_scheduler.h
+    realm_dart_sync.h
     realm-core/src/realm.h
-    subscription_set.h
-    sync_session.h
-    sync_client_config.h
```
nielsenko, line 24: 👍

src/realm_dart.hpp
```diff
@@ -16,15 +16,24 @@
...
+    , dart_callback(callback)
+    , scheduler(*scheduler)
+    { }
 
-#endif // REALM_DART_SYNC_CLIENT_CONFIG_H
+    ~realm_dart_userdata_async() {
+        Dart_DeletePersistentHandle_DL(handle);
+    }
+
+    Dart_PersistentHandle handle;
```
nielsenko, line 36: So we always take persistent handle. We will need to revisit how how to handle users who forget to cancel their stream subscriptions, but I think we can handle that more elegantly on the dart side with the `WeakReferences` of Dart 2.17

src/realm_dart_sync.cpp
```diff
@@ -0,0 +1,131 @@
...
+{
+    auto ud = reinterpret_cast<realm_dart_userdata_async_t>(userdata);
+    ud->scheduler->invoke([ud, level, message=std::string(message)]() {
+        (reinterpret_cast<realm_log_func_t>(ud->dart_callback))(ud->handle, level, message.c_str());
+    });
+}
+
+RLM_API void realm_dart_sync_error_handler_callback(realm_userdata_t userdata, realm_sync_session_t* session, realm_sync_error_t error)
+{
+    // the pointers in error are to stack values, we need to make copies and move them into the scheduler invocation
```
nielsenko, line 34: Much nicer than performing the copy directly between the C-structs 👍

src/realm_dart_sync.cpp
```diff
@@ -0,0 +1,131 @@
...
+        : realm_sync_error_code(error)
+        , message_buffer(error.message)
+        {
+            message = message_buffer.c_str();
+        }
+
+        const std::string message_buffer;
+    };
+
+    std::unique_ptr<realm_dart_sync_error_code> error_copy;
```
nielsenko, line 77: So the copies lifetime is managed here. I like it, but we need to make sure that we are not trying to deallocate the string on the dart side, which we did before

src/realm_dart_sync.cpp
```diff
@@ -0,0 +1,131 @@
...
+    });
+}
+
+RLM_API void realm_dart_sync_progress_callback(realm_userdata_t userdata, uint64_t transferred_bytes, uint64_t total_bytes)
+{
+    auto ud = reinterpret_cast<realm_dart_userdata_async_t>(userdata);
+    ud->scheduler->invoke([ud, transferred_bytes, total_bytes]() {
+        (reinterpret_cast<realm_sync_progress_func_t>(ud->dart_callback))(ud->handle, transferred_bytes, total_bytes);
+    });
+}
```
nielsenko, line 94: Here it really shines :-)

nielsenko (approved):
I'm a big fan of this approach. As _the_ someone who did most of the existing trampolines, I really appreciate what you have achieved here.

I know that not all ambitions where met. We need to add some clunky type arguments on the dart side, and we still need to explicitly wrap Dart object in persistent (or weak persistent) handles, but I still think is so much better.

We are getting close to cutting the next pre-release, and this is a big change, but you have my full endorsement for landing this!

blagoev (changes requested):
I did not review this yet, but want to make sure we are not merging this now.

lib/src/native/realm_core.dart
```diff
@@ -70,7 +72,7 @@ class _RealmCore {
     return _instance ??= _RealmCore._();
   }
 
-  String get libraryVersion => '0.3.1+beta';
+  String get libraryVersion => '0.2.1+alpha';
```
desistefanova, line 75:
```suggestion
  String get libraryVersion => '0.3.1+beta';
```

desistefanova (approved):

lib/src/list.dart
```diff
@@ -199,7 +199,7 @@ class ListNotificationsController<T extends Object> extends NotificationsControl
   }
 
   @override
-  void onChanges(Handle changesHandle) {
+  void onChanges(HandleBase changesHandle) {
```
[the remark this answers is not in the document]
blagoev: In that case I would suggest renaming to RealmHandle

blagoev, line 202:
```suggestion
  void onChanges(RealmHandle changesHandle) {
```

src/realm_dart.cpp
```diff
@@ -92,63 +86,15 @@ RLM_API void realm_dart_delete_persistent_handle(void* handle) {
...
-    auto scheduler = std::get<0>(*u);
-    auto userdata_free = std::get<2>(*u);
-    (*scheduler)->invoke([userdata_free = std::move(userdata_free), u]{
-        userdata_free();
-        delete u;
-    });
+RLM_API realm_dart_userdata_async_t realm_dart_userdata_async_new(Dart_Handle handle, void* callback, realm_scheduler_t* scheduler)
+{
+    return new realm_dart_userdata_async(handle, callback, scheduler);
 }
```
blagoev, line 92: Why not create this in Dart by exposing the realm_dart_userdata_async struct

blagoev: this will save us one native transition. But also callback registration happens once per registration so its not on a hot path.

lib/src/native/realm_core.dart
blagoev: I really think we should try allocating this struct in Dart.

lib/src/native/realm_core.dart
```diff
@@ -1059,18 +1053,13 @@ class _RealmCore {
...
-      // Don't do expensive utf8 to utf16 conversion unless needed.
-      if (logger.isLoggable(level)) {
-        logger.log(level, message.cast<Utf8>().toDartString());
-      }
-    } finally {
-      _realmLib.realm_free(message.cast()); // always free the message.
+    // Don't do expensive utf8 to utf16 conversion unless needed.
+    if (logger.isLoggable(level)) {
+      logger.log(level, message.cast<Utf8>().toDartString());
     }
```
blagoev, line 1063: I believe we should keep the `realm_free` call here

blagoev: or I am missing some new change elsewhere?

lib/src/native/realm_core.dart
```diff
@@ -1509,80 +1495,64 @@ class _RealmCore {
...
-        completer.completeError(RealmException(errorCode.toSyncError().toString()));
-      } else {
-        completer.complete();
-      }
-    } finally {
-      _realmLib.realm_free(errorCode.cast());
+    if (errorCode != nullptr) {
+      completer.completeError(errorCode.toSyncError());
+    } else {
+      completer.complete();
```
blagoev, line 1555: Are we sure `realm_free` is not needed here

lib/src/native/realm_core.dart
```diff
@@ -1509,80 +1495,64 @@ class _RealmCore {
...
-      if (errorCode != nullptr) {
-        // Throw RealmException instead of RealmError to be recoverable by the user.
-        completer.completeError(RealmException(errorCode.toSyncError().toString()));
-      } else {
-        completer.complete();
-      }
-    } finally {
-      _realmLib.realm_free(errorCode.cast());
+    if (errorCode != nullptr) {
+      completer.completeError(errorCode.toSyncError());
```
blagoev, line 1553: This will throw SyncError instead of RealmException. Why are we doing this drastic change here? There is a comment that we really want to throw RealmException at this point.

lib/src/native/realm_core.dart
```diff
@@ -1940,7 +1910,7 @@ extension on Pointer<Utf8> {
 
 extension on realm_sync_error {
   SyncError toSyncError() {
-    final message = detailed_message.cast<Utf8>().toRealmDartString(freeRealmMemory: true)!;
+    final message = detailed_message.cast<Utf8>().toRealmDartString()!;
```
blagoev, line 1913: are we sure we don't need to free the realm memory here

lib/src/native/realm_core.dart
```diff
@@ -1956,7 +1926,7 @@ extension on realm_sync_error {
 
 extension on Pointer<realm_sync_error_code_t> {
   SyncError toSyncError() {
-    final message = ref.message.cast<Utf8>().toRealmDartString(freeRealmMemory: true)!;
+    final message = ref.message.cast<Utf8>().toDartString();
```
blagoev, line 1929: are we sure we don't need to free the realm memory here

blagoev (changes requested):

src/realm_dart_sync.cpp
```diff
@@ -0,0 +1,140 @@
...
+    } buf;
+
+    realm_http_request_t request_copy = request; // copy struct
+
+    buf.url = request.url;
+    request_copy.url = buf.url.c_str();
+    buf.body = std::string(request.body, request.body_size);
+    request_copy.body = buf.body.data();
+
+    buf.headers_vector.reserve(request.num_headers);
```
blagoev, line 40: Core is not using this. I suggest we update core as well if we really thing it makes a difference

src/realm_dart.cpp
fealebenpae: `realm_dart_userdata_async_t` is a struct with C++ members and cannot be easily allocated from Dart. What's more, it has a `Dart_PersistentHandle` member which we can't set from Dart - FFI only supports `Dart_Handle` types as parameters or return values of function calls. That's why this has to be constructed in native.

lib/src/native/realm_core.dart
fealebenpae: The string memory is owned by the `realm_dart_sync_client_log_callback` trampoline, we don't need to free it in Dart.

lib/src/native/realm_core.dart
fealebenpae: The string memory is owned by the `realm_dart_sync_wait_for_completion_callback` trampoline, we don't need to free it in Dart.

lib/src/native/realm_core.dart
fealebenpae: All sync error buffers are owned by the native trampolines, we don't need to free them in Dart.

lib/src/native/realm_core.dart
fealebenpae: Same as above.

src/realm_dart_sync.cpp
fealebenpae: Core uses the headers map in the request to make authenticated REST calls using the `Authorization` header.

lib/src/list.dart
fealebenpae: Will rename accordingly.

lib/src/native/realm_core.dart
fealebenpae: That was a bad merge on my part, I'll fix it.

lib/src/list.dart
fealebenpae: Actually we already have a `RealmHandle` type. What alternative do you prefer - `CoreHandle`, `CapiHandle`, `RealmHandleBase`, or something else?

lib/src/native/realm_core.dart
fealebenpae: Reverted. Though now that I look at it I'm not sure why we're wrapping the `SyncError` message in a `RealmException` and ignoring the error code and category members - wouldn't those be useful to users handling sync errors?

lib/src/native/realm_core.dart
blagoev: I guess this is owned now in the native struct copy

lib/src/native/realm_core.dart
blagoev: leaving open just to confirm it is owned in the native struct copy

src/realm_dart.cpp
blagoev: I guess we were doing two transitions before to create the persistence handles so resolving this comment

src/realm_dart_sync.cpp
blagoev: but unrelated to this PR

blagoev: this is what @nielsenko suggested before, so lets leave it here. Would like Core to use this as well

lib/src/list.dart
blagoev: I suggest we name these `RealmHandle` to better reflect the difference with Dart `Handle`

lib/src/native/realm_core.dart
blagoev: ignore we were doing two calls before

lib/src/list.dart
blagoev: argh...

blagoev (approved):

PR #635, opened 2022-06-01
fix throws

blagoev:
throw RealmException where needed


nielsenko (approved):

desistefanova (commented):
I prefer when throwing something it to be an Exception as we discussed. I can not understand why sometimes it is RealmError, sometimes it is RealmException. We discussed to have an Exception type that to keep the error internally.

desistefanova (approved):

blagoev:
We have discussed this offline

PR #640, opened 2022-06-01
Add realm_dart_http_request_callback trampoline

nielsenko:


src/realm_dart_sync.cpp
```diff
@@ -21,10 +21,44 @@
 #include "realm_dart.hpp"
 #include "realm_dart_sync.h"
 
+#include <iostream>
+
+using namespace std;
```
fealebenpae, line 26: We don't need these.

fealebenpae (changes requested):
What about the change we discussed yesterday about making the Dart `_request_callback` consider the `body_size` member of the `realm_http_request_t` struct when converting the body bytes into a Dart string?

src/realm_dart_sync.cpp
nielsenko: Fixed

fealebenpae (approved):

desistefanova (approved):

PR #643, opened 2022-06-02
fix headers copy

blagoev:


src/realm_dart.cpp
```diff
@@ -108,19 +108,16 @@ void _callback(realm_userdata_t userdata, const realm_http_request_t request, vo
...
     } buf;
 
-    realm_http_request_t request_copy = request; // copy struct
-
     buf.url = request.url;
-    request_copy.url = buf.url.c_str();
-    buf.body = std::string(request.body, request.body_size);
-    request_copy.body = buf.body.data();
-
-    buf.headers_vector.reserve(request.num_headers);
```
nielsenko, line 118: I would keep the reserve to ensure we don't need to regrow on a later push_back

blagoev: This is a premature optimization which was causing the crash.

blagoev: let's leave optimizations for after we have identified performance issues.

nielsenko: No, the crash was due to the way the code assumed that `buf.headers_vector[i]` actually was already a `realm_http_header_t`. Lets keep the reserve.

blagoev: Calling reserve here brings tiny, unmeasurable and negligible performance difference which even Core is not doing and is not worth the discussion. I would like to merge this PR.

nielsenko: It adds memory pressure, because vector has no way of knowing how much memory will be needed. But if you feel strongly for it, skip it..

nielsenko (approved):

PR #647, opened 2022-06-03
Use only one cluster for tests

nirinchev:
This modifies the PR workflow to use a single cluster with multiple applications. Each application uses its own database, so they should not be clashing with each other.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,7 @@ jobs:
             --api-key ${{ secrets.ATLAS_QA_PUBLIC_API_KEY }} \
             --private-api-key ${{ secrets.ATLAS_QA_PRIVATE_API_KEY }} \
             --project-id ${{ secrets.ATLAS_QA_PROJECT_ID }} \
-            --database-prefix "${{ matrix.package }}-${{ matrix.platform }}"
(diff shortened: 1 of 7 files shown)


nielsenko (approved):
LGTM

desistefanova (commented):
The changes in the PR looks to be correct. And it is great to see the checks in green. But I'm thinking about the reason to have one cluster. We have enough clusters allowed, we just have to remove them on time instead of waiting all the tasks to complete after half an our. 

I see the CI in the following way: After we have a common build for all the platforms I hoped it will be easier to restructure the CI so that every test to be possible to be run separately. We can have a reusable workflow, as we talked, and this flow to be responsible for running cluster, deploying, building, executing tests and clearing the cluster. If we can achieve that we may have one matrix running this workflow for all the environments. 
This will help us:
 - to be able to re-run only specific test on fail. We don't need to re-run all the tests every time when something fails. 
 - once the tests for an environment complete the cluster to be release immediately. 
 - all the build and tests to be run in parallel and they will not depends on each other. Currently we wait all the builds to pass before to start testing at any platform.
 
 I hope we will agree to discuss this solution and to think whether it is achievable.

blagoev (approved):

blagoev:
@desistefanova Perhaps you could add a discussion topic for the meeting

PR #649, opened 2022-06-07
Run dart tests on M1

nirinchev:
This adds tests for M1 macs on a self-hosted runner. I changed the dart test job to be a matrix since all the steps were identical.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -267,6 +267,9 @@ jobs:
         with:
           submodules: false
 
+      - name: Cleanup Workspace
+        run: git clean -fdx
+
       - name: Fetch artifacts
         uses: actions/download-artifact@v2
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -277,6 +277,7 @@ jobs:
         uses: dart-lang/setup-dart@main
         with:
           sdk: stable
+          architecture: ${{ matrix.architecture == 'arm' && 'arm64' || 'x64'}}
 
       - name: Install dependencies
         run: dart pub get
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -244,7 +244,7 @@ jobs:
 
   tests-dart:
     runs-on: ${{ matrix.platform == 'linux' && 'ubuntu' || matrix.platform }}-${{ matrix.architecture || 'latest' }}
-    name: Tests macOS
+    name: Tests ${{ matrix.platform }} ${{ matrix.architecture }}
     strategy:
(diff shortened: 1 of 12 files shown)


nielsenko (approved):
LGTM

test/test.dart
```diff
@@ -415,3 +418,20 @@ extension RealmObjectTest on RealmObject {
 }
 
 void clearCachedApps() => realmCore.clearCachedApps();
+
+Future<void> _printPlatformInfo() async {
```
blagoev, line 422: Is this info useful when running locally. We print the PID cause its is useful when debugging locally. The other stuff might be only useful on CI but not sure about that. Not sure running uname each time is beneficial especially locally.  So should we be doing this only when REALM_CI is set?

.github/workflows/ci.yml
```diff
@@ -240,14 +240,23 @@ jobs:
...
-    name: Tests Linux
+  tests-dart:
+    runs-on: ${{ matrix.platform == 'linux' && 'ubuntu' || matrix.platform }}-${{ matrix.architecture || 'latest' }}
+    name: Tests ${{ matrix.platform }} ${{ matrix.architecture }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - platform: macos
+          - platform: macos
```
blagoev, line 251: hos is this playing with artifacts in binary/macos

test/test.dart
nirinchev: The main reason why I wanted to add `uname` is because on macOS it's super easy to run under rosetta accidentally (there's no prompt or anything). So it's kind of hard to determine whether you're testing on arm or x64. Locally, I don't see a meaningful impact of running `uname` but can disable it if it impacts people.

.github/workflows/ci.yml
```diff
@@ -513,8 +455,6 @@ jobs:
           flutter drive --target=test_driver/app.dart --dart-define=testName="" --suppress-analytics
         working-directory: ./flutter/realm_flutter/tests
 
-# Android jobs
```
blagoev, line 516: these were useful to navigate this big file

blagoev (changes requested):
c

.github/workflows/ci.yml
nirinchev: There's only one android job, so not sure it adds much value anymore.

.github/workflows/ci.yml
nirinchev: Those are each running on a separate machine, so there should be nothing to overwrite.

blagoev (approved):

PR #657, opened 2022-06-07
Fix the ReserveCache error in the native-build job matrix

fealebenpae:
The cache keys used by the `setup-vcpkg` action were not definite enough and this caused a ReserveCache error when two jobs tried to reserve the same cache at the same time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,8 +131,8 @@ jobs:
         uses: friendlyanon/setup-vcpkg@v1
         with:
           path: src/realm-core/tools/vcpkg/ports
-          cache-key: ${{ matrix.preset }}
-          cache-restore-keys: ${{ matrix.preset }}
+          cache-key: vcpkg-${{ matrix.preset }}-${{ hashFiles('./src/realm-core/tools/vcpkg/**') }}
+          cache-restore-keys: vcpkg-${{ matrix.preset }}-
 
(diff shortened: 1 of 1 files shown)


nielsenko (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -131,8 +131,8 @@ jobs:
         uses: friendlyanon/setup-vcpkg@v1
         with:
           path: src/realm-core/tools/vcpkg/ports
-          cache-key: ${{ matrix.preset }}
-          cache-restore-keys: ${{ matrix.preset }}
+          cache-key: vcpkg-${{ matrix.preset }}-${{ hashFiles('./src/realm-core/tools/vcpkg/**') }}
```
nirinchev, line 134: Rather than hashing the contents of vcpkg, can we use the submodule sha instead?

nirinchev (commented):
Looks good - just wondering if we can avoid the hashfiles which has unpredictable performance.

.github/workflows/ci.yml
fealebenpae: There isn't a clear-cut way I can get that outside of a multi-line git invocation, I think. I can just fall back to `github.sha` - how does that sound?

nirinchev (approved):

blagoev (approved):

PR #660, opened 2022-06-08
Fix flutter test path to realm_dart.dll

desistefanova:
When running flutter tests on windows an error appears that realm_dart.dll can not be loaded. 
It is because executable path for flutter tests is not in the local app directory.
The same problem exists also on Linux and MacOS.
Task: #791

### DIFF
--- a/flutter/realm_flutter/example/test/widget_test.dart
+++ b/flutter/realm_flutter/example/test/widget_test.dart
@@ -1,0 +1,18 @@
+// This is a basic Flutter widget test.
+//
+// To perform an interaction with a widget in your test, use the WidgetTester
+// utility in the flutter_test package. For example, you can send tap and scroll
+// gestures. You can also use WidgetTester to find child widgets in the widget
+// tree, read text, and verify that the values of widget properties are correct.
+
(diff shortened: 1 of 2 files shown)


blagoev (changes requested):
This should not be needed. Lets investigate why is this happening.

desistefanova:
> This should not be needed. Lets investigate why is this happening.

The executable file for Flutter is:
 \realm-dart\flutter\realm_flutter\example\build\windows\runner\Debug\realm_example.exe
and it is able to find the realm_dart.dll by name only.
But when we run the flutter tests the executable file is:
\flutter\bin\cache\artifacts\engine\windows-x64\flutter_tester.exe
then realm_dart.dll can not be loaded by name only. We need to pass the whole path from the current folder. "./build/windows/runner/Debug/$binaryName.dll";

nirinchev:
Why does this work on CI then? 🤔

desistefanova:
Windows:
image

MacOS:
image

Linux:
image

desistefanova:
Linked to a task. Will be implemented in a new PR.

PR #663, opened 2022-06-09
CI Jobs re-run per environment

desistefanova:


nielsenko (commented):
I can appreciate the effort that went into this PR, but I think it makes our test-runs too interdependent.

The nice thing about this PR is that it allows us to rerun only failed jobs, but I think there are easier and safer ways to achieve this.

Personally I would prefer to just "leak" the cluster on failed runs. Ie. just drop the `if: always()` on the cluster cleanup job, as compared to today,

The cluster will be reaped after 12 hours anyway and we can always run the wipe job if we hit a usage loft.

This will allow us to rerun _failed jobs only_ and also leave the cluster intact for debugging test failures.

That said, I'm not going to veto this. Just my 2 cents.

.github/workflows/binary-combine-android.yml
```diff
@@ -0,0 +1,51 @@
+name: Android Build Combine
+
+on:
+  workflow_call:
+   inputs:
+      os:
```
nirinchev, line 6: Why do we need to pass the OS here? We can just hardcode it to use ubuntu always.

.github/workflows/binary-combine-ios.yml
```diff
@@ -0,0 +1,54 @@
+name: IOS Build Combine
+
+on:
+  workflow_call:
+   inputs:
+      os:
```
nirinchev, line 6: Same here - this can only run on macos, so what's the point of specifying the os externally?

.github/workflows/ci.yml
```diff
@@ -157,117 +100,51 @@ jobs:
...
+
+      - name: Generate unique job id
+        id: generate_job_id
+        if: ${{matrix.os != 'windows'}}
+        run: echo "BAAS_DIFFERENTIATOR=d$RANDOM$100" >> $GITHUB_ENV
+        
+      - name: Generate unique job id on Windows
+        id: generate_job_id_win
+        if: ${{matrix.os == 'windows'}}
+        run: echo "BAAS_DIFFERENTIATOR=d$(Get-Random)" >> $env:GITHUB_ENV
```
nirinchev, line 146: Discussed offline, we don't need the random here.

.github/workflows/ci.yml
```diff
@@ -392,17 +302,26 @@ jobs:
...
 
+      - name: Cleanup apps
+        if: always()
+        run: dart run realm_dart delete-apps --appIds '${{ steps.deploy-apps.outputs.appIds }}' --baas-url ${{ secrets.REALM_QA_URL }} --atlas-cluster ${{ needs.deploy-cluster.outputs.clusterName }} --api-key ${{ secrets.ATLAS_QA_PUBLIC_API_KEY }} --private-api-key ${{ secrets.ATLAS_QA_PRIVATE_API_KEY }} --project-id ${{ secrets.ATLAS_QA_PROJECT_ID }} --differentiator '${{ env.BAAS_DIFFERENTIATOR }}'
+
+# Android jobs
+
   flutter-android:
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
```
nirinchev, line 312: We can't run the android emulator on ubuntu - it's insanely slow.

lib/src/cli/deployapps/deleteapps_command.dart
```diff
@@ -0,0 +1,84 @@
...
+
+      if (options.projectId == null) {
+        abort('--project-id must be supplied when --atlas-cluster is not set');
+      }
+    }
+    if (appIds.isEmpty) {
+      abort('--appIds must be supplied');
+    }
+
+    final differentiator = options.differentiator ?? 'local';
```
nirinchev, line 67: Why do we need to pass a differentiator here if we're passing the final app id to be deleted? The differentiator won't be used at all.

.github/workflows/ci.yml
```diff
@@ -279,85 +156,114 @@ jobs:
...
       - name: Setup Flutter
         uses: subosito/flutter-action@v2
         with:
           channel: 'stable'
 
       - name: Enable Flutter Desktop support
-        run: flutter config --enable-${{ matrix.platform }}-desktop
+        run: flutter config --enable-${{ matrix.os }}-desktop
+
+      - name: Deploy Apps for Dart tests on ${{ matrix.os }}
```
nirinchev, line 235: One thing I don't like about this PR is that it moves the deploy apps step inside the test step. This means, we're adding 45-60 seconds of deploy time + a few seconds of cleanup time to each test job, when those can be done in parallel before the tests even start.

lib/src/cli/deployapps/deleteapps_command.dart
```diff
@@ -0,0 +1,84 @@
...
+
+  DeleteAppsCommand() {
+    populateOptionsParser(argParser).addOption("appIds", help: "List of appIds of deployed apps on MongoDB Atlas.");
+  }
+
+  @override
+  FutureOr<void>? run() async {
+    options = parseOptionsResult(argResults!);
+
+    var result = argParser.parse(argResults!.arguments);
```
nirinchev, line 47: That doesn't seem correct to me - why not create dedicated options class for this command?

lib/src/cli/deployapps/deployapps_command.dart
```diff
@@ -67,9 +67,13 @@ class DeployAppsCommand extends Command<void> {
     final apps = await client.getOrCreateApps();
 
     print('App import is complete. There are: ${apps.length} apps on the server:');
+    String listApps = '';
```
nirinchev, line 70: Why not make that an actual list and then use `string.join` when printing the ids?

nirinchev (changes requested):

desistefanova:
> I can appreciate the effort that went into this PR, but I think it makes our test-runs too interdependent.
> 
> The nice thing about this PR is that it allows us to rerun only failed jobs, but I think there are easier and safer ways to achieve this.
> 
> Personally I would prefer to just "leak" the cluster on failed runs. Ie. just drop the `if: always()` on the cluster cleanup job, as compared to today,
> 
> The cluster will be reaped after 12 hours anyway and we can always run the wipe job if we hit a usage loft.
> 
> This will allow us to rerun _failed jobs only_ and also leave the cluster intact for debugging test failures.
> 
> That said, I'm not going to veto this. Just my 2 cents.

@nielsenko The cluster is not deleted by the workflow anymore. We leave it alive as we discussed. We just check in the beginning whether it exists and if not we create it.

.github/workflows/ci.yml
desistefanova: If we move the app deployment in another jobs we will not be able to rerun the failed tests without having the app created. BTW our tests are already creating the their apps, so that if the tests are rerun they will create their apps again. So that, we can move deployment apps in a separate job and to rely that our tests are creating their apps only if they have been rerun. Just have to think how to pass app Ids to the deleting apps action.

lib/src/cli/deployapps/deleteapps_command.dart
desistefanova: Yes, this folder and commands might be restructured.

.github/workflows/ci.yml
desistefanova: I repaired the workflow so that the deployments to be run in a matrix before all the tests. Unfortunately, the jobs for creating apps are taking more that the native builds time and they are somehow blocked.  For this reason it appeared that we are waiting even more. I returned deployment apps back into the tests execution job.

.github/workflows/binary-combine-android.yml
desistefanova: Done

.github/workflows/ci.yml
desistefanova: Deleted

.github/workflows/ci.yml
desistefanova: Done

lib/src/cli/deployapps/deleteapps_command.dart
desistefanova: I changed the implementation the apps to be discovered by differentiator instead of appIds. I did this in order to be able to move all the deployments before all the tests.

lib/src/cli/deployapps/deployapps_command.dart
desistefanova: Done, even though I don't need these appIds anymore.

lib/src/cli/deployapps/deleteapps_command.dart
desistefanova: I changed the name of the folder from deployapps to atlas_apps then I used the same Options for both commands, since after refactoring to use the differentiator both commands have the same options.

lib/src/cli/deployapps/deployapps_command.dart
blagoev: if its not needed lets remove it, but I think dumping the existing apps could help to debug something in the future.

.github/workflows/ci.yml
```diff
@@ -157,117 +104,42 @@ jobs:
...
-    name: Tests ${{ matrix.platform }} ${{ matrix.architecture }}
+  dart-tests:
     strategy:
       fail-fast: false
       matrix:
         include:
-          - platform: macos
-          - platform: macos
+          - os: macos
+            app: dm
```
blagoev, line 125: nit: Why abbreviate the app names and not name these apps `DartMac` `DartMacArm` `DartWindows` etc, hope the short name does not haunt us in the future while debugging something.

lib/src/cli/atlas_apps/baas_client.dart
```diff
@@ -154,7 +154,7 @@ class BaasClient {
   }
 
   Future<BaasApp> _createApp(String name, {String? confirmationType}) async {
-    print('Creating app $name ($name$_appSuffix)');
+    print('Creating app $name$_appSuffix');
```
blagoev, line 157: yeah definitely need better app names if we are outputting that

lib/src/cli/atlas_apps/baas_client.dart
```diff
@@ -344,9 +365,9 @@ class BaasClient {
     if (input.length < 8) {
       return input;
     }
-
-    final hash = md5.convert(utf8.encode(input)).toString();
-    return hash.substring(0, 8);
+    //Take first 4 and last 4 symbols
+    final result = "${input.substring(0, 4)}${input.substring(input.length - 4)}";
```
blagoev, line 369: seems `replaceRange` is for that purpose 
```input.replaceRange(4, input.length - 4, "");```

File: lib/src/cli/atlas_apps/deployapps_command.dart [hunk 12]
```diff
@@ -67,9 +67,13 @@ class DeployAppsCommand extends Command<void> {
     final apps = await client.getOrCreateApps();
 
     print('App import is complete. There are: ${apps.length} apps on the server:');
+    List<String> listApps = [];
     apps.forEach((_, value) {
       print("  App '${value.name}': '${value.clientAppId}'");
+      listApps.add(value.appId);
     });
+    print("appIds: ");
```
blagoev, line 75:
```suggestion
    print("appIds: ${listApps.join(",")}");
```

lib/src/cli/atlas_apps/deployapps_command.dart
```diff
@@ -67,9 +67,13 @@ class DeployAppsCommand extends Command<void> {
...
+    stdout.write(listApps.join(","));
```
blagoev, line 76:
```suggestion
   
```

lib/src/cli/atlas_apps/deployapps_command.dart
blagoev: I think this will work

blagoev (changes requested):
some final changes

lib/src/cli/atlas_apps/deployapps_command.dart
desistefanova: I printed these Ids on a new line by reason. It is easier to read them as an output of the command. I removed already the usage of this output, but I think they could help us for something else.

lib/src/cli/atlas_apps/baas_client.dart
desistefanova: We have length limitation.

.github/workflows/ci.yml
blagoev: ok I see, well in that case we don't have a choice

lib/src/cli/atlas_apps/deployapps_command.dart
blagoev: ok thanks

blagoev (approved):

lib/src/cli/atlas_apps/deployapps_command.dart
nirinchev, line 76: Why is this using stdout instead of print?

nirinchev (changes requested):
I've suggested #673 as an improvement to move the deploy/cleanup steps to separate jobs and speed up the workflow execution.

lib/src/cli/atlas_apps/deployapps_command.dart
desistefanova: I didn't know I would be able to read the output from print. It is the same actually. I can change it.

lib/src/cli/atlas_apps/baas_client.dart
```diff
@@ -344,9 +365,9 @@ class BaasClient {
     if (input.length < 8) {
       return input;
     }
-
-    final hash = md5.convert(utf8.encode(input)).toString();
-    return hash.substring(0, 8);
+    //Take first 4 and last 4 symbols
```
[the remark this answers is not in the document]
desistefanova: I wanted it to be visible in the cloud UI which database for which app is and whether we are creating apps and dbs for all the runs. For this reason I added custom short app prefixes in the matrixes for example 'dw' for 'dart windows'. And then concatenate to it run id and attempt id. When we are building an app name we take the first 4 and the last 4 symbols from differentiators in order to differentiate the app preffix, which is at the beginning, and the attempt id, which is at the end. Sample: --differentiator 'fw25079812051' -appname: flexible-fw252051-dart16AM
Database name is db_{appName} - -database: db_flexible-fw252051-dart16AM

.github/workflows/ci.yml
```diff
@@ -18,76 +18,23 @@ jobs:
...
+          realmUrl: ${{ env.BAAS_URL }}
           atlasUrl: ${{ secrets.ATLAS_QA_URL }}
-          projectId: ${{ secrets.ATLAS_QA_PROJECT_ID}}
-          apiKey: ${{ secrets.ATLAS_QA_PUBLIC_API_KEY}}
-          privateApiKey: ${{ secrets.ATLAS_QA_PRIVATE_API_KEY }}
-          differentiator: realm-dart
+          projectId: ${{ env.BAAS_PROJECT_ID }}
+          apiKey: ${{ env.BAAS_API_KEY }}
+          privateApiKey: ${{ env.BAAS_PRIVATE_API_KEY }}
+          differentiator: ${{ steps.cluster-name.outputs.clusterName }}
```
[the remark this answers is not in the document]
desistefanova: Done

lib/src/cli/atlas_apps/baas_client.dart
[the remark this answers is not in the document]
nirinchev: Ah, my bad - I didn't realize you removed the hashing.

.github/workflows/ci.yml
```diff
@@ -18,76 +18,23 @@ jobs:
...
-      - flutter-desktop
-      - flutter-ios
-      - flutter-android
-      - tests-dart
-    if: always()
-    timeout-minutes: 5
-    steps:
-      - uses: realm/ci-actions/mdb-realm/cleanup@fac1d6958f03d71de743305ce3ab27594efbe7b7
+      - uses: realm/ci-actions/mdb-realm/deployApps@916d1585f7e066431830b195cdb8cf7855747c2a
+        id: deploy-mdb-apps
```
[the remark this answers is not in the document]
desistefanova: OK

.github/workflows/ci.yml
```diff
@@ -291,73 +274,101 @@ jobs:
...
           channel: 'stable'
 
       - name: Enable Flutter Desktop support
-        run: flutter config --enable-${{ matrix.platform }}-desktop
+        run: flutter config --enable-${{ matrix.os }}-desktop
 
       - name: Install dependencies
         run: flutter pub get
 
+      - uses: realm/ci-actions/mdb-realm/deployApps@63fb0c1678dd750e92709bb6c78aaafd03c5fade
```
nirinchev, line 346: Might be a good idea to add a comment here and for the other jobs
```suggestion
      # This will be a no-op under normal circumstances since the cluster would have been deployed
      # in deploy-cluster. It is needed in case we want to re-run the job after the cluster has been reaped.
      - uses: realm/ci-actions/mdb-realm/deployApps@63fb0c1678dd750e92709bb6c78aaafd03c5fade
```

nirinchev (approved):
Still a little concerned about the connection situation, but we can tackle that later.

PR #665, opened 2022-06-10
Use better default path on Flutter

### LINKED ISSUE #450: Use better default path for Flutter Desktop
Currently the default path for Flutter Desktop defaults to current directory. This causes troubles if the application is packaged in an installer and is installed in Program Files directory on Windows for example. In such cases the application does not have write access to the current directory and could only work if it is run as elevated or the developer changes the default path. We believe we should move to the application dir similar to what path_provider enables so packaged Flutter Desktop apps work correctly by default. 

Since path_provider is a Flutter only package we might want to provide a similar implementation without depending directly on path_provider

blagoev:
This introduces better default paths on Flutter. 
/// On Flutter Android and iOS this is the application's data directory.
/// On Flutter Windows this is the `C:\Users\username\AppData\Roaming\app_name` directory.
/// On Flutter macOS this is the `/Users/username/Library/Containers/app_name/Data/Library/Application Support` directory.
/// On Flutter Linux this is the `/home/username/.local/share/app_name` directory.
/// On Dart standalone Windows, macOS and Linux this is the current directory.

fixes #450 
This also reintroduces `Configuration.defaultRealmName` and `Configuration.defaultRealmPath` which the user can set once and all realms will use these fields by default.
This also introduces `Configuration.defaultStoragePath` static getter which point to the platform specific writable application directory on Flutter and the current directory on Dart.


.vscode/settings.json
```diff
@@ -41,6 +41,8 @@
         "xlocmon": "cpp",
         "xmemory": "cpp",
         "xstring": "cpp",
-        "xutility": "cpp"
+        "xutility": "cpp",
+        "iosfwd": "cpp",
+        "sstream": "cpp"
```
nielsenko, line 46: Wy do these get added I wonder?

lib/src/app.dart
```diff
@@ -18,9 +18,9 @@
 import 'dart:io';
 
 import 'package:meta/meta.dart';
+import 'package:path/path.dart' as _path;
```
nielsenko, line 21: I think the linter will complain that you prefix with _ when naming an imported package.

lib/src/configuration.dart
```diff
@@ -47,29 +48,45 @@ typedef InitialDataCallback = void Function(Realm realm);
...
+  
+
+  /// The platform dependent path used to store realm files
+  ///
+  /// On Flutter Android and iOS this is the application's data directory.
+  /// On Flutter Windows this is the `C:\Users\username\AppData\Roaming\app_name` directory.
+  /// On Flutter macOS this is the `/Users/username/Library/Containers/app_name/Data/Library/Application Support` directory.
+  /// On Flutter Linux this is the `/home/username/.local/share/app_name` directory.
+  /// On Dart standalone Windows, macOS and Linux this is the current directory.
+  static String get defaultStoragePath {
```
nielsenko, line 64: I suggest we use `Directory` here instead of `String`

test/test.dart
```diff
@@ -383,8 +377,7 @@ Future<Realm> getIntegrationRealm({App? app, ObjectId? differentiator, String? p
   app ??= App(await getAppConfig());
   final user = await getIntegrationUser(app);
 
-  // TODO: path will not be needed after https://github.com/realm/realm-dart/pull/574
-  final config = Configuration.flexibleSync(user, [Task.schema, Schedule.schema, NullableTypes.schema], path: path);
+  final config = Configuration.flexibleSync(user, [Task.schema, Schedule.schema, NullableTypes.schema]);
```
nielsenko, line 380: Are we sure we don't want to allow users to set the path explicitly? One thing is to have a sane default, another is to prevent the user from choosing himself?

.vscode/settings.json
blagoev: not sure as well, but committing since it does not hurt it seems

lib/src/app.dart
blagoev: no warnings exists

lib/src/configuration.dart
blagoev: what's the benefit. having it as a string makes it easy to combine with other strings. also the others properties are Strings and not `File`

test/test.dart
blagoev: This was a stale TODO that I fixed. since #574 was merged. So I guess either I should be removing the path argument or keeping it?

lib/src/configuration.dart
nielsenko: The benefit being that the user needs to pass a valid path.

lib/src/app.dart
nielsenko: info - src/app.dart:21:36 - Avoid leading underscores for library prefixes. - no_leading_underscores_for_library_prefixes

nielsenko: https://dart.dev/guides/language/effective-dart/style#dont-use-a-leading-underscore-for-identifiers-that-arent-private

test/test.dart
nielsenko: Sorry.. I read this wrong. For a minute there I thought you removed the argument on the ctor 🙄

lib/src/configuration.dart
blagoev: This is a getter. There is no way for the user to pass a value here. Also on other places we use Strings for the paths

lib/src/app.dart
blagoev: I don't see a warning in the editor

blagoev: Also I think we were already using this naming elsewhere

test/test.dart
blagoev: no it was good. I removed this path argument on multiple places which were leftovers

nielsenko (approved):

PR #671, opened 2022-06-15
Update to latest Core

nirinchev:
Pulls in the fix from Core that introduces a sleep between polling when no sessions are active (https://github.com/realm/realm-core/pull/5594).


blagoev (approved):

blagoev:
seems it needs ffi update

PR #673, opened 2022-06-17
Use a matrix to deploy and cleanup apps

nirinchev:
This is building on top of https://github.com/realm/realm-dart/pull/663 by moving the deploy/cleanup steps to separate jobs. This shaves 45-60 seconds from the test jobs and those typically complete by the time the build-native job completes, so tests are ready to run in about a  minute.

The reason why this works is twofold:
1. We can split the deploy-apps job because our tests do not rely on it. The test runner will import the apps if necessary, so in the event of a re-run of the failed job, the test runner will fail to find the apps and import them itself.
2. We can split the cleanup-apps job because re-running a failed job will also re-run its dependents. In the event of a re-run, the cleanup will still be executed with the run attempt, so it will delete apps created by the test runner.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,20 +20,15 @@ jobs:
     outputs:
       clusterName: ${{ steps.cluster-name.outputs.clusterName }}
     steps:
-      - name: Get cluster suffix
-        id: cluster-name
-        run: echo "::set-output name=clusterName::dart$(date +'%d%p')"
-
       - uses: realm/ci-actions/mdb-realm/deployApps@916d1585f7e066431830b195cdb8cf7855747c2a
-        id: deploy-mdb-apps
         with:
           realmUrl: ${{ env.BAAS_URL }}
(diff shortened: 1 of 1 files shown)


blagoev (approved):

.github/workflows/ci.yml
```diff
@@ -18,22 +18,18 @@ jobs:
     runs-on: ubuntu-latest
     name: Deploy Cluster
     outputs:
-      clusterName: ${{ steps.cluster-name.outputs.clusterName }}
+      clusterName: ${{ steps.deploy-atlas-cluster.outputs.clusterName }}
     steps:
-      - name: Get cluster suffix
-        id: cluster-name
```
desistefanova, line 24: Could we return back this step for generating different names on every 12 h?

desistefanova (changes requested):
Cluster names have to be renewed on every 12 h to prevent running tests on automatically deleted clusters. Samples for names are dart16AM, dart16PM, dart17AM.

PR #709, opened 2022-07-02
Add Custom JWT Authentication to Credentials

derinwalters:
Hooked JWT Credentials into realm_bindings.dart in a fashion consistent with JS library. Needed for connecting to Firebase Authentication.

PR #715, opened 2022-07-06
Test Custom JWT Authentication to Credentials

desistefanova:
Update Changelog.
Add tests for login with JWT credentials.
Enable custom-token auth provider for the cloud apps using baas_client.dart.
Fixes: #398 
This supersedes #709


test/test.dart
```diff
@@ -320,10 +323,15 @@ Future<void> setupBaas() async {
   final privateApiKey = arguments[argBaasPrivateApiKey];
   final projectId = arguments[argBaasProjectId];
   final differentiator = arguments[argDifferentiator];
+  final jwksUrl = arguments[argJwksUrl];
 
   final client = await (cluster == null
       ? BaasClient.docker(baasUrl, differentiator)
       : BaasClient.atlas(baasUrl, cluster, apiKey!, privateApiKey!, projectId!, differentiator));
+  if (!isFlutterPlatform || (Platform.isWindows || Platform.isLinux)) {
```
desistefanova, line 331: We need to find better location of these files with keys so that they could be accessed by Flutter Android, IOS, MacOS tests also. Currently flutter tests for these platforms don't run the tests for JWT, because of missing path "test/data/jwt_keys/".

test/credentials_test.dart
```diff
@@ -299,4 +302,172 @@ Future<void> main([List<String>? args]) async {
...
+        "birthDay": "1999-10-11",
+        "minAge": "10",
+        "maxAge": "90",
+        "company": "Realm",
+      },
+      issuer: 'https://realm.io',
+      audience: Audience(["mongodb.com"]),
+    );
+
+    String privateKey = File("test/data/jwt_keys/private_key.pem").readAsStringSync();
```
nielsenko, line 325: Consider parsing this as an argument to the test exe to allow to run on Android and iOS as well

.github/workflows/dart-desktop-tests.yml
```diff
@@ -25,6 +25,7 @@ env:
   BAAS_API_KEY: ${{ secrets.ATLAS_QA_PUBLIC_API_KEY }}
   BAAS_PRIVATE_API_KEY: ${{ secrets.ATLAS_QA_PRIVATE_API_KEY }}
   BAAS_PROJECT_ID: ${{ secrets.ATLAS_QA_PROJECT_ID}}
+  BAAS_JWKS_URL: https://raw.githubusercontent.com/realm/realm-dart/${{ github.event_name == 'pull_request' && github.head_ref || github.ref_name }}/test/data/jwt_keys/jwks.json
```
nielsenko, line 28: Not too keen on storing the jwks.json in the repo. Could we generate in when creating the app, and put it in a local file on the BAAS (or perhaps on S3)

test/data/jwt_keys/private_key.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
nielsenko, line 1: Would prefer this to be stored in a github secret. I know it is just for test.

test/data/jwt_keys/public_key.pem
```diff
@@ -0,0 +1,9 @@
+-----BEGIN PUBLIC KEY-----
```
nielsenko, line 1: Would prefer this to be generated from the read secret using openssl

test/data/jwt_keys/jwks.json
```diff
@@ -0,0 +1,7 @@
+{
```
nielsenko, line 1: Would prefer this not to be stored in repo

test/test.dart
nielsenko: if we pass content on as fx argBaasApiKey we don't need to

nielsenko (commented):
We should discuss about the private key management on the a round-table. I know this is just for test, but I think it is worth considering going the extra mile, and show how to avoid it.

nielsenko (approved):

.github/workflows/ci.yml
```diff
@@ -10,6 +10,7 @@ env:
   BAAS_API_KEY: ${{ secrets.ATLAS_QA_PUBLIC_API_KEY }}
   BAAS_PRIVATE_API_KEY: ${{ secrets.ATLAS_QA_PRIVATE_API_KEY }}
   BAAS_PROJECT_ID: ${{ secrets.ATLAS_QA_PROJECT_ID}}
+  BAAS_JWKS_URL: https://raw.githubusercontent.com/realm/realm-dart/${{ github.event_name == 'pull_request' && github.head_ref || github.ref_name }}/test/data/jwt_keys/jwks.json
```
nirinchev, line 13: I feel this is a bit of an overkill. We can test JWT auth with the public key only.

.github/workflows/ci.yml
```diff
@@ -306,7 +307,8 @@ jobs:
       
       # This will be a no-op under normal circumstances since the cluster would have been deployed
       # in deploy-cluster. It is needed in case we want to re-run the job after the cluster has been reaped.
-      - uses: realm/ci-actions/mdb-realm/deployApps@3f810b2d04e9dada2bde0b33ec90102e52a0b30a
+      - name: Try to create cluster
```
nirinchev, line 310:
```suggestion
      - name: Create cluster
```

I feel using the `try to` wording is superficial - every step here is trying to do something and there's no reason to explicitly state that the attempt may fail.

.github/workflows/ci.yml
```diff
@@ -376,7 +378,8 @@ jobs:
 
       # This will be a no-op under normal circumstances since the cluster would have been deployed
       # in deploy-cluster. It is needed in case we want to re-run the job after the cluster has been reaped.
-      - uses: realm/ci-actions/mdb-realm/deployApps@3f810b2d04e9dada2bde0b33ec90102e52a0b30a
+      - name: Try to create cluster
```
nirinchev, line 381:
```suggestion
      - name: Create cluster
```

.github/workflows/flutter-desktop-tests.yml
```diff
@@ -70,7 +71,8 @@ jobs:
 
       # This will be a no-op under normal circumstances since the cluster would have been deployed
       # in deploy-cluster. It is needed in case we want to re-run the job after the cluster has been reaped.
-      - uses: realm/ci-actions/mdb-realm/deployApps@3f810b2d04e9dada2bde0b33ec90102e52a0b30a
+      - name: Try to create cluster
```
nirinchev, line 74:
```suggestion
      - name: Create cluster
```

.github/workflows/dart-desktop-tests.yml
```diff
@@ -61,7 +62,8 @@ jobs:
 
       # This will be a no-op under normal circumstances since the cluster would have been deployed
       # in deploy-cluster. It is needed in case we want to re-run the job after the cluster has been reaped.
-      - uses: realm/ci-actions/mdb-realm/deployApps@3f810b2d04e9dada2bde0b33ec90102e52a0b30a
+      - name: Try to create cluster
```
nirinchev, line 65:
```suggestion
      - name: Create cluster
```

flutter/realm_flutter/tests/test_driver/app_test.dart
```diff
@@ -28,6 +28,8 @@ void main(List<String> args) {
       testCommandWithArgs += getArgFromEnvVariable("BAAS_PRIVATE_API_KEY");
       testCommandWithArgs += getArgFromEnvVariable("BAAS_PROJECT_ID");
       testCommandWithArgs += getArgFromEnvVariable("BAAS_DIFFERENTIATOR");
+      testCommandWithArgs += getArgFromEnvVariable("BAAS_JWKS_URL");
+
```
nirinchev, line 32:
```suggestion
```

test/credentials_test.dart
```diff
@@ -299,4 +299,185 @@ Future<void> main([List<String>? args]) async {
       await authProvider.callResetPasswordFunction(username, newPassword);
     }, throws<RealmException>("failed to reset password for user $username"));
   }, appName: AppNames.autoConfirm);
+
+  /// JWT Header:
+  /// {
+  ///   "kid": "1",
+  ///   "alg": "RS256",
+  ///   "typ": "JWT"
+  /// }
```
nirinchev, line 308: The header is rarely interesting, so I don't think we need to include it in the comments.

test/credentials_test.dart
```diff
@@ -299,4 +299,185 @@ Future<void> main([List<String>? args]) async {
...
+  ///      "lastName": "Doe"
+  ///    },
+  ///    "email": "JWT_privatekey_validated_user@realm.io",
+  ///    "gender": "male",
+  ///    "birthDay": "1999-10-11",
+  ///    "minAge": "10",
+  ///    "maxAge": "90",
+  ///    "company": "Realm",
+  ///    "iat": 1660130537,
+  ///    "exp": 1660349537, //10 years after Aug 2022
```
nirinchev, line 323: I typically make these expirations 100+ years as many projects may last longer than 10 years.

test/credentials_test.dart
```diff
@@ -299,4 +299,185 @@ Future<void> main([List<String>? args]) async {
...
+  ///    "minAge": "10",
+  ///    "maxAge": "90",
+  ///    "company": "Realm",
+  ///    "iat": 1660130241,
+  ///    "exp": 1660349241, //10 years after Aug 2022
+  ///    "aud": "mongodb.com",
+  ///    "iss": "https://realm.io"
+  /// }
+  /// JWKS url validation is configured in autoConfirm app.
+  baasTest('JWT validation by using JWK URI - login', (configuration) async {
```
nirinchev, line 376: This test is identical to the one above and is testing server functionality - I don't think we need to have it. If the test above passes, this means that the token is correctly passed from dart to core to the server. What the server does with the token and how it validates it against the public key or against jwk uri should be irrelevant for us.

test/credentials_test.dart
```diff
@@ -299,4 +299,185 @@ Future<void> main([List<String>? args]) async {
...
+  ///   "birthDay": "1999-10-11",
+  ///   "minAge": "10",
+  ///   "maxAge": "90",
+  ///   "company": "Realm",
+  ///   "iat": 1660131880,
+  ///   "exp": 1660350880, //10 years after Aug 2022
+  ///   "aud": "mongodb.com",
+  ///   "iss": "https://realm.io"
+  /// }
+  baasTest('JWT - login with existing user and edit profile', (configuration) async {
```
nirinchev, line 423: This test can't work like this. Users are distinguished by provider+providerId, so an email password user cannot authenticate with JWT, even if their identities are the same. What you can do is generate two tokens for the same user identity and different profile information. Then validate that both authenticate the same user and that after the second auth, the profile has been updated.

nirinchev (changes requested):
My comments roughly reflect what we discussed at the roundtable yesterday - the code seems fine to me, but I believe we should simplify our tests.

test/credentials_test.dart
desistefanova: I'm just rewriting this test with having one user with two identities for different providers.

.github/workflows/ci.yml
desistefanova: Why not to have it?

.github/workflows/ci.yml
desistefanova: Yes, it is with "try" because the cluster may have been already created.

test/credentials_test.dart
desistefanova: Shall I remove everything related to JWK url?

test/credentials_test.dart
desistefanova: Does it make sense now?

test/data/jwt_keys/jwks.json
desistefanova: It contains only public key. But it is verry likely to remove everything related to the test with JWK url as @nirinchev suggested.

test/data/jwt_keys/private_key.pem
desistefanova: On a later discussion we decided to leave these files alive, but It is not a problem to be deleted after the last changes. @blagoev , @nirinchev shall I delete them?

test/data/jwt_keys/public_key.pem
desistefanova: It will make baas_client.dart more complicated. I think it is no more needed.

test/credentials_test.dart
desistefanova: I hope so. OK, I will change it.

test/credentials_test.dart
desistefanova: I think these both are important:
"kid": "1",
"alg": "RS256"
But I can remove them.

test/credentials_test.dart
desistefanova: It is no more needed. We don't use private key anymore.

.github/workflows/ci.yml
desistefanova: OK, removed.

test/test.dart
desistefanova: Removed.

.github/workflows/dart-desktop-tests.yml
desistefanova: Removed.

test/data/jwt_keys/jwks.json
desistefanova: Removed.

.github/workflows/ci.yml
desistefanova: Done.

test/credentials_test.dart
desistefanova: Removed.

test/credentials_test.dart
desistefanova: Done.

test/credentials_test.dart
desistefanova: JWK url is removed.

test/test.dart
```diff
@@ -272,7 +281,7 @@ Future<void> tryDeleteRealm(String path) async {
     }
   }
 
-  // TODO: File deletions does not work after tests so don't fail for now
+  // File deletions does not work after tests so don't fail for now
```
nirinchev, line 284: I'd rather we keep this as a TODO and add a github ticket to track it. It's not urgent, but there's no reason why a file deletion shouldn't succeed if our test harness is freeing resources correctly.

test/credentials_test.dart
```diff
@@ -299,4 +299,143 @@ Future<void> main([List<String>? args]) async {
...
+      await authProvider.registerUser(username, strongPassword);
+    } on RealmException catch (e) {
+      {
+        if (e.message.contains("name already in use")) {
+          // If the user exists, delete it and register a new one with the same name and empty profile
+          final user1 = await loginWithRetry(app, Credentials.emailPassword(username, strongPassword));
+          await app.deleteUser(user1);
+          await authProvider.registerUser(username, strongPassword);
+        }
+      }
```
nirinchev, line 378: Why hardcode the username here? We should be able to use a random email and avoid that extra logic.

nirinchev (commented):
Looks good, let's just clean up/simplify that profile test.

test/test.dart
desistefanova: Done.

test/credentials_test.dart
desistefanova: Because the token is already generated and it must contain the same username inside.

desistefanova: The logic gets complicated but it will allow us to run the same test many times.

desistefanova: I tried to generate a new username for email/password credentials and then to link to JWT credentials. The email and username of the user are updated with "jwt_user@#r@D@realm.io" that is coming from the token. If I run the test again it creates a new user, but when I try to link it to the credentials with the same token it fails, because "jwt_user@#r@D@realm.io" already exists and it can not rename the new user to the same name. I still have to delete "jwt_user@#r@D@realm.io" before to run the test again.

nirinchev: Yeah, the issue is that the token is already consumed to create a user. The next time the test runs, it tries to link the new user to an existing user which is forbidden.

.github/workflows/ci.yml
blagoev: @desistefanova is this resolved. Why not name it Create cluster as the flutter-ios does. Just for consistency

blagoev: Oh I may be missing something. There are multiple of these resolved. Seems you have discussed this already. Feel free to ignore if that is the case

test/data/jwt_keys/README.md
```diff
@@ -0,0 +1,57 @@
+Custom JWT Authentication
+
+The keys in this folder "test/data/jwt_keys" are used only for the purpose of the automatic tests.
```
blagoev, line 3: Lets add these files to the .pubignore file. We will never need to publish them right?
Lets ignore the whole directory `jwt_keys`

blagoev: Same for the flutter package

blagoev: While here we should do this for the data/realm_files dir as well.

test/data/jwt_keys/private_key.pem
blagoev: If these are not needed lets remove them. Right now they have not polluted the main branch so if we delete them they will not get into the master history

blagoev (changes requested):
Its all good, but I have some small suggestions before merging this

test/data/jwt_keys/README.md
desistefanova: I deleted the whole folder jwt_keys and moved the REAMDME.md content into test\README.md.
About the realm file into data/realm_files we already have in .pubignore an exclude for the file `**/*.realm` but I will also add an exclude for the whole folder.

.github/workflows/flutter-desktop-tests.yml
```diff
@@ -70,7 +70,8 @@ jobs:
...
```
blagoev, line 73:
```suggestion
      - name: Create cluster
```

blagoev: just for consistency

blagoev (approved):

PR #723, opened 2022-07-22
Update README.md

papafe:
I think the `part` should have the same name as the generating class


nielsenko:
The part file is named by the file, not the generated class. You can define multiple models in the same file. Since the file is `app.dart` the part file is `app.g.dart`

papafe:
@nielsenko Yep, I misunderstood how it worked :)

PR #730, opened 2022-07-30
Open realm async

### LINKED ISSUE #71: Open: Support async Realm open

desistefanova:
Support async Realm open syntax. Fixes #71

### DIFF
--- a/test/realm_test.dart
+++ b/test/realm_test.dart
@@ -620,4 +620,23 @@ Future<void> main([List<String>? args]) async {
     // We should not be in transaction here
     expect(realm.isInTransaction, false);
   });
+
+  test('Realm open async with local configuration throws', () async {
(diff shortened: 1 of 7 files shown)


test/realm_test.dart
```diff
@@ -620,4 +620,23 @@ Future<void> main([List<String>? args]) async {
...
+    final app = App(appConfiguration);
+    final credentials = Credentials.anonymous();
+    final user = await app.logIn(credentials);
+    final configuration = Configuration.flexibleSync(user, [
+      Task.schema,
+      Schedule.schema,
+      Event.schema,
+    ]);
+
+    final realm = await Realm.open(configuration);
```
desistefanova, line 639: Realm.open is the async operation. Realm.openSync is the sync operation. Realm.openSync is the same as Realm(config) constructor. I've added a discussion to the round table. about the names of these methods.

PR #731, opened 2022-07-30
 Open realm async

### LINKED ISSUE #71: Open: Support async Realm open

desistefanova:
Support async Realm open syntax. Fixes #71 
Done: 
- [x] Implement Realm.open only in Dart without using C-API.
- [x] CancellableFuture implementation.
- [x] Update Changelog.md
- [x] API documentation
- [x] Add more tests for download progress.
- [x] Test opening many realms with different configs and try to cancel one of them.


test/realm_test.dart
```diff
@@ -620,4 +620,23 @@ Future<void> main([List<String>? args]) async {
...
+    final app = App(appConfiguration);
+    final credentials = Credentials.anonymous();
+    final user = await app.logIn(credentials);
+    final configuration = Configuration.flexibleSync(user, [
+      Task.schema,
+      Schedule.schema,
+      Event.schema,
+    ]);
+
+    final realm = await Realm.open(configuration);
```
desistefanova, line 639: Realm.open is the async operation. Realm.openSync is the sync operation. Realm.openSync is the same as Realm(config) constructor. I've added a discussion to the round table. about the names of these methods.

desistefanova: Decided on a meeting:

- Realm.open is the async. 

- Remove Realm.openSync. Use the constructor.

lib/src/realm_class.dart
```diff
@@ -95,16 +95,45 @@ class Realm implements Finalizable {
...
+  /// * `onProgressCallback` - a function that is registered as a callback for receiving download progress notifications. It is not mandatory.
+  ///
+  /// The returned task has an awaitable [RealmAsyncOpenTask.realm] future that is completed once the remote realm is fully synchronized.
+  /// It provides a method [RealmAsyncOpenTask.cancel] that cancels any current running download.
+  static RealmAsyncOpenTask open(Configuration config, {ProgressCallback? onProgressCallback}) {
+    _createFileDirectory(config.path);
+    RealmAsyncOpenTaskHandle realmAsyncOpenTaskHandle = realmCore.createRealmAsyncOpenTask(config);
+    final completer = Completer<RealmHandle?>();
+    Future<Realm?> realm = realmCore.openRealmAsync(realmAsyncOpenTaskHandle, completer).onError((error, stackTrace) {
+      print(error);
```
desistefanova, line 119: Return null only when the error is that operation is canceled. Otherwise we may have to return an exception.

lib/src/realm_class.dart
```diff
@@ -95,16 +95,45 @@ class Realm implements Finalizable {
...
+      return null;
```
desistefanova, line 120: Return null only when the error is that operation is canceled. Otherwise we may have to return an exception.

src/realm_dart_sync.cpp
```diff
@@ -138,3 +138,27 @@ RLM_API void realm_dart_sync_on_subscription_state_changed_callback(realm_userda
         (reinterpret_cast<realm_sync_on_subscription_state_changed_t>(ud->dart_callback))(ud->handle, state);
     });
 }
+
+RLM_API void realm_dart_async_open_task_completion_callback(realm_userdata_t userdata, realm_thread_safe_reference_t* realm, realm_async_error_t* error)
+{
+    //TODO: Read and copy the error with the correct message or code.
+    struct realm_dart_async_error : realm_async_error
+    {
```
desistefanova, line 146: How could I copy this error? It doesn't have error message.

nirinchev (changes requested):
Let's discuss the async open task API - I'd prefer if we work as much as possible with the system types and not expose additional stuff.

Ideally, I'd like users to be able to do `final realm = await Realm.open(config)`.

lib/src/realm_class.dart
```diff
@@ -95,16 +95,45 @@ class Realm implements Finalizable {
...
+  /// synchronized with the server prior to the completion of the returned [RealmAsyncOpenTask].
+  /// Otherwise this method will throw an exception.
+  ///
+  /// Open realm async arguments are:
+  /// * `config`- a configuration object that describes the realm.
+  /// * `onProgressCallback` - a function that is registered as a callback for receiving download progress notifications. It is not mandatory.
+  ///
+  /// The returned task has an awaitable [RealmAsyncOpenTask.realm] future that is completed once the remote realm is fully synchronized.
+  /// It provides a method [RealmAsyncOpenTask.cancel] that cancels any current running download.
+  static RealmAsyncOpenTask open(Configuration config, {ProgressCallback? onProgressCallback}) {
```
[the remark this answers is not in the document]
desistefanova: Added a discussion point.

lib/src/native/realm_core.dart
```diff
@@ -1680,6 +1680,62 @@ class _RealmCore {
         "Delete user failed");
     return completer.future;
   }
+
+  static void _openRealmAsyncCallback(Handle userdata, Pointer<realm_thread_safe_reference_t> realm, Pointer<realm_async_error_t> errorCode) {
+    final completer = userdata as Completer<RealmHandle?>;
+    if (errorCode != nullptr) {
+      {
+        completer.completeError(RealmException(errorCode.toString()));
+      }
```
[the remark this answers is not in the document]
desistefanova: Yes, it is a mistake, maybe that's why in one of the tests I don't received the error text as it is.

lib/src/native/realm_core.dart
```diff
@@ -1680,6 +1680,62 @@ class _RealmCore {
...
+    } else {
+      final realmPtr =
+          _realmLib.invokeGetPointer(() => _realmLib.realm_from_thread_safe_reference(realm, scheduler.handle._pointer), "Error opening synchronized realm");
+      if (completer.isCompleted) {
+        //Task is canceled before to complete but the realm was created.
+        _realmLib.invokeGetBool(() => _realmLib.realm_close(realmPtr), "Realm close failed");
```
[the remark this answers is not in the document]
desistefanova: I don't actually understand your suggestion. Can you elaborate more?

nielsenko:
I have tried to advocate that we add this functionality without involving core at all.

Here are different styles this can be implemented in pure dart:
```dart
  static Future<Realm> open(FlexibleSyncConfiguration config, [void Function(int, int)? progressCallback]) async {
    final realm = Realm(config);
    final session = realm.syncSession;
    await session
        .getProgressStream(ProgressDirection.download, ProgressMode.forCurrentlyOutstandingWork)
        .forEach((s) => progressCallback?.call(s.transferredBytes, s.transferableBytes));
    await session.waitForDownload(); // should be redundant except for bug in progress stream
    return realm;
  }

  static CancelableOperation<Realm> openReturnsCancelableOperation(FlexibleSyncConfiguration config, [void Function(int, int)? progressCallback]) =>
      CancelableOperation.fromFuture(open(config));

  static Future<Realm> openDotnetStyle(FlexibleSyncConfiguration config, {void Function(int, int)? progressCallback, Completer<void>? cancelToken}) =>
      StreamGroup.merge([
        open(config).asStream(),
        if (cancelToken != null) cancelToken.future.asStream().cast<Realm>(), // only used to cancel
      ]).first;
```
Personally I prefer the first. If you don't need to cancel, you just do:
```dart
final realm = await Realm.open(config)
```
if you just need a timeout, you do:
```dart
final realm = await Realm.open(config).timeout(duration);
```
If you need to be able to cancel, you do:
```dart
final op = CancelableOperation.fromFuture(Realm.open(config));
```
You can say, that won't really cancel anything in core, but neither does the current PR, since download will proceed in the background no matter what, and this is very much by design.

Also, the open async api in core is buggy, in that If you cancel **one** operation you cancel **all** in-flight ops. I think the above this is much simpler. Obviously we should only support one of the 3, but I include them all to show the options.

lib/src/realm_class.dart
```diff
@@ -104,16 +105,62 @@ class Realm implements Finalizable {
...
+  /// * `cancelationController` - an initialized object of [RealmCancelationController] that is used to cancel the operation. It is not mandatory.
+  /// * `onProgressCallback` - a function that is registered as a callback for receiving download progress notifications. It is not mandatory.
+  ///
+  /// The returned type is [Future<Realm?>] that is completed once the remote realm is fully synchronized or canceled.
+  /// [RealmCancelationController] provides method [cancel] that cancels any current running download.
+  /// If multiple [Realm.open] operations are all in the progress for the same Realm,
+  /// then canceling one of them will cancel all of them.
+  static Future<Realm> open(Configuration config, {RealmCancellationToken? cancellationToken, ProgressCallback? onProgressCallback}) async {
+    CancelledException? exception;
+    _createFileDirectory(config.path);
```
nielsenko, line 129: This is redundant I believe. Don't Realm ctor already handle this?

lib/src/realm_class.dart
```diff
@@ -104,16 +105,62 @@ class Realm implements Finalizable {
...
+      }
+    } on CancelledException catch (e) {
+      exception = e;
+      realm.close();
+    } catch (e) {
+      rethrow;
+    }
+    if (cancellationToken?.token.isCancelled == true && exception != null) {
+      throw exception;
+    }
```
nielsenko, line 150:
```suggestion
    } on CancelledException catch (_) {
      realm.close();
      rethrow;
    }
```

lib/src/realm_class.dart
```diff
@@ -104,16 +105,62 @@ class Realm implements Finalizable {
...
+  static RealmHandle _openRealmSync(Configuration config) {
+    _createFileDirectory(config.path);
+    return realmCore.openRealm(config);
+  }
+
+  static void _createFileDirectory(String filePath) {
+    var dir = File(filePath).parent;
     if (!dir.existsSync()) {
       dir.createSync(recursive: true);
     }
```
nielsenko, line 163:
```suggestion
```

lib/src/realm_class.dart
nielsenko, line 129:
```suggestion
```

lib/src/realm_class.dart
```diff
@@ -104,16 +105,62 @@ class Realm implements Finalizable {
...
-    return realmCore.openRealm(config);
   }
```
nielsenko, line 164:
```suggestion
```

test/realm_test.dart
```diff
@@ -715,6 +714,143 @@ Future<void> main([List<String>? args]) async {
...
+    final user = await app.logIn(credentials);
+    final configuration = Configuration.flexibleSync(user, [Task.schema]);
+
+    var cancellationToken = RealmCancellationToken();
+    final realm = Realm.open(configuration, cancellationToken: cancellationToken);
+    cancellationToken.cancel();
+    expect(() async => await realm, throws<RealmCancelledException>());
+  });
+
+  // baasTest('Realm open async with the same cancelationController throws', (appConfiguration) async {
```
nielsenko, line 782: I suggest we remove this dead test code, or rewire it so it works.

nielsenko (changes requested):

test/realm_test.dart
```diff
@@ -715,6 +714,169 @@ Future<void> main([List<String>? args]) async {
...
+    cancellationToken.cancel();
+    expect(() async => await realm, throws<CancelledException>());
+  });
+
+  test('Realm open async and cancel for local configuration', () async {
+    final configuration = Configuration.local([Car.schema]);
+    var cancellationToken = CancellationToken();
+    final realm = Realm.open(configuration, cancellationToken: cancellationToken);
+    cancellationToken.cancel();
+    var openedRealm = await realm; // You can't cancel local Realm
```
desistefanova, line 794: Do you think it's okay to have different behavior for local realms? The Realm is returned, although the opening is cancelled.

lib/src/realm_class.dart
```diff
@@ -108,12 +108,49 @@ class Realm implements Finalizable {
...
     _populateMetadata();
     isFrozen = realmCore.isFrozen(this);
   }
 
-  static RealmHandle _openRealm(Configuration config) {
+  /// A method for asynchronously obtaining and opening a [Realm].
+  ///
+  /// If the configuration is [FlexibleSyncConfiguration], the realm will be downloaded and fully
+  /// synchronized with the server prior to the completion of the returned [Future].
+  /// Otherwise this method will throw an exception.
```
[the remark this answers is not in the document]
desistefanova: Yes, I will change it.

lib/src/realm_class.dart
```diff
@@ -615,3 +652,88 @@ class DynamicRealm {
...
+///
+/// It is called with the following arguments:
+/// * `transferredBytes` - the current number of bytes already transferred
+/// * `totalBytes` - the total number of transferable bytes (the number of bytes already transferred plus the number of bytes pending transfer)
+/// {@category Realm}
+typedef ProgressCallback = void Function(int transferredBytes, int totalBytes);
+
+/// An exception being thrown when a cancellable operation is cancelled by calling [CancellationToken.cancel].
+/// {@category Realm}
+class CancelledException implements Exception {
```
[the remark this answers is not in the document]
desistefanova: OK

lib/src/realm_class.dart
```diff
@@ -106,15 +106,58 @@ class Realm implements Finalizable {
...
+      final session = realm.syncSession;
+      if (onProgressCallback != null) {
+        await session
+            .getProgressStream(ProgressDirection.download, ProgressMode.forCurrentlyOutstandingWork)
+            .forEach((s) => onProgressCallback.call(s.transferredBytes, s.transferableBytes));
+      }
+      await session.waitForDownload();
+      if (!realm._openedFirstTime) {
+        await session.waitForUpload();
+      }
```
desistefanova, line 155: I'm not sure that `open` has to do two way synchronization. @nirinchev , @blagoev , @nielsenko ?

lib/src/realm_class.dart
```diff
@@ -634,3 +677,88 @@ class MigrationRealm extends DynamicRealm {
...
+///
+/// fromFutureFunction arguments are:
+/// * `futureFunction`- a function executung a Future that has to be canceled.
+/// * `cancellationToken` - [CancellationToken] that is used to cancel the Future.
+/// * `cancelledMessage` - am optional argument providing reasonable message of [CancelledException] thrown by the Future if the token is canceled.
+/// {@category Realm}
+class CancellableFuture {
+  static Future<T> fromFutureFunction<T>(Future<T> Function() futureFunction, CancellationToken? cancellationToken, {String? cancelledMessage}) async {
+    if (cancellationToken != null) {
+      final cancelException = CancelledException(cancelledMessage ?? "CancellableFuture was canceled.");
```
desistefanova, line 746:
```suggestion
      final cancelException = CancelledException(cancelledMessage ?? "Canceled operation.");
```

lib/src/realm_class.dart
nielsenko: In my understanding download should be enough

nielsenko (commented):
I have created a separate branch with my suggested changes: #916

lib/src/realm_class.dart
desistefanova: What if it is existing realm that has been modified offline?

nirinchev, line 155: Why are we doing this?

lib/src/realm_class.dart
```diff
@@ -106,15 +106,58 @@ class Realm implements Finalizable {
...
+  ///
+  /// * `config`- a configuration object that describes the realm.
+  /// * `cancellationToken` - an optional [CancellationToken] used to cancel the operation.
+  /// * `onProgressCallback` - a callback for receiving download progress notifications for synced [Realm]s.
+  ///
+  /// Returns [Future<Realm>] that completes with the `realm` once the remote realm is fully synchronized or with an `error` if operation is canceled.
+  /// When the configuration is [LocalConfiguration] this completes right after the local realm is opened or operation is canceled.
+  static Future<Realm> open(Configuration config, {CancellationToken? cancellationToken, ProgressCallback? onProgressCallback}) async {
+    Realm realm = await CancellableFuture.fromFutureFunction<Realm>(() => _open(config, cancellationToken, onProgressCallback), cancellationToken);
+    return realm;
```
nirinchev, line 136: Why not return immediately?

```suggestion
    return CancellableFuture.fromFutureFunction<Realm>(() => _open(config, cancellationToken, onProgressCallback), cancellationToken);
```

lib/src/realm_class.dart
```diff
@@ -106,15 +106,58 @@ class Realm implements Finalizable {
...
+    });
+
+    if (config is FlexibleSyncConfiguration) {
+      final session = realm.syncSession;
+      if (onProgressCallback != null) {
+        await session
+            .getProgressStream(ProgressDirection.download, ProgressMode.forCurrentlyOutstandingWork)
+            .forEach((s) => onProgressCallback.call(s.transferredBytes, s.transferableBytes));
+      }
+      await session.waitForDownload();
```
nirinchev, line 152: These two are mutually exclusive. Instead I would suggest something like:

```dart
StreamSubscription<SyncProgress>? subscription;
if (progressCallback != null) {
  final subscription = session.getProgressStream(...).listen((s) => onProgressCallback.call(...);

  cancellationToken?.onBeforeCancel(() async {
    await subscription.cancel();
  });

  subscription.onDone(() {
    // Just nullify it to release it early
    subscription = null;
  });
}

await session.waitForDownload();
await subscription?.cancel();
```

lib/src/realm_class.dart
```diff
@@ -634,3 +677,88 @@ class MigrationRealm extends DynamicRealm {
...
 }
+
+/// The signature of a callback that will be executed while the Realm is opened asynchronously with [Realm.open].
+/// This is the registered callback onProgressCallback to receive progress notifications while the download is in progress.
+///
+/// It is called with the following arguments:
+/// * `transferredBytes` - the current number of bytes already transferred
+/// * `totalBytes` - the total number of transferable bytes (the number of bytes already transferred plus the number of bytes pending transfer)
+/// {@category Realm}
+typedef ProgressCallback = void Function(int transferredBytes, int totalBytes);
```
nirinchev, line 688: Why not have it be a `Function(SyncProgress progress)`?

lib/src/realm_class.dart
nielsenko: Those changes are still available locally. Async open should ensure we are ready to _go_ locally, not ensure everything is uploaded

desistefanova: Can you check 2 comments above. I was wondering whether we need to wait for upload in case of having existing realm that has been modified offline and then opened async?

lib/src/realm_class.dart
desistefanova: I tryed something similar but before to implement `onBeforeCancel` and I didn't have a way to handle the exception and to cancel the listener. Now after having onBeforeCancel it make sense to have the cancellation of the listening.

lib/src/realm_class.dart
nirinchev: Agree, we should not be waiting for upload as that will happen in the background and should not delay the user from interacting with the app.

nirinchev: Ah missed that. replied in the correct thread.

desistefanova: OK, thanks! Then I will remove it.

lib/src/realm_class.dart
desistefanova: Hm, if I do this I will have to make `token.cancel()` to be a Future in order to be possible to `await `inside 
```dart
onBeforeCancel(() async {
    await subscription.cancel();
  });
```

nirinchev: Ah, good point - let's not make it a promise though, that'd be annoying. We can instead call something like `subscription.cancel().catchError((e) { log(e)})`. We don't really care about errors that occur while canceling the subscription, so don't need to await the cancelation.

lib/src/realm_class.dart
```diff
@@ -33,6 +34,7 @@ import 'scheduler.dart';
 import 'subscription.dart';
 import 'session.dart';
 
+export 'package:cancellation_token/cancellation_token.dart' show CancellationToken, CancelledException, CancellableCompleter, CancellableFuture;
```
nirinchev, line 37:
```suggestion
export 'package:cancellation_token/cancellation_token.dart' show CancellationToken, CancelledException;
```

Do we need to expose those extra types?

lib/src/realm_class.dart
```diff
@@ -110,12 +112,69 @@ class Realm implements Finalizable {
...
     _populateMetadata();
     isFrozen = realmCore.isFrozen(this);
   }
 
-  static RealmHandle _openRealm(Configuration config) {
+  /// A method for asynchronously opening a [Realm].
+  ///
+  /// When the configuration is [FlexibleSyncConfiguration], the realm will be downloaded and fully
+  /// synchronized with the server prior to the completion of the returned [Future].
+  /// This method could be called also for opening a local [Realm] with [LocalConfiguration].
```
nirinchev, line 124: We should clarify that there's no benefit of opening a Realm async vs sync.

lib/src/realm_class.dart
```diff
@@ -110,12 +112,69 @@ class Realm implements Finalizable {
...
+  /// * `config`- a configuration object that describes the realm.
+  /// * `cancellationToken` - an optional [CancellationToken] used to cancel the operation.
+  /// * `onProgressCallback` - a callback for receiving download progress notifications for synced [Realm]s.
+  ///
+  /// Returns `Future<Realm>` that completes with the [Realm] once the remote [Realm] is fully synchronized or with a [CancelledException] if operation is canceled.
+  /// When the configuration is [LocalConfiguration] this completes right after the local [Realm] is opened or operation is canceled.
+  static Future<Realm> open(Configuration config, {CancellationToken? cancellationToken, ProgressCallback? onProgressCallback}) async {
+    final cancellableCompleter = CancellableCompleter<Realm>(cancellationToken);
+    try {
+      Realm realm = Realm(config);
```
nirinchev, line 135: nit

```suggestion
      final realm = Realm(config);
```

lib/src/realm_class.dart
```diff
@@ -110,12 +112,69 @@ class Realm implements Finalizable {
...
+  /// This method could be called also for opening a local [Realm] with [LocalConfiguration].
+  ///
+  /// * `config`- a configuration object that describes the realm.
+  /// * `cancellationToken` - an optional [CancellationToken] used to cancel the operation.
+  /// * `onProgressCallback` - a callback for receiving download progress notifications for synced [Realm]s.
+  ///
+  /// Returns `Future<Realm>` that completes with the [Realm] once the remote [Realm] is fully synchronized or with a [CancelledException] if operation is canceled.
+  /// When the configuration is [LocalConfiguration] this completes right after the local [Realm] is opened or operation is canceled.
+  static Future<Realm> open(Configuration config, {CancellationToken? cancellationToken, ProgressCallback? onProgressCallback}) async {
+    final cancellableCompleter = CancellableCompleter<Realm>(cancellationToken);
```
nirinchev, line 133: Why do we need this completer? Seems like we can completely remove it since we're already piping the cancellation token to the async methods.

lib/src/realm_class.dart
```diff
@@ -110,12 +112,69 @@ class Realm implements Finalizable {
...
+        await session.waitForDownload(cancellationToken);
+      }
+      cancellableCompleter.complete(realm);
+    } catch (error) {
+      cancellableCompleter.completeError(error);
+    }
+    return cancellableCompleter.future;
+  }
+
+  static Future<void> _syncProgressNotifier(Session session, ProgressCallback onProgressCallback, [CancellationToken? cancellationToken]) async {
```
nirinchev, line 155: Let's remove this method and instead add something like this to `open`:

```dart
if (config is FlexibleSyncConfiguration) {
  final session = realm.syncSession;
  StreamSubscription<SyncProgress>? subscription;
  try {
    if (onProgressCallback != null) {
      subscritpion = session.getProgressStream(ProgressDirection.download, ProgressMode.forCurrentlyOutstandingWork)
        .listen(
           (syncProgress) => onProgressCallback!.call(syncProgress),
           cancelOnError: true,
         );
    }
    await session.waitForDownload(cancellationToken);
  } finally {
    await subscription?.cancel();
  }
}
```

test/test.dart
```diff
@@ -571,3 +571,16 @@ Future<void> _printPlatformInfo() async {
 
   print('Current PID $pid; OS $os, $pointerSize bit, CPU ${cpu ?? 'unknown'}');
 }
+
+class RealmA {
```
nirinchev, line 575: Should we remove this now? I would suggest we move it to a method similar to `getRealm` - e.g. `getRealmAsync` rather than this custom class.

nirinchev (changes requested):
I haven't reviewed the tests thoroughly, but the rest looks fine - I have posted a couple of ideas how to streamline things a little.

lib/src/realm_class.dart
desistefanova: @nielsenko , @blagoev What do you think? Shall we expose `CancellableCompleter` and `CancellableFuture` or not?

lib/src/realm_class.dart
desistefanova: I have added some comments describing the reason why I have implemented this completer.

lib/src/realm_class.dart
nielsenko: Personally I don't think we should export classes from other packages. I think they should pull in the package themselves if they wan't to use it.

lib/src/realm_class.dart
nielsenko: I have updated #916 to show how to avoid it.

nielsenko (changes requested):
I have created two branches for my suggested changes:

1. #916 is a simplification of `Realm.open`.
2 #960 is some suggested changes to the tests regarding `_cancelOrRealm`.

They are independent of each other.

lib/src/realm_class.dart
```diff
@@ -656,3 +702,17 @@ class MigrationRealm extends DynamicRealm {
...
+
+/// The signature of a callback that will be executed while the Realm is opened asynchronously with [Realm.open].
+/// This is the registered callback onProgressCallback to receive progress notifications while the download is in progress.
+///
+/// * syncProgress - an object of [SyncProgress] that contains `transferredBytes` and `transferableBytes`.
+/// {@category Realm}
+typedef ProgressCallback = void Function(SyncProgress syncProgress);
+
+/// @nodoc
+extension $CancellableCompleterExtension<T> on Completer<T> {
```
desistefanova, line 714: This is not currently used, but I think it is nice to have it in order to convert `Completer` to `CancellableCompleter`.

lib/src/realm_class.dart
desistefanova: This method is rewritten according to your suggestions, please check it.

desistefanova:
> @nielsenko Could you please check the changes now?

lib/src/realm_class.dart
```diff
@@ -113,12 +115,56 @@ class Realm implements Finalizable {
...
+        await session.waitForDownload(cancellationToken);
+      }
+    } catch (error) {
+      realm.close();
+      rethrow;
+    } finally {
+      await subscription?.cancel();
+    }
+    return realm;
+  }
```
nielsenko, line 165:
```suggestion
  static Future<Realm> open(Configuration config, {CancellationToken? cancellationToken, ProgressCallback? onProgressCallback}) async {
    final realm = Realm(config);
    try {
      if (config is FlexibleSyncConfiguration) {
        final session = realm.syncSession;
        await Future.wait<void>([
          session.waitForDownload(cancellationToken),
          if (onProgressCallback != null)
            session
                .getProgressStream(
                  ProgressDirection.download,
                  ProgressMode.forCurrentlyOutstandingWork,
                )
                .forEach(onProgressCallback)
                .asCancellable(cancellationToken),
        ]);
      }
      // handle cancellation even if we are not using flexible sync
      return await CancellableFuture.value(realm, cancellationToken);
    } catch (_) {
      realm.close();
      rethrow;
    }
  }
```

lib/src/realm_class.dart
```diff
@@ -656,3 +702,17 @@ class MigrationRealm extends DynamicRealm {
...
+///
+/// * syncProgress - an object of [SyncProgress] that contains `transferredBytes` and `transferableBytes`.
+/// {@category Realm}
+typedef ProgressCallback = void Function(SyncProgress syncProgress);
+
+/// @nodoc
+extension $CancellableCompleterExtension<T> on Completer<T> {
+  CancellableCompleter<T> asCancellable(CancellationToken? cancellationToken, {OnCancelCallback? onCancel}) {
+    return CancellableCompleter<T>(cancellationToken, onCancel: onCancel);
+  }
```
nielsenko, line 717: This is not really an extension method, as no state from the original completer is used. I think we should drop it
```suggestion
```

test/realm_test.dart
```diff
@@ -829,6 +829,172 @@ Future<void> main([List<String>? args]) async {
...
+  baasTest('Realm.open (flexibleSync) - cancel right after open', (appConfiguration) async {
+    final app = App(appConfiguration);
+    final credentials = Credentials.anonymous();
+    final user = await app.logIn(credentials);
+    final configuration = Configuration.flexibleSync(user, [Task.schema]);
+
+    var cancellationToken = CancellationToken();
+    final isRealmCancelled = getRealmAsync(configuration, cancellationToken: cancellationToken).thenIsCancelled();
+    cancellationToken.cancel();
+    await expectLater(await isRealmCancelled, isTrue);
```
nielsenko, line 876:
```suggestion
    expect(await isRealmCancelled, isTrue);
```

test/realm_test.dart
```diff
@@ -829,6 +829,172 @@ Future<void> main([List<String>? args]) async {
...
+    final user = await app.logIn(credentials);
+    final configuration = Configuration.flexibleSync(user, [Task.schema]);
+
+    var cancellationToken = CancellationToken();
+
+    Future<void>.delayed(Duration(milliseconds: 10), () => cancellationToken.cancel());
+    final isRealm1Cancelled = getRealmAsync(configuration, cancellationToken: cancellationToken).thenIsCancelled();
+    final isRealm2Cancelled = getRealmAsync(configuration, cancellationToken: cancellationToken).thenIsCancelled();
+    await expectLater(await isRealm1Cancelled, isTrue);
+    await expectLater(await isRealm2Cancelled, isTrue);
```
nielsenko, line 891:
```suggestion
    expect(await isRealm1Cancelled, isTrue);
    expect(await isRealm2Cancelled, isTrue);
```

test/realm_test.dart
```diff
@@ -829,6 +829,172 @@ Future<void> main([List<String>? args]) async {
...
+    final configuration = Configuration.flexibleSync(user, [Task.schema]);
+
+    var cancellationToken1 = CancellationToken();
+    final isRealm1Cancelled = getRealmAsync(configuration, cancellationToken: cancellationToken1).thenIsCancelled();
+
+    var cancellationToken2 = CancellationToken();
+    final isRealm2Cancelled = getRealmAsync(configuration, cancellationToken: cancellationToken2).thenIsCancelled();
+    cancellationToken1.cancel();
+    await expectLater(await isRealm1Cancelled, isTrue);
+    await expectLater(await isRealm2Cancelled, isFalse);
```
nielsenko, line 907:
```suggestion
    expect(await isRealm1Cancelled, isTrue);
    expect(await isRealm2Cancelled, isFalse);
```

test/realm_test.dart
```diff
@@ -829,6 +829,172 @@ Future<void> main([List<String>? args]) async {
...
+    final isRealm1Cancelled = getRealmAsync(configuration1, cancellationToken: cancellationToken1).thenIsCancelled();
+
+    final user2 = await app.logIn(Credentials.anonymous(reuseCredentials: false));
+    final configuration2 = Configuration.flexibleSync(user2, [Task.schema]);
+    var cancellationToken2 = CancellationToken();
+    final isRealm2Cancelled = getRealmAsync(configuration2, cancellationToken: cancellationToken2).thenIsCancelled();
+
+    cancellationToken2.cancel();
+    await expectLater(await isRealm2Cancelled, isTrue);
+    await expectLater(await isRealm1Cancelled, isFalse);
```
nielsenko, line 925:
```suggestion
    expect(await isRealm2Cancelled, isTrue);
    expect(await isRealm1Cancelled, isFalse);
```

test/realm_test.dart
```diff
@@ -829,6 +829,172 @@ Future<void> main([List<String>? args]) async {
...
+  baasTest('Realm.open (flexibleSync) - listen and cancel download progress of a populated realm', (appConfiguration) async {
+    final app = App(appConfiguration);
+    final config = await _addDataToAtlas(app);
+
+    var cancellationToken = CancellationToken();
+    final realmIsCancelled = getRealmAsync(config, cancellationToken: cancellationToken, onProgressCallback: (syncProgress) {
+      print("PROGRESS: transferredBytes: ${syncProgress.transferredBytes}, totalBytes:${syncProgress.transferableBytes}");
+    }).thenIsCancelled();
+    cancellationToken.cancel();
+    await expectLater(await realmIsCancelled, isTrue);
```
nielsenko, line 995:
```suggestion
    expect(await realmIsCancelled, isTrue);
```

test/realm_test.dart
```diff
@@ -853,6 +1019,57 @@ void openEncryptedRealm(List<int>? encryptionKey, List<int>? decryptionKey, {voi
...
+        print("REALM CHECK: Cancelled before to return the Realm.");
+        return true;
+      },
+      test: (error) => error is CancelledException,
+    ).catchError((Object error) {
+      print("REALM CHECK: Unhandled exception.");
+      throw error;
+    }, test: (error) => error is! CancelledException);
+  }
+}
```
nielsenko, line 1040:
```suggestion
extension _FutureRealm on Future<Realm> {
  Future<bool> thenIsCancelled() async {
    try {
      final value = await this;
      expect(value, isNotNull);
      expect(value.isClosed, false);
      print("REALM CHECK: Realm returned before cancellation.");
      return false;
    } on CancelledException {
      print("REALM CHECK: Cancelled before to return the Realm.");
      return true;
    } catch (_) {
      print("REALM CHECK: Unhandled exception.");
      rethrow;
    }
  }
}
```
Perhaps read up on https://dart.dev/guides/language/effective-dart/usage#prefer-asyncawait-over-using-raw-futures

Also, do we need the print statements?

nielsenko (changes requested):
Looking better every day.. as usual I have gathered my PR comments in a branch. You can check it out here: #916, but this time I also added them as individual suggestions in the web UI.

test/realm_test.dart
desistefanova: The ting is that I want just to attach `.then` and `.onError` here without to wait for the execution to complete. The waiting is managed in the tests accorting to the case. This extension only attach the checks that to be done when the execution completes. That's why its name is `.thenIsCancelled`. It is similar to` .then` but with a check for cancellation.

nielsenko: I'm okay with the name, but I suggest we follow best practices from [effective dart](https://dart.dev/guides/language/effective-dart/usage#prefer-asyncawait-over-using-raw-futures):
> PREFER async/await over using raw futures.
Asynchronous code is notoriously hard to read and debug, even when using a nice abstraction like futures. The async/await syntax improves readability and lets you use all of the Dart control flow structures within your async code.

lib/src/realm_class.dart
desistefanova: Removed

nielsenko (approved):
LGTM 👍

lib/src/realm_class.dart
```diff
@@ -113,12 +115,53 @@ class Realm implements Finalizable {
...
+  /// * `cancellationToken` - an optional [CancellationToken] used to cancel the operation.
+  /// * `onProgressCallback` - a callback for receiving download progress notifications for synced [Realm]s.
+  ///
+  /// Returns `Future<Realm>` that completes with the [Realm] once the remote [Realm] is fully synchronized or with a [CancelledException] if operation is canceled.
+  /// When the configuration is [LocalConfiguration] this completes right after the local [Realm] is opened or if the operation is canceled in advance.
+  /// Since opening a local Realm is a synchronous operation, there is no benefit of using Realm.open over the constructor.
+  static Future<Realm> open(Configuration config, {CancellationToken? cancellationToken, ProgressCallback? onProgressCallback}) async {
+    if (cancellationToken?.isCancelled ?? false) {
+	      return Future<Realm>.error(cancellationToken!.exception);
+	    }
```
nielsenko, line 139: One last thing :-) ..
```suggestion
    if (cancellationToken != null && cancellationToken.isCancelled) {
      throw cancellationToken.exception;
    }
```
since we are in a `async` method, we can just throw (a bit shorter). Also by testing for `!= null` the compiler can infer we don't need the bangs (`!`) later

test/realm_test.dart
```diff
@@ -829,6 +829,172 @@ Future<void> main([List<String>? args]) async {
...
+
+    expect(syncedRealm.isClosed, false);
+    expect(printCount, isNot(0));
+  });
+
+  baasTest('Realm.open (flexibleSync) - download a populated realm', (appConfiguration) async {
+    final app = App(appConfiguration);
+    final config = await _addDataToAtlas(app);
+    var syncedRealm = await getRealmAsync(config);
+    expect(syncedRealm.isClosed, false);
```
nirinchev, line 965: This should verify the contents of the Realm, otherwise there's no guarantee that `Realm.open` actually downloads the latest state.

test/realm_test.dart
```diff
@@ -829,6 +829,172 @@ Future<void> main([List<String>? args]) async {
...
+
+    var syncedRealm = await getRealmAsync(config, onProgressCallback: (syncProgress) {
+      print("PROGRESS: transferredBytes: ${syncProgress.transferredBytes}, totalBytes:${syncProgress.transferableBytes}");
+      printCount++;
+      transferredBytes = syncProgress.transferredBytes;
+    });
+
+    expect(syncedRealm.isClosed, false);
+    expect(printCount, isNot(0));
+    expect(transferredBytes > 19, isTrue); //19 bytes is the empty realm
```
nirinchev, line 983:
```suggestion
    expect(transferredBytes, greaterThan(19)); //19 bytes is the empty realm
```

test/realm_test.dart
```diff
@@ -829,6 +829,172 @@ Future<void> main([List<String>? args]) async {
...
+  });
+
+  baasTest('Realm.open (flexibleSync) - listen for download progress of a populated realm', (appConfiguration) async {
+    final app = App(appConfiguration);
+    final config = await _addDataToAtlas(app);
+
+    int printCount = 0;
+    int transferredBytes = 0;
+
+    var syncedRealm = await getRealmAsync(config, onProgressCallback: (syncProgress) {
```
nirinchev, line 975:
```suggestion
    final syncedRealm = await getRealmAsync(config, onProgressCallback: (syncProgress) {
```

test/realm_test.dart
```diff
@@ -829,6 +829,172 @@ Future<void> main([List<String>? args]) async {
...
+      print("PROGRESS: transferredBytes: ${syncProgress.transferredBytes}, totalBytes:${syncProgress.transferableBytes}");
```
nirinchev, line 976: Can we remove the print here?

test/realm_test.dart
```diff
@@ -829,6 +829,172 @@ Future<void> main([List<String>? args]) async {
...
+  });
+
+  baasTest('Realm.open (flexibleSync) - listen and cancel download progress of a populated realm', (appConfiguration) async {
+    final app = App(appConfiguration);
+    final config = await _addDataToAtlas(app);
+
+    var cancellationToken = CancellationToken();
+    final realmIsCancelled = getRealmAsync(config, cancellationToken: cancellationToken, onProgressCallback: (syncProgress) {
+      print("PROGRESS: transferredBytes: ${syncProgress.transferredBytes}, totalBytes:${syncProgress.transferableBytes}");
+    }).thenIsCancelled();
```
nirinchev, line 993: Should we verify something about the download progress?

nirinchev (changes requested):

lib/src/realm_class.dart
```diff
@@ -113,12 +115,53 @@ class Realm implements Finalizable {
...
+        await Future.wait<void>([
+          session.waitForDownload(cancellationToken),
+          if (onProgressCallback != null)
+            session
+                .getProgressStream(
+                  ProgressDirection.download,
+                  ProgressMode.forCurrentlyOutstandingWork,
+                )
+                .forEach(onProgressCallback)
+                .asCancellable(cancellationToken),
```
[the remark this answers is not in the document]
nielsenko: Why?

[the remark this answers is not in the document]
nielsenko: This is cancelled correctly as evident from this small test
```dart
  test('example of cancelling an infinite stream with asCancellable', () async {
    final ct = TimeoutCancellationToken(Duration(seconds: 1), timeoutException: CancelledException());
    await expectLater(
      Stream.periodic(Duration(milliseconds: 100), (i) => i) //
          .forEach((i) => print(i))
          .asCancellable(ct),
      throwsA(isA<CancelledException>()),
    );
  });
```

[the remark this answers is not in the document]
nirinchev: I just don't like that we're waiting on two things that are supposed to do the same and one of them is known to be broken. The authoritative indication that we're done waiting should be `session.waitForDownload`. We can totally use the progress stream to report progress, but should not treat it completing is a condition to wait on.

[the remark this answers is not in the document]
nielsenko: I think it is important that we guarantee, that if you add a `onProgressCallback` you will see `transferredBytes == transferableBytes` before returning from `Realm.open`.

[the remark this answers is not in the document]
nirinchev: Core is not closing the stream - we are. And we do it when transferredBytes exceeds transferableBytes - we don't have strict guarantees when that will happen. I don't think it's particularly important that you see those values match - if you think about the most common use case of progress notifications, it's to show a loader to the user. Whether in the last millisecond they see the loader go to 100% or it hides at 95% and you're taken to the UI immediately should be irrelevant.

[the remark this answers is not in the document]
nielsenko: Still I think we should make an effort to make them match. Mostly for the developers using our SDK.

lib/src/realm_class.dart
desistefanova: We don't throw because the exceprion occurs when Realm.open is called instead of after await the future returned by Realo.open.

nielsenko: The two pieces of code does the exact same thing? .. the methods is marked `async`

desistefanova: Right, my previous test was wrong.

desistefanova: It is applied.

lib/src/realm_class.dart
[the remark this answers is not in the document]
desistefanova: Added a discussion point to the round table meeting.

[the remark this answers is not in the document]
desistefanova: On a discussion it was decided the method with cancelling subscription to be returned back. Here is the final code:
```dart
static Future<Realm> open(Configuration config, {CancellationToken? cancellationToken, ProgressCallback? onProgressCallback}) async {
    if (cancellationToken != null && cancellationToken.isCancelled) {
      throw cancellationToken.exception;
    }
    final realm = Realm(config);
    StreamSubscription<SyncProgress>? subscription;
    try {
      if (config is FlexibleSyncConfiguration) {
        final session = realm.syncSession;
        if (onProgressCallback != null) {
          subscription = session
              .getProgressStream(
                ProgressDirection.download,
                ProgressMode.forCurrentlyOutstandingWork,
              )
              .listen(
                (syncProgress) => onProgressCallback.call(syncProgress),
                cancelOnError: true,
              );
        }
        await session.waitForDownload(cancellationToken);
      }
    } catch (_) {
      realm.close();
      rethrow;
    } finally {
      await subscription?.cancel();
    }
    return await CancellableFuture.value(realm, cancellationToken);
  }
```

lib/src/realm_class.dart
```diff
@@ -113,12 +115,56 @@ class Realm implements Finalizable {
...
+                cancelOnError: true,
+              );
+        }
+        await session.waitForDownload(cancellationToken);
+      }
+    } catch (_) {
+      realm.close();
+      rethrow;
+    } finally {
+      await subscription?.cancel();
```
nielsenko, line 162: I would like this to happen before `realm.close()`, but it doesn't seem to be a problem with current core.

nielsenko (approved):
I'm not a fan of replacing `Future.wait` here, but I _reluctantly accept_  🥲 

I have added a few suggestions..

lib/src/realm_class.dart
```diff
@@ -113,12 +115,56 @@ class Realm implements Finalizable {
...
+        if (onProgressCallback != null) {
+          subscription = session
+              .getProgressStream(
+                ProgressDirection.download,
+                ProgressMode.forCurrentlyOutstandingWork,
+              )
+              .listen(
+                (syncProgress) => onProgressCallback.call(syncProgress),
+                cancelOnError: true,
+              );
```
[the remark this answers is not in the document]
nielsenko: The `(syncProgress) => onProgressCallback.call(syncProgress)` is just redundant, and we don't really need `cancelOnError: true` here as we call cancel in the finally clause.

lib/src/realm_class.dart
desistefanova: Ok I change it.

test/realm_test.dart
```diff
@@ -1109,6 +1278,50 @@ void openEncryptedRealm(List<int>? encryptionKey, List<int>? decryptionKey, {voi
   }
 }
 
+extension _FutureRealm on Future<Realm> {
+  Future<bool> thenIsCancelled() async {
```
blagoev, line 1282:
```suggestion
  Future<bool> getIsCancelled() async {
```

test/realm_test.dart
```diff
@@ -1085,6 +1085,175 @@ Future<void> main([List<String>? args]) async {
...
+    expect(await isRealm2Cancelled, isFalse);
+  });
+
+  baasTest('Realm.open (flexibleSync) - open two different Realms(for user1 and user2) and cancel only the second', (appConfiguration) async {
+    final app = App(appConfiguration);
+
+    final user1 = await app.logIn(Credentials.anonymous());
+    final configuration1 = Configuration.flexibleSync(user1, [Task.schema]);
+    var cancellationToken1 = CancellationToken();
+    final isRealm1Cancelled = getRealmAsync(configuration1, cancellationToken: cancellationToken1).thenIsCancelled();
```
blagoev, line 1172:
```suggestion
    final isRealm1Cancelled = getRealmAsync(configuration1, cancellationToken: cancellationToken1).getIsCancelled();
```

test/realm_test.dart
```diff
@@ -1109,6 +1278,50 @@ void openEncryptedRealm(List<int>? encryptionKey, List<int>? decryptionKey, {voi
   }
 }
 
+extension _FutureRealm on Future<Realm> {
```
blagoev, line 1281:
```suggestion
extension on Future<Realm> {
```
this will make the extension private

test/realm_test.dart
```diff
@@ -1085,6 +1085,175 @@ Future<void> main([List<String>? args]) async {
...
+    final realm = await getRealmAsync(configuration, cancellationToken: cancellationToken);
+
+    expect(realm, isNotNull);
+    expect(realm.isClosed, false);
+
+    cancellationToken.cancel();
+    expect(realm.isClosed, false);
+  });
+
+  baasTest('Realm.open (flexibleSync) - listen for download progress of an empty realm', (appConfiguration) async {
```
blagoev, line 1200:
```suggestion
  baasTest('Realm.open (flexibleSync) - listen for download progress on an empty realm', (appConfiguration) async {
```

test/test.dart
```diff
@@ -332,6 +341,17 @@ Realm getRealm(Configuration config) {
   return realm;
 }
 
+Future<Realm> getRealmAsync(Configuration config, {CancellationToken? cancellationToken, ProgressCallback? onProgressCallback}) async {
+  {
```
blagoev, line 345: is this formatted correctly. Or the `{` `}` seems not needed here.

lib/src/realm_class.dart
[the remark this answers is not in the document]
desistefanova: We have then in the arguments of the external API. Do we expect users to add the `cancellation_token` themselves?

test/realm_test.dart
blagoev: or perhaps
```
Future<bool> isCancelled() async {
```

lib/src/realm_class.dart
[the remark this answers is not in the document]
desistefanova: I removed them.

PR #732, opened 2022-07-30
Initial subscriptions callback

### LINKED ISSUE #739: Initial subscriptions callback
[https://www.mongodb.com/docs/realm-sdks/kotlin/1.0.1/library-sync/-realm%20-kotlin%20-s-d-k%20-%20-sync/io.realm.kotlin.mongodb.sync/-sync-configuration/-builder/initial-subscriptions.html](https://www.mongodb.com/docs/realm-sdks/kotlin/1.0.1/library-sync/-realm%20-kotlin%20-s-d-k%20-%20-sync/io.realm.kotlin.mongodb.sync/-sync-configuration/-builder/initial-subscriptions.html)

desistefanova:
Fixes #739 
Done:
- [x]  Update Changelog.md
- [x]  API documentation
- [x] Call initial subscription on Realm.open async method and Realm constructor.
It is decided this callback to be send for implementation in Core. Issue https://github.com/realm/realm-core/issues/5962


lib/src/realm_class.dart
```diff
@@ -116,7 +116,14 @@ class Realm {
     if (!dir.existsSync()) {
       dir.createSync(recursive: true);
     }
-    return realmCore.openRealm(config);
+
+    RealmHandle realm = realmCore.openRealm(config);
+    if (config is FlexibleSyncConfiguration) {
+      if (config.initialSubscriptionsConfiguration != null && config.initialSubscriptionsConfiguration!.rerunOnOpen) {
```
desistefanova, line 122: How can we detect whether it is the first opening of the Realm?

lib/src/realm_class.dart
```diff
@@ -112,11 +113,22 @@ class Realm implements Finalizable {
...
   Realm(Configuration config) : this._(config);
 
-  Realm._(this.config, [RealmHandle? handle, this._isInMigration = false]) : _handle = handle ?? _openRealm(config) {
+  Realm._(this.config, [RealmHandle? handle, this._isInMigration = false])
+      : _openedFirstTime = !File(config.path).existsSync(),
+        _handle = handle ?? _openRealm(config) {
     _populateMetadata();
+    if (config is FlexibleSyncConfiguration) {
+      final flxConfig = config as FlexibleSyncConfiguration;
+      if (flxConfig.initialSubscriptionsConfiguration?.callback != null && (_openedFirstTime || flxConfig.initialSubscriptionsConfiguration!.rerunOnOpen)) {
```
blagoev, line 128: This `if` is duplicated below. 
How about we move this to a helper private method and named it `_shouldInvokeInitialSubscriptions` which takes the config as parameter.

lib/src/configuration.dart
```diff
@@ -311,6 +316,7 @@ class FlexibleSyncConfiguration extends Configuration {
     super.encryptionKey,
     this.syncErrorHandler = defaultSyncErrorHandler,
     this.syncClientResetErrorHandler = const ManualSyncClientResetHandler(_defaultSyncClientResetHandler),
+    this.initialSubscriptionsConfiguration,
```
blagoev, line 319: So now we have configuration inside configuration. I think this is counter intuitive. 

I suggest the following alternative

```dart
FlexibleSyncConfiguration._(
.....
{
 initialSubscriptionCallback = ....,
 reRunInitialSubscriptionCallback = false
....
}
```

so flexibleSyncConfig now has a callback and a boolen flag to rerun

lib/src/realm_class.dart
```diff
@@ -112,11 +113,22 @@ class Realm implements Finalizable {
   /// and will not update when writes are made to the database.
   late final bool isFrozen = realmCore.isFrozen(this);
 
+  /// Returns true if the [Realm] is opened for the first time and the realm file is created.
+  late final bool _openedFirstTime;
```
blagoev, line 117:
```suggestion
  late final bool _initialOpen;
```

test/realm_test.dart
```diff
@@ -1226,7 +1226,7 @@ Future<void> main([List<String>? args]) async {
     expect(realm.isInTransaction, true);
   });
 
-  baasTest('Realm.open (flexibleSync)', (appConfiguration) async {
+  baasTest('Realm.open async (flexibleSync)', (appConfiguration) async {
```
blagoev, line 1229:
```suggestion
  baasTest('Realm.open async with flexibleSync', (appConfiguration) async {
```

blagoev: Not sure  why we are using brackets in the test names. It is supposed to be readable

test/realm_test.dart
```diff
@@ -1272,7 +1272,7 @@ Future<void> main([List<String>? args]) async {
     expect(await isRealmCancelled, isTrue);
   });
 
-  baasTest('Realm.open (flexibleSync) - open twice the same realm with the same CancelationToken cancels all', (appConfiguration) async {
+  baasTest('Realm.open async (flexibleSync) - open the same realm twice with the same CancelationToken cancel all', (appConfiguration) async {
```
blagoev, line 1275:
```suggestion
  baasTest('Realm.open async with flexibleSync - open the same realm twice with the same CancelationToken', (appConfiguration) async {
```

test/realm_test.dart
```diff
@@ -1353,20 +1353,43 @@ Future<void> main([List<String>? args]) async {
...
     expect(data.length, itemCount);
   });
 
-  baasTest('Realm.open (flexibleSync) - listen for download progress of a populated realm', (appConfiguration) async {
+  baasTest('Realm.open async (flexibleSync) with initial subscriptions - listen for download progress', (appConfiguration) async {
     final app = App(appConfiguration);
-    final config = await _addDataToAtlas(app);
+    final queryDifferentiator = generateRandomString(10);
+    final itemCount = 100;
+    await _addDataToAtlas(app, queryDifferentiator: queryDifferentiator, itemCount: itemCount);
```
blagoev, line 1382: Can we add this data while we are creating the apps? Seems we need this data on many tests now. And each test will be adding this data over and over again

blagoev (changes requested):

lib/src/configuration.dart
```diff
@@ -405,3 +411,19 @@ class SyncClientResetErrorHandler {
...
+class InitialSubscriptionsConfiguration {
+
+  const InitialSubscriptionsConfiguration(this.callback, {this.rerunOnOpen = false});
+
+  /// The callback that will be called in order to populate the initial [SubscriptionSet] for the [Realm].
+  final void Function(Realm) callback;
+
+  /// The default behavior is that callback is only invoked the first time the Realm is opened,
+  /// but if rerunOnOpen is true, it will be invoked every time the realm is opened.
+  final bool rerunOnOpen;
```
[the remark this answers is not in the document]
desistefanova: I don't understand. We don't open the app.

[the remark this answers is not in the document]
nirinchev: The idea is that the callback is invoked every time the app is launched, not every time the Realm file is opened. These may be the same if you only open the Realm once for the lifetime of your app, but if you open the Realm on every screen, it will be a problem.

test/realm_test.dart
desistefanova: Adding data takes time and will slow down the tests execution. Also the data that I add have dfferent Product name for each test.

nirinchev: I agree - we should not have test data added as part of setting up the apps because tests should not depend on each other - imagine having one test that deletes items from the test data, then another test that requires there are 100 elements. Then the second test will pass or fail depending on the order in which tests are run.

lib/src/configuration.dart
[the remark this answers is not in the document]
desistefanova: Added a discussion point.

lib/src/configuration.dart
```diff
@@ -405,3 +411,19 @@ class SyncClientResetErrorHandler {
 
 /// A client reset strategy where the user needs to fully take care of a client reset.
 typedef ManualSyncClientResetHandler = SyncClientResetErrorHandler;
+
+/// Represents an object with [FlexibleSyncConfiguration] configuration options for adding subscriptions initially before a [Realm] is opened.
+///
+/// {@category Configuration}
+class InitialSubscriptionsConfiguration {
+
+  const InitialSubscriptionsConfiguration(this.callback, {this.rerunOnOpen = false});
```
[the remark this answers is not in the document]
desistefanova: I removed this class InitialSubscriptionsConfiguration.

lib/src/realm_class.dart
```diff
@@ -112,11 +113,22 @@ class Realm implements Finalizable {
...
+
   /// Opens a `Realm` using a [Configuration] object.
   Realm(Configuration config) : this._(config);
 
-  Realm._(this.config, [RealmHandle? handle, this._isInMigration = false]) : _handle = handle ?? _openRealm(config) {
+  Realm._(this.config, [RealmHandle? handle, this._isInMigration = false])
+      : _openedFirstTime = !File(config.path).existsSync(),
```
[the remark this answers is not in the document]
desistefanova: Added a discussion point. I will try to provide a solution.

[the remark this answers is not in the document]
desistefanova: Added core issue: https://github.com/realm/realm-core/issues/5962

lib/src/configuration.dart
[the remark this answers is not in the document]
desistefanova: Added core issue: https://github.com/realm/realm-core/issues/5962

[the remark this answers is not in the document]
tgoyne: I disagree very very strongly on the idea that we should make rerunOnOpen literally once per run of an app.

As sort of a broad thing, we should be trying to have *less* implicit global state in core, not more. What we have already has always been a problem for things like tests and platforms where multiple VM states per process are normal, and we shouldn't be making the problem worse.

More specific to this, it's not an accident or bug that things like compactOnLaunch and the next-launch file actions for sync are actually "when the Realm is opened from scratch" and not once per run of the app. We've implied that they're once per app launch because we expect that to be how they normally work out (especially with synchronized Realms), but performing some operation a second time should never require killing an app and relaunching it.

Opening the Realm on each screen will typically not actually result in the Realm being reopened from scratch each time as the SyncSession isn't torn down synchronously. If the app does manage to actually fully close the Realm and reopen it on each screen transition then performance will already be very poor.

PR #740, opened 2022-08-02
Apple, Facebook and Google authentication

nielsenko:
- Implement support for Apple, Facebook and Google based authentication
- Update CHANGELOG

### DIFF
--- a/lib/src/credentials.dart
+++ b/lib/src/credentials.dart
@@ -28,11 +28,17 @@ enum AuthProviderType {
   /// For authenticating without credentials.
   anonymous,
 
-  _facebook,
-  _google,
-  _apple,
+  /// Authenticate with Facebook account.
+  facebook,
+
+  /// Authenticate with Google account
+  google,
+
+  /// Authenticate with Apple Id
+  apple,
+
   _custom,
-  
+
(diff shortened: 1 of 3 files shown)


lib/src/credentials.dart
```diff
@@ -28,11 +28,17 @@ enum AuthProviderType {
...
-  _apple,
+  /// Authenticate with Facebook account.
+  facebook,
+
+  /// Authenticate with Google account
+  google,
+
+  /// Authenticate with Apple Id
+  apple,
+
```
desistefanova, line 39:
```suggestion
  /// Authenticate with Apple Id
  apple,

  /// Authenticate with Facebook account.
  facebook,

  /// Authenticate with Google account
  google,
  
```

desistefanova: The order is important because the enums are matched to core by index.

desistefanova (changes requested):

lib/src/credentials.dart
nielsenko: We never actually send the enum to core. I just kept the order as it was. Anyway, it is fine with me.

desistefanova (approved):

PR #750, opened 2022-08-10
Update realm-core to v12.5.0

nielsenko:


lib/src/credentials.dart
```diff
@@ -28,6 +28,9 @@ enum AuthProviderType {
   /// For authenticating without credentials.
   anonymous,
 
+  /// For authenticating without credentials.
```
blagoev, line 31: Should we have more info here. If one looks at the API doc in the IDE it will not be clear how is this different from the other value. Perhaps we could write something like:
```
/// For authenticating without credentials using a new anonymous user.
```

blagoev (approved):
I have a little suggestion on the API doc for the enum

desistefanova (commented):
It would be nice to have a test with anonymousNoReuse option.

nielsenko:
> It would be nice to have a test with anonymousNoReuse option.

I have added the test-case: _Anonymous - new user_ to `credentials_test.dart`

nielsenko:
> I have a little suggestion on the API doc for the enum

👍

desistefanova (approved):

PR #754, opened 2022-08-11
Use Finalizable to ensure lexical liveness

nielsenko:
- HandleBase implements Finalizable, and uses a common NativeFinalizer
- [WIP] Classes with Finalizable fields implements Finalizable


nirinchev:
This looks reasonable, but still seems kind of hacky. I'm worried that a linker static analysis can discover those methods are unused and strip them from the compiled application. I'm not sure how much (if any) code stripping the dart linker does, especially when compiling to native, but it'd be interesting to confirm with the dart team this is the idiomatic approach for what we're trying to do.

lib/src/native/realm_core.dart
```diff
@@ -1738,10 +1749,14 @@ class ReleasableHandle<T extends NativeType> extends HandleBase<T> {
     if (released) {
       return;
     }
-
-    _realmLib.realm_dart_delete_finalizable(_finalizableHandle, this);
+    _nativeFinalizer.detach(this);
     _realmLib.realm_release(_pointer.cast());
     released = true;
+    assert(() {
```
blagoev, line 1755: we can remove these per above

lib/src/realm_class.dart
```diff
@@ -237,6 +238,11 @@ class Realm {
     _subscriptions?.handle.release();
     _subscriptions = null;
 
+    final c = config;
```
blagoev, line 241: I think we don't need this one now

lib/src/user.dart
```diff
@@ -29,14 +29,14 @@ import './app.dart';
 /// locally on the device, and should be treated as sensitive data.
 /// {@category Application}
 class User {
-  final App? _app;
+  App? _app;
```
blagoev, line 32: I am not sure why we are changing this. Should we keep the `final` here

blagoev (changes requested):
Added some suggestions. Also seems containing some debug left overs.

nielsenko:
> This looks reasonable, but still seems kind of hacky. I'm worried that a linker static analysis can discover those methods are unused and strip them from the compiled application. I'm not sure how much (if any) code stripping the dart linker does, especially when compiling to native, but it'd be interesting to confirm with the dart team this is the idiomatic approach for what we're trying to do.

See reply by dartlang developer here: https://github.com/dart-lang/sdk/issues/49643#issuecomment-1214872921

lib/src/user.dart
nielsenko: No. I think it was a mistake. Making this final prevents us from using the `??=` operator in the actual getter.

lib/src/realm_class.dart
nielsenko: Correct that is an old residue

lib/src/native/realm_core.dart
```diff
@@ -1672,15 +1672,26 @@ class LastError {
...
 }
 
-abstract class HandleBase<T extends NativeType> {
+void _traceFinalization(Object o) {
+  Realm.logger.log(RealmLogLevel.trace, 'Finalizing: $o');
+}
+
+final _debugFinalizer = Finalizer<Object>(_traceFinalization);
+
+final _nativeFinalizer = NativeFinalizer(_realmLib.addresses.realm_release);
```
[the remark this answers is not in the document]
nielsenko: Can't we just add the same debug code to `realm_release` itself?

lib/src/native/realm_core.dart
nielsenko: asserts are automatically dropped in release mode. I find it convenient to have the ability to get trace from the dart side as well.

lib/src/native/realm_core.dart
```diff
@@ -1672,15 +1672,26 @@ class LastError {
   }
 }
 
-abstract class HandleBase<T extends NativeType> {
+void _traceFinalization(Object o) {
+  Realm.logger.log(RealmLogLevel.trace, 'Finalizing: $o');
+}
+
+final _debugFinalizer = Finalizer<Object>(_traceFinalization);
```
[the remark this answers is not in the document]
nielsenko: This is just for tracing what happens on the dart side. It is automatically dropped in release mode, since it runs as part of an assert, and all it does is call `Realm.logger.log(RealmLogLevel.trace, 'Finalizing: $o');`.

lib/src/native/realm_core.dart
```diff
@@ -1672,15 +1672,26 @@ class LastError {
...
+  void keepAlive() {}
 
   HandleBase(this._pointer, int size) {
-    _finalizableHandle = _realmLib.realm_dart_attach_finalizer(this, _pointer.cast(), size);
-    if (_finalizableHandle == nullptr) {
-      throw Exception("Error creating $runtimeType");
-    }
+    _nativeFinalizer.attach(this, _pointer.cast(), detach: this, externalSize: size);
+    assert(() {
+      _debugFinalizer.attach(this, _pointer);
```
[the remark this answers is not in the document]
nielsenko: During development I found it quite convenient to monitor finalization on both the dart and the native side.

nielsenko:
BTW, in case it isn't obvious. The need to have these `keepAlive` extension methods should eventually disappear, once https://github.com/dart-lang/sdk/issues/49643 is resolved.

lib/src/realm_object.dart
```diff
@@ -178,7 +179,7 @@ class RealmCoreAccessor implements RealmAccessor {
   }
 }
 
-mixin RealmEntity {
+mixin RealmEntity implements Finalizable {
```
nirinchev, line 182: Should we implement `Finalizable` on `RealmObject` and add `keepAlive` extension?

nielsenko: I believe it is enough to have it on the base `RealmEntity` as `RealmEntityInternal` extension class implements `keepAlive`.

nielsenko: Fixed

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
nielsenko: Reworked as discussed on round-table meeting

lib/src/native/realm_core.dart
nielsenko: Reworked as discussed on round-table meeting

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
nielsenko: Reworked as discussed on round-table meeting

blagoev (approved):

PR #757, opened 2022-08-12
Compensate for breaking change to TypeImpl

nielsenko:
PseudoType inherits from TypeImpl that introduced an abstract
element2 getter and deprecated element. This PR compensate for that.

It also updated all deps to latest major versions.


build.yaml
```diff
@@ -11,6 +11,9 @@ targets:
       realm_generator|realm_object_builder:
         generate_for:
           include:
```
desistefanova, line 13: Why do we need to include test and example folder?

desistefanova (approved):

build.yaml
```diff
@@ -11,6 +11,9 @@ targets:
...
+            - test/**.dart
+            - example/**.dart
```
desistefanova, line 15: Why do we need to include test and example folder?

nielsenko: This is more to ensure that only test and example folders are processed.

desistefanova: What about "flutter\realm_flutter\example"? It should be also included, I think.

nielsenko: That is a project of its own with its own pubspec.yaml and buiild.yaml. You should run the generator in the directory.

desistefanova: Yes, right.

PR #851, opened 2022-08-17
Allow nullable PKs

### LINKED ISSUE #458: Primary keys should allow to be nullable

nirinchev:
Fixes #458

Re-implementation of https://github.com/realm/realm-dart/pull/459


nielsenko (approved):
LGTM

desistefanova (commented):
We will have to update https://jira.mongodb.org/secure/attachment/382379/dart.txt after this PR is merged or released. PrimaryKey should be added to _AllPrimitiveOptional.

generator/test/good_test_data/primary_key.expected
```diff
@@ -30,6 +30,35 @@ class IntPK extends _IntPK with RealmEntity, RealmObject {
   }
 }
 
+class NullableIntPK extends _NullableIntPK with RealmEntity, RealmObject {
+  NullableIntPK(
+    int? id,
```
desistefanova, line 35: Isn't it strange that nullable primary key is always expected as an argument even though it allows null? It is reasonable if only one null value is allowed in the collection. I suppose it is like that.

desistefanova (approved):

generator/test/good_test_data/primary_key.expected
nirinchev: Yeah - that's what I was thinking - I wanted people to explicitly specify `null` for the PK rather than think they can omit it.

desistefanova: OK

blagoev (approved):

PR #853, opened 2022-08-18
Add support for string based/dynamic API on RealmObject

nirinchev:
Adds a string-based API to read properties of a RealmObject. Right now we only support reading as that's required to get migrations working, but it can be extended in the future to also support setting.

```dart
final obj = realm.dynamic.find("Foo", 123);

// Get simple primitive
final str = obj.dynamic.get<String>("stringProp");

// Get another object
final bar = obj.dynamic.get<RealmObject>("bar");
final barString = bar.dynamic.get<String>("stringProp");

// Get a list of objects
final bars = obj.dynamic.getList<RealmObject>("bars");

// Iterate the list as normal
for (final bar in bars) {
  print(bar.dynamic.get<DateTime>("dateProp");
}
```

Additionally, implements `noSuchMethod` on `RealmObject` in order to support dynamic invocation:

```dart
dynamic obj = realm.dynamic.find("Foo", 123);

// Get simple primitive
String str = obj.stringProp;

// Get another object
dynamic bar = obj.bar;
String barString = bar.stringProp;

// Get a list of objects
final bars = obj.bars;

// Iterate the list as normal
for (final bar in bars) {
  print(bar.dateProp);
}
```

Part of #70.

**Note**: This only adds support for reading properties, but not for setting them. It won't be difficult to extend the API and support setting of properties, but is out of scope for this PR.

### DIFF
--- a/lib/src/list.dart
+++ b/lib/src/list.dart
@@ -30,7 +30,7 @@ import 'results.dart';
 /// added to or deleted from the collection or from the Realm.
 ///
 /// {@category Realm}
-abstract class RealmList<T extends Object> with RealmEntity implements List<T>, Finalizable {
+abstract class RealmList<T extends Object?> with RealmEntity implements List<T>, Finalizable {
   late final RealmObjectMetadata? _metadata;
 
   /// Gets a value indicating whether this collection is still valid to use.
@@ -44,7 +44,7 @@ abstract class RealmList<T extends Object> with RealmEntity implements List<T>, 
   factory RealmList(Iterable<T> items) => UnmanagedRealmList(items);
 }
 
-class ManagedRealmList<T extends Object> extends collection.ListBase<T> with RealmEntity implements RealmList<T> {
+class ManagedRealmList<T extends Object?> extends collection.ListBase<T> with RealmEntity implements RealmList<T> {
   final RealmListHandle _handle;
 
   @override
@@ -97,7 +97,7 @@ class ManagedRealmList<T extends Object> extends collection.ListBase<T> with Rea
   bool get isValid => realmCore.listIsValid(this);
 }
 
-class UnmanagedRealmList<T extends Object> extends collection.ListBase<T> with RealmEntity implements RealmList<T> {
+class UnmanagedRealmList<T extends Object?> extends collection.ListBase<T> with RealmEntity implements RealmList<T> {
   final _unmanaged = <T?>[]; // use T? for length=
 
   UnmanagedRealmList([Iterable<T>? items]) {
@@ -156,7 +156,7 @@ extension RealmListOfObject<T extends RealmObject> on RealmList<T> {
 }
 
 /// @nodoc
-extension RealmListInternal<T extends Object> on RealmList<T> {
+extension RealmListInternal<T extends Object?> on RealmList<T> {
   @pragma('vm:never-inline')
   void keepAlive() {
     final self = this;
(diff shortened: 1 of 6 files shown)


lib/src/list.dart
```diff
@@ -30,7 +30,7 @@ import 'results.dart';
 /// added to or deleted from the collection or from the Realm.
 ///
 /// {@category Realm}
-abstract class RealmList<T extends Object> with RealmEntity implements List<T>, Finalizable {
+abstract class RealmList<T extends Object?> with RealmEntity implements List<T>, Finalizable {
```
nirinchev, line 33: This liberalizes the type args on `RealmList` to support nullable objects. It doesn't actually add support for lists of nullables - that is done in https://github.com/realm/realm-dart/pull/708, but is necessary to simplify a lot of the generic constraints elsewhere.

lib/src/native/realm_core.dart
```diff
@@ -2053,7 +2055,7 @@ extension on Pointer<realm_value_t> {
       case realm_value_type.RLM_TYPE_INT:
         return ref.values.integer;
       case realm_value_type.RLM_TYPE_BOOL:
-        return ref.values.boolean != 0;
+        return ref.values.boolean;
```
nirinchev, line 2058: This must be a change with the new ffi and seems like it was a bug - not sure how/why our tests didn't catch it, but likely warrants an investigation.

test/dynamic_realm_test.dart
```diff
@@ -69,8 +69,117 @@ Future<void> main([List<String>? args]) async {
...
+    expect(actualDynamic.dates, expected.dates);
+    expect(actualDynamic.doubles, expected.doubles);
+    expect(actualDynamic.objectIds, expected.objectIds);
+    expect(actualDynamic.uuids, expected.uuids);
+    expect(actualDynamic.ints, expected.ints);
+  }
+
   for (var isDynamic in [true, false]) {
-    Realm getDynamicRealm(Realm original) {
+    Realm _getDynamicRealm(Realm original) {
```
nirinchev, line 182: The changes in this file are mostly whitespace due to grouping tests together. I recommend reviewing with `?w=1`.

lib/src/native/realm_core.dart
blagoev: added a test and fixed it in #854. This can be left here as well

lib/src/realm_object.dart
```diff
@@ -18,13 +18,16 @@
 
 import 'dart:async';
 import 'dart:ffi';
+import 'dart:mirrors';
```
blagoev, line 21: If I am not mistaken we can not use mirrors on Flutter

lib/src/realm_object.dart
```diff
@@ -18,13 +18,16 @@
...
 
 import 'list.dart';
 import 'native/realm_core.dart';
 import 'realm_class.dart';
 
+typedef DartDynamic = dynamic;
```
blagoev, line 27: I suggest we don't hide this under our own name

blagoev (changes requested):

lib/src/realm_object.dart
nirinchev: This is just a typedef - it's not going to affect users. The reason why I need to typedef it is to avoid a collision with the `dynamic` property on `RealmObject`.

lib/src/realm_object.dart
nirinchev: Ugh... that's tricky. I'm using `MirrorSystem.getName` to get the member name in `noSuchMethod`. Without it, I'm not sure how we can implement it.

blagoev (approved):
I guess we can add more tests for the new validations

lib/src/realm_object.dart
nielsenko: Won't users be hit by the same?

nielsenko (approved):

PR #854, opened 2022-08-19
get set all prop types

blagoev:
We were lacking a read/write test for all types, so we missed the boolean bug with the new ffigen


nielsenko (approved):
Could consider also adding a test that roundtrip all property types over BaaS.

PR #863, opened 2022-08-22
Add support for migrations

nirinchev:
Adds support for providing a migration callback in local configs.

- [x] Callback executed when schema needs migration
- [x] Support `migration.findInNewRealm`
- [x] Support `migration.renameProperty`
- [x] Support `migration.removeType`. Blocked on https://github.com/realm/realm-core/issues/5766
- [x] Migration correctly propagates user exceptions through Core

Supersedes https://github.com/realm/realm-dart/pull/494.

Fixes https://github.com/realm/realm-dart/issues/70

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
     }
 
     if (oldSchemaVersion == 2) {
-      // Between v2 and v3 we fixed a typo where someone had mispelled the 'Foo.name' property.
+      // Between v2 and v3 we fixed a typo where someone had mispelled the 'Person.name' property.
       migration.renameProperty('Person', 'nmae', 'name');
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,11 @@
   final config = Configuration.local([Person.schema], schemaVersion: 4, migrationCallback: (migration, oldSchemaVersion) {
     if (oldSchemaVersion == 1) {
       // Between v1 and v2 we removed the Bar type
-      migration.removeType('Bar');
+      migration.deleteType('Bar');
     }
 
     if (oldSchemaVersion == 2) {
-      // Between v2 and v3 we fixed a typo where someone had mispelled the 'Person.name' property.
+      // Between v2 and v3 we fixed a typo in the 'Person.name' property.
       migration.renameProperty('Person', 'nmae', 'name');
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,9 @@
 
 ### Enhancements
 * Added support for "frozen objects" - these are objects, queries, lists, or Realms that have been "frozen" at a specific version. All frozen objects can be accessed and queried as normal, but attempting to mutate them or add change listeners will throw an exception. `Realm`, `RealmObject`, `RealmList`, and `RealmResults` now have a method `freeze()` which returns an immutable version of the object, as well as an `isFrozen` property which can be used to check whether an object is frozen. (Issue [#56](https://github.com/realm/realm-dart/issues/56))
-* Added support for migrations for local Realms. You can now construct a configuration with a migration callback that will be invoked if the schema version of the file on disk is lower than the schema version supplied by the callback. A minimal example looks like this:
+* Added support for migrations for local Realms. You can now construct a configuration with a migration callback that will be invoked if the schema version of the file on disk is lower than the schema version supplied by the callback. (Issue [#70](https://github.com/realm/realm-dart/issues/70))
+
+  A minimal example looks like this:
   ```dart
   final config = Configuration.local([Person.schema], schemaVersion: 4, migrationCallback: (migration, oldSchemaVersion) {
     if (oldSchemaVersion == 1) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,34 @@
 **This project is in the Beta stage. The API should be quite stable, but occasional breaking changes may be made.**
 
 ### Enhancements
-* None
+* Added support for migrations for local Realms. You can now construct a configuration with a migration callback that will be invoked if the schema version of the file on disk is lower than the schema version supplied by the callback. A minimal example looks like this:
+  ```dart
+  final config = Configuration.local([Person.schema], schemaVersion: 4, migrationCallback: (migration, oldSchemaVersion) {
+    if (oldSchemaVersion == 1) {
+      // Between v1 and v2 we removed the Bar type
+      migration.removeType('Bar');
+    }
+
+    if (oldSchemaVersion == 2) {
+      // Between v2 and v3 we fixed a typo where someone had mispelled the 'Foo.name' property.
+      migration.renameProperty('Person', 'nmae', 'name');
+    }
+
+    if (oldSchemaVersion == 3) {
+      final oldPeople = migration.oldRealm.dynamic.all('Person');
+      for (final oldPerson in oldPeople) {
+        final newPerson = migration.findInNewRealm<Person>(oldPerson);
+        if (newPerson == null) {
+          // That person must have been deleted, so nothing to do.
+          continue;
+        }
+
+        // Between v3 and v4 we're obfuscating the users' exact age by storing age group instead.
+        newPerson.ageGroup = calculateAgeGroup(oldPerson.dynamic.get<int>('age'));
+      }
+    }
+  });
+  ```
 
 ### Fixed
 * Allow null arguments on query. ([#872](https://github.com/realm/realm-dart/pull/872)). Fixes [#871](https://github.com/realm/realm-dart/issues/871)
--- a/generator/test/good_test_data/all_types.expected
+++ b/generator/test/good_test_data/all_types.expected
@@ -76,7 +76,7 @@ class Bar extends _Bar with RealmEntity, RealmObject {
   @override
   String get id => RealmObject.get<String>(this, 'id') as String;
   @override
-  set id(String value) => throw RealmUnsupportedSetError();
+  set id(String value) => RealmObject.set(this, 'id', value);
 
   @override
   bool get aBool => RealmObject.get<bool>(this, 'aBool') as bool;
--- a/lib/src/configuration.dart
+++ b/lib/src/configuration.dart
@@ -48,7 +48,8 @@ typedef ShouldCompactCallback = bool Function(int totalSize, int usedSize);
 /// Realms, even if all objects in the Realm are deleted.
 typedef InitialDataCallback = void Function(Realm realm);
 
-/// the signature of a callback that will be executed when the schema of the Realm changes.
+/// The signature of a callback that will be executed when the schema of the Realm changes.
+///
 /// The `oldSchemaVersion` argument indicates the version from which the Realm migrates while
 /// the `migration` argument contains references to the Realm just before and just after the migration.
 typedef MigrationCallback = void Function(Migration migration, int oldSchemaVersion);
@@ -241,7 +242,7 @@ enum SessionStopPolicy {
   afterChangesUploaded, // Once all Realms/Sessions go out of scope, wait for uploads to complete and stop.
 }
 
-///The signature of a callback that will be invoked whenever a [SyncError] occurs for the synchronized Realm.
+/// The signature of a callback that will be invoked whenever a [SyncError] occurs for the synchronized Realm.
 ///
 /// Client reset errors will not be reported through this callback as they are handled by [SyncClientResetErrorHandler].
 typedef SyncErrorHandler = void Function(SyncError);
--- a/lib/src/configuration.dart
+++ b/lib/src/configuration.dart
@@ -50,8 +50,8 @@ typedef InitialDataCallback = void Function(Realm realm);
 
 /// The signature of a callback that will be executed when the schema of the Realm changes.
 ///
+/// The `migration` argument contains references to the Realm just before and just after the migration.
 /// The `oldSchemaVersion` argument indicates the version from which the Realm migrates while
-/// the `migration` argument contains references to the Realm just before and just after the migration.
 typedef MigrationCallback = void Function(Migration migration, int oldSchemaVersion);
 
 /// Configuration used to create a [Realm] instance
--- a/lib/src/configuration.dart
+++ b/lib/src/configuration.dart
@@ -25,6 +25,7 @@ import 'native/realm_core.dart';
 import 'realm_class.dart';
 import 'init.dart';
 import 'user.dart';
+import 'migration.dart';
 
 /// The signature of a callback used to determine if compaction
 /// should be attempted.
@@ -46,6 +47,11 @@ typedef ShouldCompactCallback = bool Function(int totalSize, int usedSize);
 /// add some initial data that your app needs. The function will not execute for existing
 /// Realms, even if all objects in the Realm are deleted.
 typedef InitialDataCallback = void Function(Realm realm);
+
+/// the signature of a callback that will be executed when the schema of the Realm changes.
+/// The `oldSchemaVersion` argument indicates the version from which the Realm migrates while
+/// the `migration` argument contains references to the Realm just before and just after the migration.
+typedef MigrationCallback = void Function(Migration migration, int oldSchemaVersion);
 
 /// Configuration used to create a [Realm] instance
 /// {@category Configuration}
@@ -122,6 +128,7 @@ abstract class Configuration implements Finalizable {
     bool disableFormatUpgrade = false,
     bool isReadOnly = false,
     ShouldCompactCallback? shouldCompactCallback,
+    MigrationCallback? migrationCallback,
   }) =>
       LocalConfiguration._(
         schemaObjects,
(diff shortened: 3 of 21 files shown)


lib/src/native/realm_core.dart
```diff
@@ -1832,15 +1891,15 @@ class RealmLinkHandle {
...
   RealmResultsHandle._(Pointer<realm_results> pointer) : super(pointer, 872);
 }
 
-class RealmListHandle extends HandleBase<realm_list> {
+class RealmListHandle extends ReleasableHandle<realm_list> {
   RealmListHandle._(Pointer<realm_list> pointer) : super(pointer, 88);
 }
 
-class RealmQueryHandle extends HandleBase<realm_query> {
+class RealmQueryHandle extends ReleasableHandle<realm_query> {
```
nirinchev, line 1902: These changes are not strictly necessary, but without them we can't implement some tests (there are collections that outlive the migration, which crashes the tests when they try to refresh themselves. This will be addressed as part of the handle ownership PR, so I'd rather leave it as is and have better test coverage.

generator/test/good_test_data/all_types.expected
```diff
@@ -76,7 +76,7 @@ class Bar extends _Bar with RealmEntity, RealmObject {
   @override
   String get id => RealmObject.get<String>(this, 'id') as String;
   @override
-  set id(String value) => throw RealmUnsupportedSetError();
+  set id(String value) => RealmObject.set(this, 'id', value);
```
nielsenko, line 79: Why?

test/migration_test.dart
```diff
@@ -0,0 +1,391 @@
+////////////////////////////////////////////////////////////////////////////////
```
nielsenko, line 1: Remember to add this to the flutter tests as well

nielsenko (changes requested):
Everything is suddenly settable on the typed realm objects, why is that?

generator/test/good_test_data/all_types.expected
nirinchev: You can change PKs during migration.

nirinchev: This still throws an exception if you're outside of migration.

lib/src/configuration.dart
```diff
@@ -47,6 +48,11 @@ typedef ShouldCompactCallback = bool Function(int totalSize, int usedSize);
 /// Realms, even if all objects in the Realm are deleted.
 typedef InitialDataCallback = void Function(Realm realm);
 
+/// the signature of a callback that will be executed when the schema of the Realm changes.
```
desistefanova, line 51:
```suggestion
/// The signature of a callback that will be executed when the schema of the Realm changes.
```

lib/src/migration.dart
```diff
@@ -0,0 +1,60 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////////
+
+import 'realm_class.dart';
+import 'native/realm_core.dart';
+import './realm_object.dart';
+
+class Migration {
```
desistefanova, line 23: Missing API documentation

lib/src/native/realm_core.dart
```diff
@@ -187,6 +197,9 @@ class _RealmCore {
             nullptr,
           );
         }
+        if (config.migrationCallback != null) {
+          _realmLib.realm_config_set_migration_function(configHandle._pointer, Pointer.fromFunction(migration_callback, false), config.toWeakHandle(), nullptr);
```
desistefanova, line 201: Isn't it better to use the approach with passing scheduler, similar to errorHandlerUserdata below in the same method?

test/test.g.dart
```diff
@@ -18,7 +18,7 @@ class Car extends _Car with RealmEntity, RealmObject {
   @override
   String get make => RealmObject.get<String>(this, 'make') as String;
   @override
-  set make(String value) => throw RealmUnsupportedSetError();
+  set make(String value) => RealmObject.set(this, 'make', value);
```
desistefanova, line 21: It will be nice to re-generate the models also in dart and flutter examples under realm-dart folder.

lib/src/native/realm_core.dart
nirinchev: We know that the migration function is going to be executed on the caller thread, so we don't need to wrap the Core callback in our own that dispatches to the isolate thread.

test/migration_test.dart
```diff
@@ -0,0 +1,391 @@
...
+part 'migration_test.g.dart';
+
+@RealmModel()
+@MapTo("Person")
+class _PersonIntName {
+  late int name;
+}
+
+@RealmModel()
+@MapTo('Student')
```
desistefanova, line 37: Do you think we can extend the "MapTo" API doc so users know it's being used during the migration process?

test/migration_test.dart
```diff
@@ -0,0 +1,391 @@
...
+
+    final v2Config = Configuration.local([MyObjectWithoutTypo.schema], schemaVersion: 2, migrationCallback: (migration, oldSchemaVersion) {
+      expect(oldSchemaVersion, 1);
+
+      migration.renameProperty('MyObject', 'nmae', 'name');
+    });
+
+    final v2Realm = getRealm(v2Config);
+
+    // We renamed nmae to name, but don't rename vlaue to value so we expect
```
desistefanova, line 249:
```suggestion
// We renamed 'nmae' to 'name', but don't rename 'vlaue' to 'value' so we expect
```

desistefanova (commented):
I added some small comments. 
What about CHANGELOG.md? I think we have to add a couple of lines there.

test/migration_test.dart
nirinchev: Not sure what you mean - it's not strictly necessary to use `MapTo` here. The reason why I'm doing it is to simulate going from v1 of the app to v2.

generator/test/good_test_data/all_types.expected
nielsenko: Eventually (when static meta programming lands) these setter should not exist at all, and it will be a compile time error to use them. I would like if we could use `RealmObject.set` (or `instanceSchema.properties['id'].setValue()`) directly where needed.

desistefanova (approved):

generator/test/good_test_data/all_types.expected
nielsenko: I relent 😄  We will revisit this once static meta programming has landed, and we re-work the generator.

lib/src/realm_object.dart
```diff
@@ -369,6 +381,13 @@ class RealmException implements Exception {
   }
 }
 
+class UserCallbackException extends RealmException {
```
blagoev, line 384: no API doc

lib/src/realm_object.dart
```diff
@@ -369,6 +381,13 @@ class RealmException implements Exception {
...
+  final Object userException;
```
blagoev, line 385: we are registering the exception instance and later we are assigning it to an object in the `UserCallbackError`. Should this property be an Exception instead? So at least the `message` can be accessed.

lib/src/native/realm_core.dart
```diff
@@ -648,6 +685,27 @@ class _RealmCore {
...
+  }
+
+  void renameProperty(Realm realm, String objectType, String oldName, String newName, SchemaHandle schema) {
+    using((Arena arena) {
+      _realmLib.invokeGetBool(() => _realmLib.realm_schema_rename_property(
+          realm.handle._pointer, schema._pointer, objectType.toCharPtr(arena), oldName.toCharPtr(arena), newName.toCharPtr(arena)));
+    });
+  }
+
+  bool removeType(Realm realm, String objectType) {
```
blagoev, line 701:
```suggestion
  bool deleteType(Realm realm, String objectType) {
```
as per the above suggestion

lib/src/realm_object.dart
```diff
@@ -182,9 +185,18 @@ class RealmCoreAccessor implements RealmAccessor {
...
       }
+
       if (value is RealmObject && !value.isManaged) {
         object.realm.add(value, update: update);
       }
+
+      if (propertyMeta.isPrimaryKey && !isInMigration) {
+        final currentValue = realmCore.getProperty(object, propertyMeta.key);
+        if (currentValue != value) {
+          throw RealmException("Primary key cannot be changed (original value: '$currentValue', supplied value: '$value')");
```
blagoev, line 196: Do we really need to be so nice. Why not throw the exception on set, as we did previously. Is there a chance this equality compare to not go as planned and skip some values. 
To me we better no allow `set` in this case is just because it adds one more hidden case to think about.  `Set` always throws on PKs is better than `Set` throws on PKs unless the value is the same. This might catch someone off-guard and we better remove all possibilities for mistakes.

test/realm_object_test.dart
```diff
@@ -133,7 +133,14 @@ Future<void> main([List<String>? args]) async {
       realm.write(() {
         car.make = "Audi";
       });
-    }, throws<RealmUnsupportedSetError>());
+    }, throws<RealmException>("Primary key cannot be changed (original value: 'Tesla', supplied value: 'Audi'"));
+
+    // If we don't change the PK, setting it is a no-op
```
blagoev, line 138: you are being so kind :)

CHANGELOG.md
```diff
@@ -3,7 +3,34 @@
 **This project is in the Beta stage. The API should be quite stable, but occasional breaking changes may be made.**
 
 ### Enhancements
-* None
+* Added support for migrations for local Realms. You can now construct a configuration with a migration callback that will be invoked if the schema version of the file on disk is lower than the schema version supplied by the callback. A minimal example looks like this:
+  ```dart
+  final config = Configuration.local([Person.schema], schemaVersion: 4, migrationCallback: (migration, oldSchemaVersion) {
+    if (oldSchemaVersion == 1) {
+      // Between v1 and v2 we removed the Bar type
+      migration.removeType('Bar');
```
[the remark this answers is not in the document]
nirinchev: I didn't realize that - yes, I can change it to delete.

lib/src/migration.dart
```diff
@@ -0,0 +1,81 @@
...
+/// to another. It contains the properties for the Realm before and after the migration.
+/// After the migration is complete, [newRealm] will become the authoritative version of
+/// the file.
+class Migration {
+  final SchemaHandle _schema;
+
+  /// The Realm as it existed just before the migration. Since the models have changed,
+  /// this Realm can only be accessed via the dynamic API (i.e. [Realm.dynamic] and
+  /// [RealmObject.dynamic]).
+  final Realm oldRealm;
```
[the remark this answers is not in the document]
nirinchev: There are more properties on the realm class, which are useful - e.g. the schema/config.

lib/src/migration.dart
```diff
@@ -0,0 +1,81 @@
...
+  /// The Realm as it exists after the migration. Before the end of the callback, you need
+  /// to make sure that all relevant data has been migrated from [oldRealm] into [newRealm].
+  final Realm newRealm;
+
+  Migration._(this.oldRealm, this.newRealm, this._schema);
+
+  /// Finds an object obtained from [oldRealm] in [newRealm]. This is useful when you
+  /// are working with objects without primary keys and want to update some information
+  /// about the object as part of the migration.
+  T? findInNewRealm<T extends RealmObject>(RealmObject oldObject) {
```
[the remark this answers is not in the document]
nirinchev: We can expose a more general-purpose API. My thinking here was that people would be iterating over the old objects and looking up their versions in the new realm as that's the conventional way of doing migrations. But that may be too prescriptive - I don't have strong opinions and would be fine either way. 

@nielsenko @desistefanova what are your thoughts here - do you prefer a more specialized API or a more general-purpose one?

lib/src/realm_object.dart
nirinchev: My understanding is that in dart you can throw any object, which is why the catch clause receives an object as an argument. We're simply storing that and rethrowing it. E.g.

```dart
config.migrationCallback = (migration, oldVersion) {
	throw 5;
};
```

If we tried to cast that to an exception instance, we would fail.

lib/src/realm_object.dart
nirinchev: These are primitive values, so the equality operator should always produce a correct result. I'm generally more inclined to avoid throwing exceptions if absolutely necessary - e.g. if they're using some automapper-like library where you can copy the values from one object to another, it would be kind of annoying if they have to do extra configuration to skip primary key properties.

lib/src/migration.dart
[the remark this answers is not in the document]
blagoev: we can include `config` and `schema` as properties here next to the `DynamicRealm` property?

lib/src/migration.dart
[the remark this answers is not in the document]
desistefanova: I prefer to know where I'm searching to find something in the new realm or in the old realm. findInNewRealm is OK for me. It could be also findInMigratedRealm.

lib/src/migration.dart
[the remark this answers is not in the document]
nirinchev: I tried to do something similar by creating a new migration-specific version of the Realm class which exposes the dynamic API. Can you check if that is in line with what you were thinking?

lib/src/migration.dart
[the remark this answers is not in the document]
nielsenko: I think the behaviour of `findInNewRealm` is quite different from `newRealm.find` in that it does not require a primary key. If we re-worked `find` so that it could actually look up a realm object from another realm without using the primary key, then we can drop `findInNewRealm`. Otherwise I think we should keep it, and with it the name, since there are two realms on a `MIgration` and you can only use this, to look up in one of them.

CHANGELOG.md
```diff
@@ -4,6 +4,34 @@
 
 ### Enhancements
 * Added support for "frozen objects" - these are objects, queries, lists, or Realms that have been "frozen" at a specific version. All frozen objects can be accessed and queried as normal, but attempting to mutate them or add change listeners will throw an exception. `Realm`, `RealmObject`, `RealmList`, and `RealmResults` now have a method `freeze()` which returns an immutable version of the object, as well as an `isFrozen` property which can be used to check whether an object is frozen. (Issue [#56](https://github.com/realm/realm-dart/issues/56))
+* Added support for migrations for local Realms. You can now construct a configuration with a migration callback that will be invoked if the schema version of the file on disk is lower than the schema version supplied by the callback. A minimal example looks like this:
```
blagoev, line 7: This should link to the same PR #863

```suggestion
* Added support for migrations for local Realms. You can now construct a configuration with a migration callback that will be invoked if the schema version of the file on disk is lower than the schema version supplied by the callback. ([#863](https://github.com/realm/realm-dart/issues/863))

A minimal example looks like this:
```

lib/src/native/realm_core.dart
blagoev: This got resolved but the method is the same. Should we rename to `deleteType`?

lib/src/native/realm_core.dart
```diff
@@ -421,6 +434,29 @@ class _RealmCore {
...
+        return false;
+      }
+
+      final oldSchemaVersion = _realmLib.realm_get_schema_version(oldRealmHandle);
+      final oldConfig = Configuration.local([], path: config.path, isReadOnly: true, schemaVersion: oldSchemaVersion);
+      final oldRealm = RealmInternal.getUnowned(oldConfig, RealmHandle._unowned(oldRealmHandle), isInMigration: true);
+
+      final newRealm = RealmInternal.getUnowned(config, RealmHandle._unowned(newRealmHandle), isInMigration: true);
+
+      final migration = MigrationInternal.create(oldRealm.getMigrationRealm(), newRealm, SchemaHandle.unowned(schema));
```
blagoev, line 451: Why not create the `MigrationRealm` inside the `MigrationInternal.create`. Thus we hide the `getMigrationRealm` cause right now it's available everywhere. So the signature of `MigrationInternal.create` would take two Realms and a SchemaHandle.

blagoev (changes requested):
Some small things left

lib/src/migration.dart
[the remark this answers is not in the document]
blagoev: Yes I like that. 
Thus we don't have the limitation of the previous API where some methods should not be called and we capture this in an API doc. 
My hope is we don't get into a situation in the future where we need really a full blown (old) Realm instance in that place. Then we would need a breaking change.

lib/src/native/realm_core.dart
nirinchev: Renamed the public API but forgot the Core API, will fix.

lib/src/native/realm_core.dart
nirinchev: `MigrationInternal` is in a different file and doesn't have access to the private constructor of `MigrationRealm`. I could have created a `MigrationRealmInternal` and expose the private constructor but that seemed like an overkill.

blagoev: oh that is in `RealmInternal` that will work. What is tripping me over is its not static. Can we make this static and pass the realm instance. Similar to `getUnowned`

blagoev (approved):

JamesMcIntosh:
@nirinchev Are migrations ready for proper usage as the main README.md lists them as a limitation? see:
https://github.com/realm/realm-dart#limitations

PR #872, opened 2022-08-25
Fix signature on query method, and add test with null

nielsenko:


test/results_test.dart
```diff
@@ -315,6 +315,20 @@ Future<void> main([List<String>? args]) async {
     expect(() => teams.length, throws<RealmException>("Access to invalidated Results objects"));
   });
 
+  test('Results - query with nil', () {
```
blagoev, line 318:
```suggestion
  test('Results - query with null', () {
```

blagoev (approved):

PR #873, opened 2022-08-25
Add support for frozen Realms

nirinchev:
Adds support for freezing Realms and entities:

```dart
final liveRealm = Realm(Configuration.local([...]));

final frozenRealm = liveRealm.freeze();
expect(frozenRealm.isFrozen, true);

final frozenResults = liveRealm.all<Foo>().freeze();
expect(frozenResults.isFrozen, true);

final frozenObject = liveRealm.find<Foo>(123).freeze();
expect(frozenObject.isFrozen, true);

final frozenList = liveRealm.find<Foo>(123).bars.freeze();
expect(frozenList.isFrozen, true);
```

Fixes https://github.com/realm/realm-dart/issues/56
Fixes https://github.com/realm/realm-dart/issues/54
Fixes https://github.com/realm/realm-dart/issues/55
Fixes https://github.com/realm/realm-dart/issues/808
Fixes https://github.com/realm/realm-dart/issues/807
Fixes https://github.com/realm/realm-dart/issues/806

### DIFF
--- a/generator/lib/src/realm_model_info.dart
+++ b/generator/lib/src/realm_model_info.dart
@@ -87,6 +87,10 @@ class RealmModelInfo {
       yield 'Stream<RealmObjectChanges<$name>> get changes => RealmObject.getChanges<$name>(this);';
       yield '';
 
+      yield '@override';
+      yield '$name freeze() => RealmObject.freezeObject<$name>(this);';
+      yield '';
+
       yield 'static SchemaObject get schema => _schema ??= _initSchema();';
       yield 'static SchemaObject? _schema;';
       yield 'static SchemaObject _initSchema() {';
--- a/lib/src/realm_class.dart
+++ b/lib/src/realm_class.dart
@@ -302,14 +302,13 @@ class Realm implements Finalizable {
   ///
   /// A frozen Realm is an immutable snapshot view of a particular version of a
   /// Realm's data. Unlike normal [Realm] instances, it does not live-update to
-  /// reflect writes made to the Realm, and can be accessed from any thread. Writing
-  /// to a frozen Realm is not allowed, and attempting to begin a write transaction
-  /// will throw an exception.
+  /// reflect writes made to the Realm. Writing to a frozen Realm is not allowed,
+  /// and attempting to begin a write transaction will throw an exception.
   ///
   /// All objects and collections read from a frozen Realm will also be frozen.
   ///
-  /// Note: Keeping a large number of frozen Realms with different versions alive can have a negative impact on the filesize
-  /// of the underlying database.
+  /// Note: Keeping a large number of frozen Realms with different versions alive can
+  /// have a negative impact on the file size of the underlying database.
   Realm freeze() {
     if (isFrozen) {
       return this;
@@ -433,6 +432,14 @@ extension RealmInternal on Realm {
   RealmMetadata get metadata => _metadata;
 
   T? resolveObject<T extends RealmObject>(T object) {
+    if (!object.isManaged) {
+      throw RealmStateError("Can't resolve unmanaged objects");
+    }
+
+    if (!object.isValid) {
+      throw RealmStateError("Can't resolve invalidated (deleted) objects");
+    }
+
     final handle = realmCore.resolveObject(object, this);
     if (handle == null) {
       return null;
--- a/lib/src/realm_class.dart
+++ b/lib/src/realm_class.dart
@@ -101,11 +101,16 @@ class Realm implements Finalizable {
   /// the schema will be read from the file.
   late final RealmSchema schema;
 
+  /// Gets a value indicating whether this [Realm] is frozen. Frozen Realms are immutable
+  /// and will not update when writes are made to the database.
+  late final bool isFrozen;
+
   /// Opens a `Realm` using a [Configuration] object.
   Realm(Configuration config) : this._(config);
 
   Realm._(this.config, [RealmHandle? handle]) : _handle = handle ?? _openRealm(config) {
     _populateMetadata();
+    isFrozen = realmCore.isFrozen(this);
   }
 
   static RealmHandle _openRealm(Configuration config) {
@@ -292,6 +297,26 @@ class Realm implements Finalizable {
 
   /// Deletes all [RealmObject]s of type `T` in the `Realm`
   void deleteAll<T extends RealmObject>() => deleteMany(all<T>());
+
+  /// Returns a frozen (immutable) snapshot of this Realm.
+  ///
+  /// A frozen Realm is an immutable snapshot view of a particular version of a
+  /// Realm's data. Unlike normal [Realm] instances, it does not live-update to
+  /// reflect writes made to the Realm, and can be accessed from any thread. Writing
+  /// to a frozen Realm is not allowed, and attempting to begin a write transaction
+  /// will throw an exception.
+  ///
+  /// All objects and collections read from a frozen Realm will also be frozen.
+  ///
+  /// Note: Keeping a large number of frozen Realms with different versions alive can have a negative impact on the filesize
+  /// of the underlying database.
+  Realm freeze() {
+    if (isFrozen) {
+      return this;
+    }
+
+    return Realm._(config, realmCore.freeze(this));
+  }
 
   SubscriptionSet? _subscriptions;
 
--- a/lib/src/realm_class.dart
+++ b/lib/src/realm_class.dart
@@ -431,6 +431,31 @@ extension RealmInternal on Realm {
   }
 
   RealmMetadata get metadata => _metadata;
+
+  T? resolveObject<T extends RealmObject>(T object) {
+    final handle = realmCore.resolveObject(object, this);
+    if (handle == null) {
+      return null;
+    }
+
+    final metadata = (object.accessor as RealmCoreAccessor).metadata;
+
+    return RealmObjectInternal.create(T, this, handle, RealmCoreAccessor(metadata)) as T;
+  }
+
+  RealmList<T>? resolveList<T extends Object?>(ManagedRealmList<T> list) {
+    final handle = realmCore.resolveList(list, this);
+    if (handle == null) {
+      return null;
+    }
+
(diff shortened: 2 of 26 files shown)


lib/src/realm_object.dart
```diff
@@ -309,10 +333,24 @@ mixin RealmObject on RealmEntity implements Finalizable {
...
 
+    if (invocation.isSetter) {
+      final name = _symbolRegex.firstMatch(invocation.memberName.toString())?.namedGroup("symbolName");
+      if (name == null) {
+        throw RealmError(
+            "Could not find symbol name for ${invocation.memberName}. This is likely a bug in the Realm SDK - please file an issue at https://github.com/realm/realm-dart/issues");
+      }
+
+      return set(this, name, invocation.positionalArguments.single);
+    }
```
nirinchev, line 344: I needed this for a test - in a follow-up PR we can add more tests and announce this as a supported functionality.

lib/src/realm_object.dart
```diff
@@ -309,10 +333,24 @@ mixin RealmObject on RealmEntity implements Finalizable {
...
+
     return super.noSuchMethod(invocation);
   }
 
+  /// An object exposing dynamic API for this [RealmObject] instance.
```
nirinchev, line 349: API docs were missing for a few methods - I figured it's easier to add them here rather than open a PR only for that.

lib/src/realm_object.dart
```diff
@@ -458,9 +496,9 @@ class DynamicRealmObject {
   /// Gets a list by the property name. If a generic type is specified, the property
   /// type will be validated against the type. Otherwise, a `List<Object>` will be
   /// returned.
-  List<T> getList<T extends Object?>(String name) {
+  RealmList<T> getList<T extends Object?>(String name) {
```
nirinchev, line 499: I think it was a mistake to expose this as `List`. It may be moot since @nielsenko plans to remove this API, but it was useful for now.

generator/lib/src/realm_model_info.dart
```diff
@@ -87,6 +87,10 @@ class RealmModelInfo {
       yield 'Stream<RealmObjectChanges<$name>> get changes => RealmObject.getChanges<$name>(this);';
       yield '';
 
+      yield '@override';
+      yield '$name freeze() => RealmObject.freezeObject<$name>(this);';
+      yield '';
```
nirinchev, line 92: The reason why we want to override this in the generated class is to specialize the returned value to match the object type.

nielsenko: I see a two options here to consider:
* `void freeze()` swizzle the existing realm and handle to frozen versions. Ie. change the object in place.
* `X get frozen` use property style (matches better with `dynamic`) and a bit shorter

Any thoughts on pros & cons?

lib/src/realm_object.dart
nielsenko: Yes, you loose the extension methods :-) ..

nielsenko (commented):
Would like a discussion on:
* `void freeze()` vs
* `X get frozen` vs
* `X freeze()`

But otherwise it looks fine

generator/lib/src/realm_model_info.dart
nirinchev: I don't like the property syntax as it looks misleadingly cheap. Not sure if that's the case for dart, but when I look at properties, I don't expect some complicated logic to be happening when I invoke them. It's not like freezing is super expensive, but it does a fair amount of allocations and I'd like this to be more explicit. It's not like we have the frozen version of the object lying around and you just access it through the property - you're performing some operation which is better conveyed with a method.

Regarding freezing in place - I mildly dislike it because it's less flexible. Let's say you have a method that accepts a RealmObject as an argument. You may want to obtain a frozen version of that object, but if it's an in-place operation, you'll modify your inputs and make the object unusable for the caller. Let's say you have data-bound an object to your UI, if you pass it to some method that freezes it, your databinding becomes unusable.

nielsenko: As I understand freezing is super cheap for core (@finnschiermer convinced me so in a video once 😄 ). Basically it has no upfront cost - like O(1) with extremely low constant factor. But it may make it difficult to reap storage from old versions later.

If we froze in-place it would literally just be:
```dart
    final frozenRealm = object.realm.freeze();
    final frozenHandle = realmCore.resolveObject(object, frozenRealm)!;
``` 
I think `freeze()` reads like a `void` method that changes the object in place. So, if we go with `freeze` I would prefer that semantic. Even though you need to create a new object if we go with the property `obj.frozen` then it is no more expensive than say indexing into a realm list.

nirinchev: I think Finn may be overselling the cheapness here. It may be super cheap for Core, but it's noticeably expensive for us, particularly when we don't have a cached frozen Realm at the version that we're at. You can see the amount of work that goes into `Realm::get_frozen_realm` both in terms of validations, but also copying structures around. While I haven't measured it for dart, I remember it was many times more expensive than resolving a link for .NET.

Re: freezing in place - I believe all SDKs are using the `freeze()` terminology and returning a new instance rather than freezing an existing instance in place. While we could use `getFrozen()` as an alternative, I'd rather not deviate from the terminology already adopted.

nielsenko:
```c++
SharedRealm Realm::get_frozen_realm(Config config, VersionID version)
{
    auto coordinator = RealmCoordinator::get_coordinator(config.path);
    SharedRealm realm = coordinator->get_realm(std::move(config), util::Optional<VersionID>(version));
    realm->set_auto_refresh(false);
    return realm;
}
```
don't smell expensive to me, but I would need to investigate more, and measure. 

Anyway, the argument that both kotlin, swift and dotnet adopted `T freeze()` is a winning one, but it leads me to another question. Should we capture in the type system that the object is frozen, ie. return a `FrozenT` instead of a `T`?

nirinchev: I think that may be a bit too annoying as we'd need to essentially shadow all our types + generate frozen version for the user types. This also means that a method accepting/being constrained on `RealmObject` won't work with `FrozenObject`. I'm open to being convinced otherwise, but I feel it's a fair amount of work that may end up being more confusing than helpful.

nielsenko: It would be significant work yes, but I'm a big fan of capturing constraints in the type system. However, let us land this PR separately, and discuss later.

nielsenko (approved):

blagoev:
@nielsenko Should we add a discussion topic in the agenda

test/realm_object_test.dart
```diff
@@ -602,4 +602,92 @@ Future<void> main([List<String>? args]) async {
...
+    final realm = getRealm(config);
+
+    final liveTeam = realm.write(() {
+      return realm.add(Team('team', players: [Person('Peter')], scores: [123]));
+    });
+    final frozenTeam = freezeObject(liveTeam);
+
+    expect(frozenTeam.isFrozen, true);
+    expect(frozenTeam.realm.isFrozen, true);
+    expect(frozenTeam.players.single.isFrozen, true);
```
desistefanova, line 617:
```suggestion
expect(frozenTeam.players.isFrozen, true);
expect(frozenTeam.players.single.isFrozen, true);
```

desistefanova (approved):
Missing update in CHANGELOG.md. I added a small suggestion in tests.

lib/src/realm_class.dart
```diff
@@ -406,6 +431,31 @@ extension RealmInternal on Realm {
   }
 
   RealmMetadata get metadata => _metadata;
+
+  T? resolveObject<T extends RealmObject>(T object) {
+    final handle = realmCore.resolveObject(object, this);
```
blagoev, line 436: Should we check if the object is managed?

blagoev: there might be a performance benefit at least

blagoev: also it will guard the `object.accessor as RealmCoreAccessor` below

lib/src/results.dart
```diff
@@ -23,22 +23,21 @@ import 'dart:ffi';
 import 'collections.dart';
 import 'native/realm_core.dart';
 import 'realm_class.dart';
-import 'realm_object.dart' show RealmObjectMetadata;
+import 'realm_object.dart' show RealmObjectMetadata, RealmEntityInternal;
```
blagoev, line 26: That's not right. We should not be exposing `Internal` classes.

test/realm_object_test.dart
blagoev: @nirinchev ping

blagoev (changes requested):
This one should not be exposed `RealmEntityInternal`. Added some comments

lib/src/realm_class.dart
```diff
@@ -293,6 +298,26 @@ class Realm implements Finalizable {
   /// Deletes all [RealmObject]s of type `T` in the `Realm`
   void deleteAll<T extends RealmObject>() => deleteMany(all<T>());
 
+  /// Returns a frozen (immutable) snapshot of this Realm.
+  ///
+  /// A frozen Realm is an immutable snapshot view of a particular version of a
+  /// Realm's data. Unlike normal [Realm] instances, it does not live-update to
+  /// reflect writes made to the Realm, and can be accessed from any thread. Writing
```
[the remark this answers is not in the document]
nirinchev: Oops, didn't reread after copy pasting 😬

lib/src/results.dart
nirinchev: Not sure how that got there - will investigate.

nirinchev: Rereading this, we're not exposing the internal class, we're importing it.

nirinchev: I removed the `show ...` clause as it seemed redundant.

blagoev: oh I misread that line.

blagoev (approved):
missing changelog entry

PR #875, opened 2022-08-29
Support indexes on all compatible types

### LINKED ISSUE #797: Timestamp indexes (are there other indexes we miss?)

nielsenko:
Allow (and correctly apply) `@Indexed()` on `int, bool, String, ObjectId, Uuid,` and `DateTime` (and the nullable version as well).
```dart
@RealmModel()
class _Indexable {
  @Indexed()
  late bool aBool;
  @Indexed()
  bool? aNullableBool;
  @Indexed()
  late int anInt;
  @Indexed()
  int? aNullableInt;
  @Indexed()
  late String aString;
  @Indexed()
  String? aNullableString;
  @Indexed()
  late ObjectId anObjectId;
  @Indexed()
  ObjectId? aNullableObjectId;
  @Indexed()
  late Uuid anUuid;
  @Indexed()
  Uuid? aNullableUuid;
  @Indexed()
  late DateTime aDateTime;
  @Indexed()
  DateTime? aNullableDateTime;
}
```
Fix #797,


desistefanova (commented):
Missing an update in Changelog.

common/lib/src/realm_types.dart
```diff
@@ -49,30 +51,30 @@ class Mapping<T> {
   List<T?> getListOfNullables(RealmAccessorMarker accessor, RealmObjectMarker object, String propertyName) => accessor.getList<T?>(object, propertyName);
 }
 
-const _intMapping = Mapping<int>();
-const _boolMapping = Mapping<bool>();
+const _intMapping = Mapping<int>(indexable: true);
+const _boolMapping = Mapping<bool>(indexable: true); // oddly enough :-)
```
desistefanova, line 55: Let's remove the comment :)

generator/lib/src/error.dart
```diff
@@ -43,7 +43,7 @@ class RealmInvalidGenerationSourceError extends InvalidGenerationSourceError {
         color = color ?? session.color,
         super(message, todo: todo, element: element) {
     if (element is FieldElement || element is ConstructorElement) {
-      final classElement = element.enclosingElement!;
+      final classElement = element.enclosingElement3!;
```
desistefanova, line 46: Maybe should be element2.

generator/lib/src/field_element_ex.dart
```diff
@@ -128,15 +127,17 @@ extension FieldElementEx on FieldElement {
...
 
-      // Validate indexes
-      if ((primaryKey != null || indexed != null) &&
-          (![RealmPropertyType.string, RealmPropertyType.int, RealmPropertyType.objectid, RealmPropertyType.uuid].contains(type.realmType) ||
-              type.isRealmCollection)) {
+      // Validate primary keys and indexes
+      if ((primaryKey != null || indexed != null) && !(type.realmType?.mapping.indexable ?? false)) {
         final file = span!.file;
         final annotation = (primaryKey ?? indexed)!.annotation;
+        final listOfIndexableTypes = RealmPropertyType.values //
```
desistefanova, line 134:
```suggestion
        final listOfIndexableTypes = RealmPropertyType.values
```

generator/lib/src/field_element_ex.dart
```diff
@@ -128,15 +127,17 @@ extension FieldElementEx on FieldElement {
         */
       }
 
-      // Validate indexes
-      if ((primaryKey != null || indexed != null) &&
-          (![RealmPropertyType.string, RealmPropertyType.int, RealmPropertyType.objectid, RealmPropertyType.uuid].contains(type.realmType) ||
-              type.isRealmCollection)) {
+      // Validate primary keys and indexes
+      if ((primaryKey != null || indexed != null) && !(type.realmType?.mapping.indexable ?? false)) {
```
desistefanova, line 131:
```suggestion
      if ((primaryKey != null || indexed != null) && (type.realmType?.mapping.indexable ?? false)) {
```

generator/lib/src/dart_type_ex.dart
```diff
@@ -32,7 +32,7 @@ extension DartTypeEx on DartType {
   bool get isRealmAny => const TypeChecker.fromRuntime(RealmAny).isAssignableFromType(this);
   bool get isRealmBacklink => false; // TODO: Implement Backlink support https://github.com/realm/realm-dart/issues/693
   bool get isRealmCollection => realmCollectionType != RealmCollectionType.none;
-  bool get isRealmModel => element != null ? realmModelChecker.annotationsOfExact(element!).isNotEmpty : false;
+  bool get isRealmModel => element2 != null ? realmModelChecker.annotationsOfExact(element2!).isNotEmpty : false;
```
desistefanova, line 35: Chech if `override element2` exists an remove it.

generator/lib/src/pseudo_type.dart
```diff
@@ -52,5 +52,5 @@ class PseudoType extends TypeImpl {
   }
 
   @override
-  Element? get element2 => throw UnimplementedError();
+  Element? get element2 => null;
```
desistefanova, line 55: Why do we need `override element2` that returns null?

generator/test/test_util.dart
```diff
@@ -180,36 +179,27 @@ class LinesEqualsMatcher extends Matcher {
...
   }
-  return LineSplitter.split(content).join('\n');
-}
 
-String _stringReplacements(String content) {
-  final formatter = DartFormatter();
-  var lines = LineSplitter.split(content);
-  String formattedContent = lines.where((element) => !element.startsWith("part of")).join('\n');
-  return formatter.format(formattedContent);
+  Future<String> readFileAsDartCode() async {
```
desistefanova, line 198:
```suggestion
  Future<String> readAsDartCode() async {
```

desistefanova, line 198: I added a suggestion to change the name of the method, because it could be used also for formatting any string content, not only from file.

lib/src/realm_property.dart
```diff
@@ -34,6 +34,9 @@ abstract class SchemaProperty {
   /// `true` if the property is a primary key.
   bool get primaryKey;
 
+  /// `true` if the property is a primary key.
```
desistefanova, line 37:
```suggestion
  /// `true` if the property is an index.
```

desistefanova (commented):
Added some comments and suggestions.

generator/lib/src/error.dart
nielsenko: No - `enclosingElement3` is correct. `enclosingElement2` ids deprecated

common/lib/src/realm_types.dart
nielsenko: So strict 😄 .. but it is odd. Anyway, I will remove

generator/lib/src/field_element_ex.dart
nielsenko: On purpose. Impacts formatting.

generator/lib/src/dart_type_ex.dart
nielsenko: Elaborate?

generator/test/test_util.dart
nielsenko: This is named to follow convention. LIke `readFileAsString`.

lib/src/realm_property.dart
nielsenko: Changing to 
```dart
/// `true` if the property is indexed.
```

generator/lib/src/field_element_ex.dart
desistefanova: Oh, I see. It's my mistake.

generator/lib/src/dart_type_ex.dart
desistefanova: Hah, this was a remark for me. I think a deleted this comment. Sorry. :)

generator/lib/src/field_element_ex.dart
desistefanova: Isn't it correct to remove the `!` in front of the second condition?

generator/lib/src/field_element_ex.dart
```diff
@@ -133,14 +133,16 @@ extension FieldElementEx on FieldElement {
...
-      if ((primaryKey != null || indexed != null) &&
-          (![RealmPropertyType.string, RealmPropertyType.int, RealmPropertyType.objectid, RealmPropertyType.uuid].contains(type.realmType) ||
-              type.isRealmCollection)) {
+      if ((primaryKey != null || indexed != null) && !(type.realmType?.mapping.indexable ?? false)) {
         final file = span!.file;
         final annotation = (primaryKey ?? indexed)!.annotation;
+        final listOfIndexableTypes = RealmPropertyType.values //
+            .map((t) => t.mapping)
+            .where((m) => m.indexable) //
+            .expand((m) => [m.type, m.nullableType]);
```
nirinchev, line 142: This will likely make the error message fairly long - do we need to explicitly list the nullable types? Can't we change the error message to something like `Realm only supports indexes on fields of type int, bool, ... as well as their nullable versions`?

generator/lib/src/field_element_ex.dart
```diff
@@ -133,14 +133,16 @@ extension FieldElementEx on FieldElement {
...
 
         throw RealmInvalidGenerationSourceError(
-          'Realm only support indexes on String, int, and bool fields',
+          'Realm only support indexes on fields of type\n${listOfIndexableTypes.join(', ')}',
```
nirinchev, line 145:
```suggestion
          'Realm only supports indexes on fields of type\n${listOfIndexableTypes.join(', ')}',
```

nirinchev (commented):
Not sure if it's worth writing a test for this, but we could try to prove that the index exists in Core and is being used by opening 2 realms - one with indexed fields, the other with non-indexed fields, then adding 10000 objects and querying for a single object. The test should expect that the query utilizing the index completes much faster.

generator/lib/src/field_element_ex.dart
nielsenko: Fixed

nielsenko:
> Not sure if it's worth writing a test for this, but we could try to prove that the index exists in Core and is being used by opening 2 realms - one with indexed fields, the other with non-indexed fields, then adding 10000 objects and querying for a single object. The test should expect that the query utilizing the index completes much faster.

Added test of 1.000 lookups against 100.000 objects and compared indexed vs not-indexed for all indexable types, and assert the speedup is at least 10 fold.

test/indexed_test.dart
```diff
@@ -0,0 +1,199 @@
...
+    expect(allNotIndexed.length, max);
+
+    // Inefficient, but fast enough for this test
+    final searchOrder = (List.generate(max, (i) => i)..shuffle(Random(42))).take(1000).toList();
+
+    QueryBuilder<T, U> builder<T extends RealmObject, U extends Object>(RealmResults<T> results, String fieldName) {
+      return (U value) => results.query('$fieldName == \$0', [value]);
+    }
+
+    @pragma('vm:no-interrupts')
```
blagoev, line 147: Why do we need this. Shouldn't we be testing in real world scenarios. If this is needed perhaps it makes sense to relax the expectation a bit instead of trying too hard to hit the current ones

test/indexed_test.dart
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    void compareSpeed<U extends Object>(
+      String fieldName,
+      U Function(int index) indexToValue,
+    ) {
+      final lookupCount = searchOrder.length;
+
+      display(Type type, Duration duration) {
+        print('$lookupCount lookups of ${'$type'.padRight(12)} on ${fieldName.padRight(10)} : ${duration.inMicroseconds ~/ lookupCount} us/lookup');
```
blagoev, line 178: Do we need to print every time. Why not do that only if something is off.

common/lib/src/realm_types.dart
```diff
@@ -21,32 +21,46 @@ import 'dart:typed_data';
...
+  // Types
+  Type get type => T;
+  Type get nullableType => _typeOf<T?>();
+}
+
+const _intMapping = Mapping<int>(indexable: true);
+const _boolMapping = Mapping<bool>(indexable: true);
+
+/// @nodoc
 enum RealmPropertyType {
```
blagoev, line 41: RealmPropertyType is exposed and should have an API docs.

common/lib/src/realm_types.dart
```diff
@@ -21,32 +21,46 @@ import 'dart:typed_data';
...
-  // ignore: unused_field, constant_identifier_names
-  _13,
+  _13, // ignore: unused_field, constant_identifier_names
   linkingObjects,
-  objectid,
-  // ignore: unused_field, constant_identifier_names
-  _16,
-  uuid,
+  objectid(Mapping<ObjectId>(indexable: true)),
+  _16, // ignore: unused_field, constant_identifier_names
```
blagoev, line 58: Could we use the opportunity to remove these fields?

test/indexed_test.dart
```diff
@@ -0,0 +1,199 @@
...
+import 'dart:typed_data';
+
+import 'package:collection/collection.dart';
+import 'package:test/test.dart';
+
+import '../lib/realm.dart';
+
+part 'indexed_test.g.dart';
+
+// IMPORTANT: Don't import our own test.dart here. It will break AOT compilation!
```
blagoev, line 29: Why do we need AOT compilation for these tests? Where do we invoke it?

test/indexed_test.dart
```diff
@@ -0,0 +1,199 @@
...
+
+// Realm models cannot extend other classes, but they can implement interfaces.
+class Base {
```
blagoev, line 32: Why do we need this here. If we want to test this out it seems we need to make this test part of `realm_object.dart`. Here it seems irrelevant since its never used.

blagoev (commented):
Its good, I have couple of questions.

generator/test/error_test_data/bool_not_allowed_as_primary_key.dart
```diff
@@ -1,10 +0,0 @@
-import 'package:realm_common/realm_common.dart';
-
-//part 'bool_not_allowed_as_primary_key.g.dart';
-
-@RealmModel()
-@MapTo('Bad')
-class _Foo {
-  @PrimaryKey()
-  late bool bad;
```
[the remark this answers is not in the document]
nielsenko: My mistake. I assumed that since bool are now indexable they could also be used for primary keys. But you are right they cannot (did a test to validate) so I have re-instituted this test.

test/indexed_test.dart
nielsenko: Are you. asking about the `@pragma(vm:no-interrupts)`? Then https://mrale.ph/blog/2021/01/21/microbenchmarking-dart-part-1.html is a good place start.

Since this it a timing call, I rather not have the GC disturb us.

test/indexed_test.dart
nielsenko: No we don't need to

common/lib/src/realm_types.dart
nielsenko: The `_16` etc.?

test/indexed_test.dart
nielsenko: For CI it doesn't matter, but when working on improving performance, it is nice to be able to do
```sh
dart compile exe test/indexed_test.dart
```

test/indexed_test.dart
nielsenko: We no longer do. This is some residue from an earlier approach. I will remove `Base`

test/indexed_test.dart
blagoev: Yes I know this post. I am asking about that yes.

common/lib/src/realm_types.dart
blagoev: yep

test/indexed_test.dart
blagoev: how can we improve performance here. Isn't it a Core work.  I suggest we don't restrict ourselves with this limitation

common/lib/src/realm_types.dart
blagoev: if you are afraid this will introduce too much code churn lets not do this now but create an issue for it for later.

test/indexed_test.dart
nielsenko: Done

test/indexed_test.dart
blagoev: Why not relax the test expectaions instead of doing it like this. Hence we will be capturing and expecting on a real world usage.

test/indexed_test.dart
nielsenko: Changed it so measurements are only displayed on test failure

common/lib/src/realm_types.dart
nielsenko: The only reason it is exposed is because we use it in the generated classes, but I will add API docs back.

nielsenko: Done

test/indexed_test.dart
nielsenko: Why are you worried about this pragma? I don't understand.

I have relaxed the expected speedup. If this test ever fails on CI, something is really amiss.

test/indexed_test.dart
nielsenko: Yes, this is just a test. It has already exposed that indexes on timestamps are not very beneficial currently.

test/indexed_test.dart
blagoev: It's just not running the test in a real world usage. I don't expect many will use that in their code.

common/lib/src/realm_types.dart
blagoev: sure

test/indexed_test.dart
blagoev: can we at least have a comment why do we need AOT compilation in here. Just so it is more clear what we are trying to achieve. 
Something like 
`We may use AOT compilation locally to run the performance tests in this file`

test/indexed_test.dart
nielsenko: This is not a real world usage. I'm timing indexed vs non-indexed fields to ensure we see a difference in performance when doing `field == $0` queries, ie. that we didn't somehow forget to communicate the index to core or whatever, or that the index handling in core regressed.

For that to be less brittle, I would like to avoid GC while it runs, hence `@pragma(vm:no-interrupts)`. It is only applied to the `measureSpeed` function - I fail to see why that is controversial.

test/indexed_test.dart
nielsenko: Sure, I'll elaborate on the comment. We should probably also figure out why it breaks AOT compilation of the tests to include our own `test.dart` eventually.

nielsenko:
```dart
// Don't import our own test.dart here. It will break AOT compilation.
// We may use AOT compilation locally to manually run the performance
// tests in this file
```

test/indexed_test.dart
```diff
@@ -0,0 +1,176 @@
...
+        objectIdFactory(i),
+        uuidFactory(i),
+      ),
+    );
+    realm.write(() => realm.addAll(notIndexed));
+
+    expect(allNotIndexed.length, max);
+
+    // Inefficient, but fast enough for this test
+    final searchOrder = (List.generate(max, (i) => i)..shuffle(Random(42))).take(1000).toList();
```
nirinchev, line 116: Why not

```suggestion
    final random = Random(42);
    final searchOrder = List.generate(1000, i => random.nextInt(max));
```

I know perf doesn't matter that much, it's just a little easier to read.

test/indexed_test.dart
blagoev: ok resolving, I feel not worth to discuss more

blagoev (approved):

test/indexed_test.dart
```diff
@@ -0,0 +1,176 @@
...
+      display(Type type, Duration duration) {
+        print('$lookupCount lookups of ${'$type'.padRight(12)} on ${fieldName.padRight(10)} : ${duration.inMicroseconds ~/ lookupCount} us/lookup');
+      }
+
+      final indexedTime = measureSpeed(allIndexed, fieldName, indexToValue);
+      final notIndexedTime = measureSpeed(allNotIndexed, fieldName, indexToValue);
+      try {
+        // skip timestamp for now, as timestamps are not indexed properly it seems
+        if (fieldName != 'timestamp') {
+          expect(indexedTime, lessThan(notIndexedTime)); // indexed should be faster
```
nirinchev, line 161: I think we should be stricter here - something like `expect(indexedTime, lessThan(0.5 * notIndexedTime))`

nirinchev (approved):
Small nits for the tests, otherwise looks good.

test/indexed_test.dart
nielsenko: Problem is it can create duplicates. That is what I avoid above

test/indexed_test.dart
nielsenko: I used to have 
```dart
expect(indexedTime * 10, lessThan(notIndexedTime));
```
but I was asked to loosen it

nielsenko: .. but maybe 2 would be fine 😄

test/indexed_test.dart
nirinchev: It can, but do we care?

PR #876, opened 2022-08-29
Add handle ownership again

### LINKED ISSUE #527: Introduce ownership semantics for our handles
The `RealmHandle` should have the ability to own other handles and close them if necessary. Examples:

1. Session handle - otherwise the session may outlive the Realm preventing deletion.
2. SubscriptionSetHandle - same as 1.
3. Object/List/Results handles when the Realm is unowned, e.g. migrations, initial data.

We can look at the .NET handle logic for inspiration.

### LINKED ISSUE #837: Handle ownership
The `RealmHandle` should have the ability to own other handles and close them if necessary. Examples:

1. Session handle - otherwise the session may outlive the Realm preventing deletion.
1. SubscriptionSetHandle - same as 1.
1. Object/List/Results handles when the Realm is unowned, e.g. migrations, initial data.

We can look at the .NET handle logic for inspiration.

nirinchev:
This is a refactoring of our native handle hierarchy. A summary of the changes:

* Introduces a method to release the native pointer to `HandleBase<T>`
  * The unowned version of the handle will not attach a native finalizer and the `release` method is a no-op.
* Adds `RootedHandleBase<T>` that is owned by a Realm and it's lifetime is at most as long as the lifetime of the parent Realm.
* Modifies all the realm types (list, results, object, query) to inherit from `RootedHandleBase ` and modifies their ctors to optionally pass in their parent Realm as root in case it's unowned.
  * The reason we only care about the unowned case is because that's a situation where the user doesn't have explicit control over the lifetime of the Realm instance.
* `RealmHandle` cleans up all its children when released.

Supersedes https://github.com/realm/realm-dart/pull/528. I'd like to add more meaningful tests but those need to wait until the migrations PR is in.

Fixes #527
Fixes #837


test/list_test.dart
```diff
@@ -215,43 +215,6 @@ Future<void> main([List<String>? args]) async {
...
-    //Ensure teams and player are in realm
-    var teams = realm.all<Team>();
-    expect(teams.length, 1);
-    expect(teams[0].players, isNotNull);
-    expect(teams[0].players.length, 1);
-
-    var players = teams[0].players;
-
-    realm.close();
-    expect(
```
nielsenko, line 238: Should we still throw when writing on a closed realm?

lib/src/native/realm_core.dart
```diff
@@ -1811,59 +1870,64 @@ class ConfigHandle extends HandleBase<realm_config> {
 }
 
 class RealmHandle extends HandleBase<shared_realm> {
+  int _counter = 0;
+
+  final Map<int, WeakReference<RootedHandleBase>> _children = {};
```
nielsenko, line 1875: Could we use an `Expando` here instead?

nielsenko (approved):
I think we can get rid of the ints by using an `Expando`. Might be worth exploring. But otherwise LGTM.

lib/src/native/realm_core.dart
nirinchev: I don't see how - we need to be able to iterate over all keys of the collection and `Expando` doesn't expose an API to do that.

lib/src/native/realm_core.dart
```diff
@@ -1787,19 +1802,63 @@ final _nativeFinalizer = NativeFinalizer(_realmLib.addresses.realm_release);
...
+    }
+  }
+}
+
+abstract class RootedHandleBase<T extends NativeType> extends HandleBase<T> {
+  final RealmHandle? _root;
+  int? _id;
+
+  RootedHandleBase(this._root, Pointer<T> pointer, int size) : super(pointer, size) {
+    _id = _root?.addChild(this);
```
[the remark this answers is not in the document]
nirinchev: I'm not sure I see how. The reason for the int is to avoid a situation where the root handle holds a reference to its children. If we keep the current design where the children are kept as weak references, then we need to iterate all children to try and find the one being removed, which seems more inefficient.

lib/src/native/realm_core.dart
```diff
@@ -1787,19 +1802,63 @@ final _nativeFinalizer = NativeFinalizer(_realmLib.addresses.realm_release);
...
+  /// A method that will be invoked just before the handle is released. Allows to cleanup
+  /// any custom data that inheritors are storing.
+  void _releaseCore() {}
+
+  void release() {
+    if (released) {
+      return;
+    }
+
+    _releaseCore();
```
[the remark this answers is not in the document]
nirinchev: We don't need to run prior to the native finalizer. This is only intended as a cleanup step to prevent the list of children from growing endlessly. I added a managed finalizer based on your suggestion - LMK if that's what you had in mind.

lib/src/native/realm_core.dart
```diff
@@ -1811,59 +1888,64 @@ class ConfigHandle extends HandleBase<realm_config> {
 }
 
 class RealmHandle extends HandleBase<shared_realm> {
+  int _counter = 0;
+
+  final Map<int, WeakReference<RootedHandleBase>> _children = {};
```
desistefanova, line 1893: I'm wondering why do we need a collection here `final Map<int, WeakReference<RootedHandleBase>> _children = {};`. Could you help me to understand?

desistefanova: I see. You keep the handles to all the objects from a specific realm into the RealmHandle. Then you know what to release while releasing the realm.

desistefanova: But doesn't this mean that:
- If I have a RealmOdject in a Realm;
- then I delete this RealmOdject;
- its handle will still be kept into RealmHandle
- RealmOdject's handle won't be released until the Realm is released.
Is that correct?

test/configuration_test.dart
```diff
@@ -375,6 +375,26 @@ Future<void> main([List<String>? args]) async {
...
+    late Person george;
+    final config = Configuration.local([Person.schema], initialDataCallback: (realm) {
+      george = realm.add(Person('George'));
+      people = realm.all<Person>();
+      expect(people.length, 1);
+    });
+
+    final realm = getRealm(config);
+
+    expect(() => people.length, throws<RealmClosedError>());
```
desistefanova, line 390: Maybe have to check also `people.isManaged` and `people.isValid`.

lib/src/native/realm_core.dart
nirinchev: We're keeping a map of weak references, so if your object goes out of scope and gets garbage collected, then its handle should be released whenever that happens. If it doesn't, then we'll release the handle whenever the Realm is closed.

lib/src/native/realm_core.dart
```diff
@@ -1811,59 +1888,64 @@ class ConfigHandle extends HandleBase<realm_config> {
...
+
   RealmHandle._(Pointer<shared_realm> pointer) : super(pointer, 24);
 
   RealmHandle._unowned(Pointer<shared_realm> pointer) : super.unowned(pointer);
+
+  int addChild(RootedHandleBase child) {
+    final id = _counter++;
```
desistefanova, line 1900: Why not to keep these handles only in case the root is unowned `if (this.isUnowned)`. I think we musn't keep the RealmObjects pointers for Realms that are not unowned.

lib/src/native/realm_core.dart
desistefanova: But is it going to be garbage collected if we keep its handle into a realm that is still not out of the scope?

nirinchev: Why wouldn't it be? There's nothing keeping a reference to the object handle.

lib/src/native/realm_core.dart
nirinchev: We already do that - e.g.:

https://github.com/realm/realm-dart/blob/19107dca53780a2806e148852fd963b6574fe58d/lib/src/native/realm_core.dart#L1924

The object/result/list handle will set their `root` to `null` if the Realm is not unowned.

lib/src/native/realm_core.dart
desistefanova: Aren't they still in the `children ` collection in the `RealmHandle` of the Realm? It is just interesting to me. It won't be a problem since you are releasing the realm when the callback (`initialDataCallback` for example) completes.

nirinchev: Those are weak references, so they should not be counted when the GC runs.

lib/src/native/realm_core.dart
```diff
@@ -1787,19 +1802,81 @@ final _nativeFinalizer = NativeFinalizer(_realmLib.addresses.realm_release);
...
+    if (_enableFinalizerTrace) {
+      _setupFinalizationTrace(this, _pointer);
+    }
   }
 
-  HandleBase.unowned(this._pointer);
+  HandleBase.unowned(this._pointer) : isUnowned = true;
 
   @override
   String toString() => "${_pointer.toString()} value=${_pointer.cast<IntPtr>().value}";
```
blagoev, line 1821: Should we output here if the handle is `unowned`

lib/src/native/realm_core.dart
```diff
@@ -1811,59 +1888,64 @@ class ConfigHandle extends HandleBase<realm_config> {
...
+    return id;
+  }
+
+  void removeChild(int id) {
+    _children.remove(id);
+  }
+
+  @override
+  void _releaseCore() {
+    final keys = _children.keys.toList();
```
blagoev, line 1911: Do we need the `toList` here. Can't we do a `.forEach`. or iterate over the `_children` collection

lib/src/list.dart
```diff
@@ -55,7 +55,7 @@ class ManagedRealmList<T extends Object?> extends collection.ListBase<T> with Re
   }
 
   @override
-  int get length => realmCore.getListSize(_handle);
+  int get length => realmCore.getListSize(handle);
 
   @override
```
blagoev, line 61: empty line

blagoev (approved):

desistefanova (commented):
Do you think we may say something in the changelog about this change?

desistefanova (changes requested):
There are many places from the existing code that require two arguments, but only one is passed. Is it only on my local machine or it is because of the merge?
Please make a search for the following constructors:
RealmResultsHandle._
RealmObjectHandle._
RealmListHandle._

nirinchev:
> Do you think we may say something in the changelog about this change?

This should be purely internal, but I can add something to the internal section.

> There are many places from the existing code that require two arguments, but only one is passed.

Yeah, that happened when I merged from master via the Github UI. Will fix it before merging.

desistefanova (approved):

desistefanova:
Maybe this PR is the correct place to remove all the `keepAlive` methods. I think we don't need them anymore. @nielsenko ?

nirinchev:
The `keepAlive` is necessary to make sure we don't get parts of the objects finalized mid-scope. I don't think this PR does anything that would allow us to remove them.

CHANGELOG.md
```diff
@@ -71,6 +71,7 @@
 
 ### Internal
 * Uses Realm Core v12.9.0
+* Added tracking of child handles for objects/results/lists obtained from an unowned Realm. This ensures that all children are invalidate as soon as the parent Realm gets released at the end of the callback. (Issue [#527](https://github.com/realm/realm-dart/issues/527))
```
desistefanova, line 74:
```suggestion
* Added tracking of child handles for objects/results/lists obtained from an unowned Realm. This ensures that all children are invalidated as soon as the parent Realm gets released at the end of the callback. (Issue [#527](https://github.com/realm/realm-dart/issues/527))
```

nirinchev: Need to install a proofreader in VS Code 😅

Issue #877, opened 2022-08-31
Allow realm DynamicLibrary to be loaded from a different path than the current directory


nsivraj:
### Description

1) Goals - Allow realm DynamicLibrary to be loaded from a different path than the current directory
2) Expected results - When running a dart compiled exe allow the realm DynamicLibrary to be loaded from a different path than the current directory
3) Actual results - When running a dart compiled exe from the command line the librealm_dart.dylib library cannot be found
4) Steps to reproduce - Try to run a dart compiled exe which uses the realm DynamicLibrary from a directory other than the one it was compiled in and it fails with the error: Invalid argument(s): Failed to load dynamic library '/Users/normanjarvis/forge/personal/code/coderockit/cr-e2e-tests/binary/macos/librealm_dart.dylib': dlopen(/Users/normanjarvis/forge/personal/code/coderockit/cr-e2e-tests/binary/macos/librealm_dart.dylib, 0x0001): tried: '/Users/normanjarvis/forge/personal/code/coderockit/cr-e2e-tests/binary/macos/librealm_dart.dylib' (no such file)
5) Code sample that highlights the issue (full Flutter or Dart project that we can run ourselves are ideal) - Please see the code here in this github repo https://gitlab.com/coderockit/cr-cli
6) Version of Realm (0.4.0+beta), Flutter(3.0.3) and Dart (2.17.5)
7) Version of desktop OS - Mac, Windows or Linux - Mac macOS Monterey Version 12.4 Apple M1 Max
8) Version of target mobile OS - this is a command line tool for a command prompt

### How important is this improvement for you?

Dealbreaker

blagoev:
Hi, We would like to understand the problem more. 
Right now if you compile a dart executable and you copy it to a different directory than the one it was compiled from, the Dart app will continue to work but as the error says, you need to copy the `binary` directory containing the realm binary as well next to the executable. 
So you will have this folder structure
```
/Users/normanjarvis/forge/personal/code/coderockit/cr-e2e-tests/myapp   <--- the dart executable
/Users/normanjarvis/forge/personal/code/coderockit/cr-e2e-tests/binary/macos/librealm_dart.dylib <---- the realm binary
```
We would like to understand the reasons why you can not do that. Is that not working for you?

nielsenko:
We grok the problem. See comments on: #877 and #881

nielsenko:
[...]

PR #878, opened 2022-08-31
Issue 877: Allow the realm DynamicLibrary to be opened from a path defined from …

nsivraj:
…an environment variable or dart define

https://github.com/realm/realm-dart/issues/877

### DIFF
--- a/flutter/realm_flutter/pubspec.yaml
+++ b/flutter/realm_flutter/pubspec.yaml
@@ -6,7 +6,7 @@ homepage: https://www.realm.io
 repository: https://github.com/realm/realm-dart
 issue_tracker: https://github.com/realm/realm-dart/issues
 
-publish_to: none
+# publish_to: none
 
 environment:
   sdk: ">=2.17.0 <3.0.0"
--- a/flutter/realm_flutter/pubspec.yaml
+++ b/flutter/realm_flutter/pubspec.yaml
@@ -34,6 +34,7 @@ dependencies:
   tar: ^0.5.4
   build_runner: ^2.1.0
   http: ^0.13.4
+  dylib: any
 
 dev_dependencies:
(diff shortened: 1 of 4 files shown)


nsivraj:
@cla-bot check

nsivraj:
This code is working for me in my command line dart compiled exe on my M1 mac. I have a few more tests to run to make sure it works correctly on linux and windows.

nsivraj:
> This code is working for me in my command line dart compiled exe on my M1 mac. I have a few more tests to run to make sure it works correctly on linux and windows.

I have now verified that this works on windows, linux, and macos.

nielsenko:
Hi Norman

Thanks for bringing this to our attention. I have merged your PR to a CI integration branch in order to run the full test suite (#882), and it passes fine. However, I have some quarrels merging this to master. 

I'm not too fond of the `dylib` package used, I would rather play nice with `DYLD_LIBRARY_PATH`, `LD_LIBRARY_PATH`, etc. instead of introducing this `LIBREALM_DART_PATH` environment variable (/ dart define).

I have created another PR: #881 that I would encourage you to look at. It basically changes the dlopen strategy to:

1. First try calling dlopen directly, allowing user to influence searched paths using `DYLD_LIBRARY_PATH`, `LD_LIBRARY_PATH`, etc depending on platform
2. Try to load it form the same directory as the executable, ie. if you compile to an exe you can place the realm lib next to it, and it will work.
3. Try to find a root project from either the placement of the script (Platform.script) and resolve `binary/<platform>/<realm_lib>` from there (ie. load `binary/macos/librealm_dart.so` on macos, etc.)
4. Do what we used to.

I wonder if that would not solve the issue you reported for you as well? Or is the custom environment variable and the dart define important for you?

PR #884, opened 2022-09-05
Fix removeAt

### LINKED ISSUE #883: [Bug]: `List.removeAt` doesn't work
### What happened?

when use list.remove data cannot be deleted
data  should be deleted.

### Repro steps

```
FutureOr<void> removeCoordinatesImage(CoordPoint point, int index) async {
    var image = File(point.imagePaths[index]);
    if (image.existsSync()) image.delete();
    point.realm.write(() {
      point.imagePaths.removeAt(index);
    });
  }
```
imagePaths.length will not change
imagePaths type is `List<String>`

### Version

0.4.0-beta

### What Realm SDK flavor are you using?

Local Database only

Flutter Application

### Client OS and version

Android 10

### Code snippets

_No response_

_No response_

### Relevant log output

_No response_

nielsenko:
Apparently we never wired up and tested `removeAt`.

This fixes #883

### DIFF
--- a/test/list_test.dart
+++ b/test/list_test.dart
@@ -18,6 +18,8 @@
 
 // ignore_for_file: unused_local_variable
 
+import 'dart:math';
+
 import 'package:test/test.dart' hide test, throws;
 import '../lib/realm.dart';
 import 'test.dart';
@@ -362,7 +364,7 @@ Future<void> main([List<String>? args]) async {
     expect(result, [person]);
   });
 
-  test('List removeAt', () {
+  test('ManagedRealmList.removeAt', () {
     final config = Configuration.local([Team.schema, Person.schema]);
     final realm = getRealm(config);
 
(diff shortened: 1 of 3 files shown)


nirinchev:
Isn't `removeAt` also implemented in `ListMixin`? The issue is that their implementation wouldn't work with Realm (the same way `removeAt` doesn't work) because they're manipulating the length of the list.

test/list_test.dart
```diff
@@ -361,4 +361,27 @@ Future<void> main([List<String>? args]) async {
...
+    expect(team.players, [alice, bob, carol, dan]);
+
+    expect(realm.write(() => team.players.removeAt(2)), carol);
+    //expect(team.players.length, 3);
+    expect(team.players, [alice, bob, dan]);
+
+    expect(realm.write(() => team.players.removeAt(0)), alice);
+    expect(team.players.length, 2);
+    expect(team.players, [bob, dan]);
+  });
```
desistefanova, line 386: We need a check that removed items still exist in the ream, detached from the list. Similar to the tests for Clear https://github.com/realm/realm-dart/blob/77e730551b01a12c49c47367b16090a47942a557/test/list_test.dart#L98

nielsenko:
> Isn't `removeAt` also implemented in `ListMixin`? The issue is that their implementation wouldn't work with Realm (the same way `removeAt` doesn't work) because they're manipulating the length of the list.

Yes. I have implemented `length` setter to truncate, if `newLength` is larger than `length`, and implemented `removeRange` to support it.

I have also added tests for:
> * `contains`
> * `set length` -> should throw
> * `removeWhere`
> * `retainWhere`
> * `removeLast`
> * `removeRange`
> * `replaceRange`

test/list_test.dart
nielsenko: I have added checks for that as well

nirinchev (approved):

PR #890, opened 2022-09-08
Cleanup README.md while at it

nielsenko:


blagoev (commented):
Not sure what we are cleaning. the changes are too verbose. why are some rows split into two. Also the formatting is changed too much. Not sure what is wrong with the old readme from these changes.

nielsenko:
Well, if you feel like that. I was just aligning layout and cleaning up while fixing the license mess..

PR #891, opened 2022-09-08
Validate dynamic library version on Startup

nielsenko:
Stamps library version into both dart and native code when preparing
release.

Compare the version when opening the library and throw a descriptive
exception if they don't match.


desistefanova (approved):

wilsleygermano:
Any expected merging date? I really need this...

PR #893, opened 2022-09-09
Results of primitives

nielsenko:
kn/indexes


blagoev:
@cla-bot check

CHANGELOG.md
```diff
@@ -3,7 +3,8 @@
 **This project is in the Beta stage. The API should be quite stable, but occasional breaking changes may be made.**
 
 ### Enhancements
-* None
+* Support results of primitives, ie. `RealmResult<int>`.
+  Result of primitives do not support notifications and that is captured in the type system. ([#162](https://github.com/realm/realm-dart/issues/162))
```
```suggestion
  Result of primitives do not support notifications. ([#162](https://github.com/realm/realm-dart/issues/162))
```
I suggest we go with this shorter description.

lib/src/list.dart
```diff
@@ -42,6 +42,8 @@ abstract class RealmList<T extends Object?> with RealmEntity implements List<T>,
   /// and it's parent object hasn't been deleted.
   bool get isValid;
 
+  RealmResults<T> get asResults;
```
blagoev, line 45: I really think this should be a method. Similarly to `List.asMap()` and others

lib/src/list.dart
```diff
@@ -165,6 +167,9 @@ class ManagedRealmList<T extends Object?> with RealmEntity, ListMixin<T> impleme
     final frozenRealm = realm.freeze();
     return frozenRealm.resolveList(this)!;
   }
+
+  @override
+  RealmResults<T> get asResults => RealmResultsInternal.createFromList(this, realm, _metadata);
```
blagoev, line 172:
```suggestion
  RealmResults<T> asResults() => RealmResultsInternal.createFromList(this, realm, _metadata);
```

lib/src/results.dart
```diff
@@ -41,17 +42,14 @@ class RealmResults<T extends RealmObjectBase> extends collection.IterableBase<T>
...
-  }
-
-  /// Returns a new [RealmResults] filtered according to the provided query.
-  ///
-  /// The Realm Dart and Realm Flutter SDKs supports querying based on a language inspired by [NSPredicate](https://academy.realm.io/posts/nspredicate-cheatsheet/)
-  /// and [Predicate Programming Guide.](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Predicates/AdditionalChapters/Introduction.html#//apple_ref/doc/uid/TP40001789)
-  RealmResults<T> query(String query, [List<Object> args = const []]) {
-    final handle = realmCore.queryResults(this, query, args);
-    return RealmResultsInternal.create<T>(handle, realm, _metadata);
+    final meta = _metadata;
```
blagoev, line 45: Why not check if T is RealmObjectBase which is the real marker if this can create objects or it is a results of primitives.

blagoev (changes requested):
I have some suggestions

lib/src/results.dart
```diff
@@ -94,6 +82,35 @@ class RealmResults<T extends RealmObjectBase> extends collection.IterableBase<T>
...
+extension RealmResultsOfObject<T extends RealmObjectBase> on RealmResults<T> {
+  RealmResults<T> snapshot() {
+    final handle = realmCore.resultsSnapshot(this);
+    return RealmResults<T>._(handle, realm, _metadata);
+  }
+
+  /// Returns a new [RealmResults] filtered according to the provided query.
+  ///
+  /// The Realm Dart and Realm Flutter SDKs supports querying based on a language inspired by [NSPredicate](https://academy.realm.io/posts/nspredicate-cheatsheet/)
+  /// and [Predicate Programming Guide.](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Predicates/AdditionalChapters/Introduction.html#//apple_ref/doc/uid/TP40001789)
```
nirinchev, line 97: This should link to https://www.mongodb.com/docs/realm/realm-query-language/ not the incorrect cheatsheet.

lib/src/results.dart
```diff
@@ -110,11 +127,21 @@ extension RealmResultsInternal on RealmResults {
...
+    RealmObjectMetadata? metadata,
+  ) =>
+      RealmResults<T>._(handle, realm, metadata);
+
+  static RealmResults<T> createFromList<T extends Object?>(
+    RealmList list,
+    Realm realm,
+    RealmObjectMetadata? metadata,
+  ) =>
+      RealmResults<T>._(realmCore.resultsFromList(list), realm, metadata);
```
nirinchev, line 144: It's kind of weird this is here. Why not have `list.asResults` call `realmCore.resultsFromList(list)` and then use the handle to call `ResultsInternal.create`?

lib/src/list.dart
nirinchev, line 172: There are no tests for this API when used on a list of objects. I think we should add some to make sure it behaves as expected when it comes to querying and sorting.

nirinchev (changes requested):

lib/src/results.dart
nirinchev: We can check either one, but we should assert on the other - i.e. if we get a `RealmObjectBase` with no meta or meta with a non-RealmObjectBase generic, that would be a bug in the SDK.

lib/src/native/realm_core.dart
```diff
@@ -875,7 +875,20 @@ class _RealmCore {
...
+
+  Object? resultsGetElementAt(RealmResults results, int index) {
+    return using((Arena arena) {
+      final realm_value = arena<realm_value_t>();
+      _realmLib.invokeGetBool(() => _realmLib.realm_results_get(results.handle._pointer, index, realm_value));
+      return realm_value.toDartValue(results.realm);
+    });
+  }
+
+  RealmObjectHandle resultsGetObjectAt(RealmResults results, int index) {
```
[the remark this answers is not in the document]
blagoev: one is returning a `RealmObject` while the other is returning a `realm_value`

lib/src/results.dart
blagoev: I agree we can assert the metadata, but I think we should  only check the type is `RealmObjectBase` since this is the real marker.

lib/src/results.dart
```diff
@@ -94,6 +82,35 @@ class RealmResults<T extends RealmObjectBase> extends collection.IterableBase<T>
   }
 }
 
+// The query operations on results, as well as the ability to subscribe for notifications,
+// only work for results of objects (core restriction), so we add these as an extension methods
+// to allow the compiler to prevent misuse.
+extension RealmResultsOfObject<T extends RealmObjectBase> on RealmResults<T> {
+  RealmResults<T> snapshot() {
```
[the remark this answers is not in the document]
blagoev: that's the public api

lib/src/results.dart
blagoev: good point. essentially removing `createFromList`

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
nielsenko: Correct

CHANGELOG.md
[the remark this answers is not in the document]
nielsenko: You tell me 😄 .. We don't support notifications on list of primitives either. It goes back to this internal discussion: https://mongodb.slack.com/archives/C010R3CMP3N/p1644327500575779, but perhaps I misunderstood?

lib/src/results.dart
nielsenko: Fixed

lib/src/list.dart
```diff
@@ -194,8 +202,7 @@ extension RealmListOfObject<T extends RealmObjectBase> on RealmList<T> {
   /// The Realm Dart and Realm Flutter SDKs supports querying based on a language inspired by [NSPredicate](https://academy.realm.io/posts/nspredicate-cheatsheet/)
   /// and [Predicate Programming Guide.](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/Predicates/AdditionalChapters/Introduction.html#//apple_ref/doc/uid/TP40001789)
   RealmResults<T> query(String query, [List<Object> arguments = const []]) {
-    final managedList = asManaged();
-    final handle = realmCore.queryList(managedList, query, arguments);
+    final handle = realmCore.queryList(asManaged, query, arguments);
```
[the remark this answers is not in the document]
nielsenko: Okay

lib/src/results.dart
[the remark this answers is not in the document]
nielsenko: I have added a test case for it

CHANGELOG.md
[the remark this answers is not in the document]
nielsenko: I will try to remove this restriction, and on list as well.

lib/src/results.dart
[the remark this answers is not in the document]
nirinchev: Why do we have `snapshot` in the public API? That seems wrong to me. We should use it internally, but shouldn't expect users to know when and why to snapshot.

lib/src/list.dart
nirinchev: Yeah, this is testing it for results of primitives - I was thinking that we need a test for results of objects - something like:

```dart
final results = myList.asResults.query('foo > 5 SORT(bar)');

// assert on the order and contents

realm.deleteMany(results);
// assert that only the results matches were deleted and not everything from the list
```

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
nirinchev: I can see that. Since `realm_value` can hold objects, why not use the same method for both?

lib/src/results.dart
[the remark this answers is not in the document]
nielsenko: You are right, that snapshot is hard to explain, and freeze probably covers most of the cases where a user would want it. I'll remove from now, including the newly added test.

CHANGELOG.md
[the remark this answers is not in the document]
nielsenko: My mistake. We do support notifications. Has exposed for both lists and results, and added tests.

lib/src/results.dart
nielsenko: I have reworked this to check on type instead. Assert on `_meta != null` is implicit in:
```dart
      final handle = realmCore.resultsGetObjectAt(this, index);
      final accessor = RealmCoreAccessor(metadata, realm.isInMigration);
```
I expect the first line to fail, and otherwise `metadata => _metadata!` will.

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
nielsenko: You still need to convert the handle, and the later is less efficient.

lib/src/list.dart
```diff
@@ -165,6 +171,18 @@ class ManagedRealmList<T extends Object?> with RealmEntity, ListMixin<T> impleme
...
   }
+
+  @override
+  RealmResults<T> asResults() => RealmResultsInternal.create<T>(realmCore.resultsFromList(this), realm, metadata);
+
+  @override
+  Stream<RealmListChanges<T>> get changes {
+    if (isFrozen) {
+      throw RealmStateError('List is frozen and cannot emit changes');
+    }
```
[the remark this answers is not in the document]
nielsenko: This bit of code was just moved from `RealmListOfObject` extension class, so as to allow notifications on list of primitives as well. I didn't change the semantics.

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Enhancements
 * Added `MutableSubscriptionSet.removeByType` for removing subscriptions by their realm object type. (Issue [#317](https://github.com/realm/realm-dart/issues/317))
+* Support results of primitives, ie. `RealmResult<int>`. (Issue [#162](https://github.com/realm/realm-dart/issues/162))
```
desistefanova, line 7:
```suggestion
* Support `RealmResults` of primitive types (bool, int, string ...), ie. `RealmResults<int>`. (Issue [#162](https://github.com/realm/realm-dart/issues/162))
```

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
...
+* Support notifications on all managed realm lists, including list of primitives, ie. `RealmList<int>.changes` is supported. ([#893](https://github.com/realm/realm-dart/pull/893))
```
desistefanova, line 8:
```suggestion
* Support notifications on all managed `RealmList`s, including list of primitive types, for example: `RealmList<int>.changes` is supported. ([#893](https://github.com/realm/realm-dart/pull/893))
```

lib/src/results.dart
```diff
@@ -82,15 +87,17 @@ class RealmResults<T extends RealmObjectBase> extends collection.IterableBase<T>
...
-  RealmResults<T> freeze() {
-    if (isFrozen) {
-      return this;
-    }
-
-    final frozenRealm = realm.freeze();
-    return frozenRealm.resolveResults(this);
+// The query operations on results only work for results of objects (core restriction),
+// so we add it as an extension methods to allow the compiler to prevent misuse.
+extension RealmResultsOfObject<T extends RealmObjectBase> on RealmResults<T> {
```
desistefanova, line 94:
```suggestion
extension RealmResultsOfObject<T extends RealmObject> on RealmResults<T> {
```
It shouldn't be allowed for embedded objects, I think.
Can you make a query on a RealmResults of embedded objects?

nielsenko: Yes you can, see `embedded_test.dart:724` as an example

desistefanova: If query is working only for RelmObjects then we have to change also tha same extension on RealmList<T>.

CHANGELOG.md
nielsenko: I think this would be inconsistent. We have used the word primitives before in the CHANGELOG without defining it, and we often refer to the class `RealmList` as realm list(s) in free text.

lib/src/results.dart
desistefanova: Ah, OK.

CHANGELOG.md
desistefanova: It sounds better to me. But it is just a suggestion. It is not important. As you decide.

desistefanova (approved):

nirinchev (approved):

blagoev (approved):

PR #894, opened 2022-09-09
Add realm specific implementation for List.add and List.insert

nielsenko:


nirinchev (approved):

test/list_test.dart
```diff
@@ -699,4 +699,15 @@ Future<void> main([List<String>? args]) async {
     expect(team.players, isNot(players));
     expect(team.players, unorderedMatches(players));
   });
+
+  test('ManagedRealmList.length= throws', () {
```
blagoev, line 703: should we have a test for length decrease as well. Now the `removeRange` path is not tested

blagoev (approved):
I think we should add a test for the `removeRange` path of length setter

nielsenko:
> I think we should add a test for the `removeRange` path of length setter

I have added the requested test

lib/src/list.dart
```diff
@@ -63,7 +63,11 @@ class ManagedRealmList<T extends Object?> extends collection.ListBase<T> with Re
   @override
```
desistefanova, line 63:
```suggestion
  /// Setting the `length` is a required method on [List], but makes less sense
  /// for [RealmList]s. You can only decrease the length, increasing it throws an [RealmException].
```

PR #900, opened 2022-09-12
Fix results query bug

### LINKED ISSUE #908: Queries on results are broken
```dart
expect(realm.query<Person>('FALSEPREDICATE').query('TRUEPREDICATE'), isEmpty);
```
fails, if there are any `Person` in `realm`

nielsenko:
Fixes #908

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@
 * Allow null arguments on query. ([#872](https://github.com/realm/realm-dart/pull/872)). Fixes [#871](https://github.com/realm/realm-dart/issues/871)
 * Previously removeAt did not truncate length. ([#884](https://github.com/realm/realm-dart/pull/884)). Fixes [#883](https://github.com/realm/realm-dart/issues/883)
 * List.length= now throws, if you try to increase length, ([#894](https://github.com/realm/realm-dart/pull/894)).
+* Queries on results didn't filter the existing results. Ex:
+  ```dart
+  expect(realm.query<Person>('FALSEPREDICATE').query('TRUEPREDICATE'), isEmpty); // would fail if any Persons exists
+  ```
+  * Queries on results didn't filter the existing results. ([#908](https://github.com/realm/realm-dart/issues/908))
+  ```dart
+  expect(realm.query<Person>('FALSEPREDICATE').query('TRUEPREDICATE'), isEmpty); // would fail if any Persons exists
 
 ### Compatibility
 * Realm Studio: 12.0.0 or later.
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -15,6 +15,7 @@
     "dylib",
     "ejson",
     "errno",
+    "FALSEPREDICATE",
     "ffigen",
     "finalizable",
     "finalizer",
--- a/lib/src/native/realm_bindings.dart
+++ b/lib/src/native/realm_bindings.dart
@@ -3845,6 +3845,26 @@ class RealmLibrary {
   late final _realm_dictionary_size = _realm_dictionary_sizePtr.asFunction<
       bool Function(ffi.Pointer<realm_dictionary_t>, ffi.Pointer<ffi.Size>)>();
 
+  /// Convert a dictionary to results.
+  ///
+  /// @return A non-null pointer if no exception occurred.
+  ffi.Pointer<realm_results_t> realm_dictionary_to_results(
+    ffi.Pointer<realm_dictionary_t> arg0,
+  ) {
+    return _realm_dictionary_to_results(
+      arg0,
+    );
+  }
+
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * Allow null arguments on query. ([#872](https://github.com/realm/realm-dart/pull/872)). Fixes [#871](https://github.com/realm/realm-dart/issues/871)
 * Previously removeAt did not truncate length. ([#884](https://github.com/realm/realm-dart/pull/884)). Fixes [#883](https://github.com/realm/realm-dart/issues/883)
 * List.length= now throws, if you try to increase length, ([#894](https://github.com/realm/realm-dart/pull/894)).
+* Queries on results was broken. ([#900](https://github.com/realm/realm-dart/pull/900)). Fixes issue [#908](https://github.com/realm/realm-dart/issues/908)
```
nirinchev, line 12: That's way too vague. Clearly they were working for a range of cases, otherwise our tests wouldn't have passed.

nielsenko: Can't interested parties just follow the link to the issue?

nielsenko: Anyway, I have clarified:
```markdown
* Queries on results didn't filter the existing results. Fixes issue [#908](https://github.com/realm/realm-dart/issues/908)
```

nielsenko: I have added the example from the issue in the CHANGELOG, so that people won't need to press a link in the browser they are using to read the same.

nirinchev: Thank you. I want to clarify I'm not arbitrarily asking for this and I'd be happy to discuss the level of detail we aim for in changelog entries if you think that'll be valuable.

nirinchev (approved):

CHANGELOG.md
```diff
@@ -9,6 +9,11 @@
 * Allow null arguments on query. ([#872](https://github.com/realm/realm-dart/pull/872)). Fixes [#871](https://github.com/realm/realm-dart/issues/871)
 * Previously removeAt did not truncate length. ([#884](https://github.com/realm/realm-dart/pull/884)). Fixes [#883](https://github.com/realm/realm-dart/issues/883)
 * List.length= now throws, if you try to increase length, ([#894](https://github.com/realm/realm-dart/pull/894)).
+* Queries on results didn't filter the existing results. Ex:
+  ```dart
+  expect(realm.query<Person>('FALSEPREDICATE').query('TRUEPREDICATE'), isEmpty); // would fail if any Persons exists
+  ```
+  Fixes issue [#908](https://github.com/realm/realm-dart/issues/908)
```
blagoev, line 16: I think we are diverging from our changelog format. Why do we list the issue here? Previously we just list the PR with the change. In that PR there should be a link to the issue.

blagoev (commented):
changelog format looks weird

CHANGELOG.md
nirinchev: I believe in all other repos we're listing the issue rather than the PR. There are two main reasons:
1. Users who file bug reports would like to see which release their bug was fixed in.
2. A single issue may be implemented over multiple PRs.

I don't know if that's as relevant today since Github now detects the release that includes a fix, but that was the reason before.

nielsenko: I previously found this documented in:

According to https://wiki.corp.mongodb.com/display/REALM/Release+Notes+and+changelogs:
> Fixed
> Describe how the bug can be observed or provoked by the end-user and the consequences. Do not describe any implementation details. You can do that in the internal section or keep that in the issue/PR you reference.
> 
> If possible with reasonable effort (git blame) mention which version the bug was introduced in. This helps the end-user understand if they are at all affected by this bug.
> 
> General format:
> 
> * <when doing this, this could happen>. <Thanks to @external-contributor!> ([#_issue_](_url_), since v?)

So according to that we should link to the issue (not the PR) and describe how to hit it according to the above.

Personally I don't mind if we link it to the PR or the issue, I just aligned with what @nirinchev  suggested on my list related pr #906. 

Wrt. to the small sample that was added:
```dart
expect(realm.query<Person>('FALSEPREDICATE').query('TRUEPREDICATE'), isEmpty); // would fail if any Persons exists
```
I personally find it a bit overkill, as the user can just follow the link to the issue and get the same information, but I agree it has value that the users can evaluate if the fixed bug may/or may not impact them without traversing the link.

blagoev: Ok lets list the issue. Then I have a formatting suggestion. Will try to do that in code

blagoev:
```suggestion
  * Queries on results didn't filter the existing results. ([#908](https://github.com/realm/realm-dart/issues/908))
  ```dart
  expect(realm.query<Person>('FALSEPREDICATE').query('TRUEPREDICATE'), isEmpty); // would fail if any Persons exists
  ```
```

PR #902, opened 2022-09-14
Implement list of nullables with minimal refactorings

### LINKED ISSUE #163: Support nullable types in collections
RealmResulsts should support null values

depends on https://github.com/realm/realm-dart/issues/178

nielsenko:
Allow list of nullable primitives in realm models, such as:
```dart
@RealmModel()
class _Player {
  @PrimaryKey()
  late String name;
  final scoresByRound = <int?>[]; // null means player didn't finish
}
```
Fixes #163 (supersedes  #708)

### DIFF
--- a/lib/src/realm_object.dart
+++ b/lib/src/realm_object.dart
@@ -178,7 +178,7 @@ class RealmCoreAccessor implements RealmAccessor {
   void set(RealmObject object, String name, Object? value, {bool isDefault = false, bool update = false}) {
     final propertyMeta = metadata[name];
     try {
-      if (value is RealmList<Object>) {
+      if (value is RealmList<Object?>) {
         final handle = realmCore.getListProperty(object, propertyMeta.key);
         if (update) realmCore.listClear(handle);
         for (var i = 0; i < value.length; i++) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 * You can now set a realm property of type `T` to any object `o` where `o is T`. Previously it was required that `o.runtimeType == T`. ([#904](https://github.com/realm/realm-dart/issues/904))
 * Performance of indexOf on realm lists has been improved. It now uses realm-core instead of the generic version from ListMixin. ([#911](https://github.com/realm/realm-dart/pull/911))
 * Performance of remove on realm list has been improved. It now uses indexOf and removeAt. ([#915](https://github.com/realm/realm-dart/pull/915))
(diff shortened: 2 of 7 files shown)


nirinchev (approved):
Looks fine, even though I'd have preferred the const schema approach.

desistefanova (approved):

lib/src/realm_object.dart
```diff
@@ -178,7 +178,7 @@ class RealmCoreAccessor implements RealmAccessor {
   void set(RealmObject object, String name, Object? value, {bool isDefault = false, bool update = false}) {
     final propertyMeta = metadata[name];
     try {
-      if (value is RealmList<Object>) {
+      if (value is RealmList<Object?>) {
```
blagoev, line 181: if `value is RealmList<Object?>` then value is not `RealmList<Object>`. Are we sure we are not breaking the old behavior? Should this be 
```dart
if (value is RealmList<Object?> || value is RealmList<Object>)
```

blagoev (changes requested):
Looks good, just want to make sure we have noticed the change in my comment.

lib/src/realm_object.dart
nielsenko:
```dart
value is RealmList<Object>
```
implies
```dart
value is RealmList<Object?>
```
https://dartpad.dev/?id=d9aecd56b5a115b6a5672f0fbefbd49f

blagoev: I don't understand the answer. We are changing a check from `if (value is RealmList<Object>)` to `if (value is RealmList<Object?>)`. Which now does not capture  `value is RealmList<Object>` . Isn't this a issue?

nirinchev: The way I understand the answer, if I have `final foo = RealmList<Object>()`, then `foo is RealmList<Object?>` will return `true`. So the new check is a looser one - it will return `true` for both `RealmList<Object>` and `RealmList<Object?>`. Looking at the dartpad link, you can see that the compiler correctly detects that and gives you a hint saying `Unnecessary type check; the result is always 'true'.`.

nielsenko: Correct

nielsenko: I have added an extra test case that demonstrates covariance of realm lists.

@blagoev What may confuse you is that in dart generics are covariant by default.

nielsenko: https://dart.dev/guides/language/type-system#generic-type-assignment

blagoev: Thanks. Indeed. I was testing the other way around. Then there is no issue

blagoev (approved):

PR #905, opened 2022-09-14
Fix property set with subtype

### LINKED ISSUE #904: Cannot set property to subtype
The class `TZDateTime` from `timezone` package implements `DateTime`, but it is not possible to set `DateTime` property on a realm model to a `TZDateTime`. This is due to how `_intoRealmValue` uses a switch on the values `runtimeType` instead of testing with `is`.

nielsenko:
This fixes #904

### DIFF
--- a/lib/src/native/realm_core.dart
+++ b/lib/src/native/realm_core.dart
@@ -2035,56 +2035,48 @@ void _intoRealmValue(Object? value, Pointer<realm_value_t> realm_value, Allocato
     realm_value.ref.values.link.target_table = link.classKey;
     realm_value.ref.type = realm_value_type.RLM_TYPE_LINK;
   } else {
-    switch (value.runtimeType) {
-      case int:
-        realm_value.ref.values.integer = value as int;
-        realm_value.ref.type = realm_value_type.RLM_TYPE_INT;
-        break;
-      case bool:
-        realm_value.ref.values.boolean = value as bool;
-        realm_value.ref.type = realm_value_type.RLM_TYPE_BOOL;
-        break;
-      case String:
-        String string = value as String;
-        final units = utf8.encode(string);
-        final result = allocator<Uint8>(units.length);
-        final Uint8List nativeString = result.asTypedList(units.length);
-        nativeString.setAll(0, units);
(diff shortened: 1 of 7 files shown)


lib/src/native/realm_core.dart
```diff
@@ -1997,56 +1997,48 @@ void _intoRealmValue(Object? value, Pointer<realm_value_t> realm_value, Allocato
...
-        if (microseconds < 0 && nanoseconds != 0) {
-          nanoseconds = nanoseconds - _nanosecondsPerMicrosecond * _microsecondsPerSecond;
-        }
-        realm_value.ref.values.timestamp.seconds = seconds;
-        realm_value.ref.values.timestamp.nanoseconds = nanoseconds;
-        realm_value.ref.type = realm_value_type.RLM_TYPE_TIMESTAMP;
-        break;
-      default:
-        throw RealmException("Property type ${value.runtimeType} not supported");
+    if (value is int) {
```
nirinchev, line 2000: If only we had pattern matching ;)

nirinchev (approved):

lib/src/native/realm_core.dart
blagoev, line 2000: The code will work but why are we doing this change. Did the previous version had drawbacks?

pubspec.yaml
```diff
@@ -37,6 +37,7 @@ dev_dependencies:
   json_serializable: ^6.3.1
   lints: ^2.0.0
   test: ^1.14.3
+  timezone: ^0.9.0 # Used for testing only
```
blagoev, line 40:
```suggestion
  timezone: ^0.9.0
```

blagoev: if we need it as dev_dep then we need it. I suggest we don't have the comment here

lib/src/native/realm_core.dart
blagoev: ok I understand now

blagoev (approved):

blagoev (approved):
No changelog entry

desistefanova (approved):

nielsenko:
> No changelog entry

Added

PR #906, opened 2022-09-14
Fix list query bug

nielsenko:
Fixes #909

PR #911, opened 2022-09-15
Implement indexOf using realm_list_find

### LINKED ISSUE #910: Support list indexOf
Currently we don't use realm core to implement `indexOf`, but relies on `ListMixin` for the implementation. That is somewhat inefficient, but until https://github.com/realm/realm-core/pull/5848 landed, we kind of had to.

nielsenko:
Fixes #910

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Enhancements
 * Added support for "frozen objects" - these are objects, queries, lists, or Realms that have been "frozen" at a specific version. All frozen objects can be accessed and queried as normal, but attempting to mutate them or add change listeners will throw an exception. `Realm`, `RealmObject`, `RealmList`, and `RealmResults` now have a method `freeze()` which returns an immutable version of the object, as well as an `isFrozen` property which can be used to check whether an object is frozen. (Issue [#56](https://github.com/realm/realm-dart/issues/56))
+* Performance of indexOf on realm lists has been improved. It now uses realm-core instead of the generic version from ListMixin. ([#911](https://github.com/realm/realm-dart/pull/911)). 
 
 ### Fixed
 * Allow null arguments on query. ([#872](https://github.com/realm/realm-dart/pull/872)). Fixes [#871](https://github.com/realm/realm-dart/issues/871)
--- a/lib/src/realm_object.dart
+++ b/lib/src/realm_object.dart
@@ -391,13 +391,16 @@ extension RealmObjectInternal on RealmObject {
     return object;
   }
 
-  RealmObjectHandle get handle {
-    final h = _handle;
-    if (h != null) return h;
-    throw RealmStateError('$this is not managed');
-  }
+  RealmObjectHandle get handle => _handle ?? failUnmanaged(this);
 
   RealmAccessor get accessor => _accessor;
+}
+
+Never failUnmanaged(RealmEntity entity) {
+  final message = '$entity is not managed';
+  assert(false, message);
+  // in production we fall back to throwing state error as a precaution
+  throw RealmStateError(message);
 }
 
 /// An exception being thrown when a `Realm` operation or [RealmObject] access fails.
--- a/lib/src/realm_object.dart
+++ b/lib/src/realm_object.dart
@@ -404,14 +404,10 @@ extension RealmObjectInternal on RealmObject {
     return object;
   }
 
-  RealmObjectHandle get handle => _handle ?? failUnmanaged(this);
+  // if we ever see a _CastError here, we forgot to guard against misuse further up the call-stack
+  RealmObjectHandle get handle => _handle!;
 
   RealmAccessor get accessor => _accessor;
-}
-
-Never failUnmanaged(RealmEntity entity) {
-  assert(false, '$entity is not managed');
-  exit(-1);
 }
 
 /// An exception being thrown when a `Realm` operation or [RealmObject] access fails.
--- a/lib/src/realm_object.dart
+++ b/lib/src/realm_object.dart
@@ -18,6 +18,7 @@
 
 import 'dart:async';
 import 'dart:ffi';
+import 'dart:io';
 
 import 'list.dart';
 import 'native/realm_core.dart';
@@ -409,10 +410,8 @@ extension RealmObjectInternal on RealmObject {
 }
 
 Never failUnmanaged(RealmEntity entity) {
-  final message = '$entity is not managed';
-  assert(false, message);
-  // in production we fall back to throwing state error as a precaution
-  throw RealmStateError(message);
+  assert(false, '$entity is not managed');
+  exit(-1);
 }
 
 /// An exception being thrown when a `Realm` operation or [RealmObject] access fails.
--- a/lib/src/realm_object.dart
+++ b/lib/src/realm_object.dart
@@ -391,7 +391,12 @@ extension RealmObjectInternal on RealmObject {
     return object;
   }
 
-  RealmObjectHandle get handle => _handle!;
+  RealmObjectHandle get handle {
+    final h = _handle;
+    if (h != null) return h;
+    throw RealmStateError('$this is not managed');
+  }
+
   RealmAccessor get accessor => _accessor;
 }
 
--- a/test/list_test.dart
+++ b/test/list_test.dart
@@ -808,7 +808,7 @@ Future<void> main([List<String>? args]) async {
     final dartList = <int>[1, 2, 3];
     expect((dartList as List<Object>).indexOf("abc"), -1); // ignore: unnecessary_cast
 
-    // Test that realm list behaves the same
-    expect((players as List<Object>).indexOf(1), -1); // ignore: unnecessary_cast
+    // Realm list behaves differently
+    expect(() => (players as List<Object>).indexOf(1), throwsA(isA<RealmException>())); // ignore: unnecessary_cast
   });
 }
--- a/test/list_test.dart
+++ b/test/list_test.dart
@@ -815,7 +815,8 @@ Future<void> main([List<String>? args]) async {
     realm.write(() => realm.add(team));
     final players = team.players;
 
-    // expect(players, isA<ManagedRealmList<Person>>());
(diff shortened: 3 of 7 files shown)


test/list_test.dart
```diff
@@ -724,4 +725,31 @@ Future<void> main([List<String>? args]) async {
...
+
+    // expect(players, isA<ManagedRealmList<Person>>());
+    expect(players.indexOf(players.first, -1), 0); // okay to start from negative index
+    expect(players.indexOf(players.first, 1), -1); // start respected
+    expect(players.indexOf(players.first, 101), -1); // okay to start from non-existent index
+
+    var index = 0;
+    final r = Random(42); // deterministic
+    for (final p in players) {
+      expect(players.indexOf(p, r.nextInt(index + 1) - 1), index++);
```
nirinchev, line 745: Can we add a test that uses `indexOf` for an unmanaged object (i.e. `players.indexOf(Person('a')`).

nirinchev (approved):
Looks good, two minor things from me.

test/list_test.dart
```diff
@@ -20,6 +20,7 @@
 
 import 'dart:math';
 
+import 'package:realm_dart/src/list.dart';
```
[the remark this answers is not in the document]
nielsenko: It is not.. it will break the build in spectacular ways see: https://github.com/dart-lang/sdk/issues/49968

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 **This project is in the Beta stage. The API should be quite stable, but occasional breaking changes may be made.**
 
 ### Enhancements
-* None
+* Performance of indexOf on realm lists has been improved. It now uses realm-core instead of the generic version from ListMixin. ([#911](https://github.com/realm/realm-dart/pull/911)). Fixes issue [#910](https://github.com/realm/realm-dart/issues/910).
```
blagoev, line 6:
```suggestion
* Performance of indexOf on realm lists has been improved. It now uses realm-core instead of the generic version from ListMixin. ([#911](https://github.com/realm/realm-dart/pull/911)).
```

test/list_test.dart
```diff
@@ -724,4 +724,33 @@ Future<void> main([List<String>? args]) async {
...
+    final r = Random(42); // deterministic
+    for (final p in players) {
+      expect(players.indexOf(p, r.nextInt(index + 1) - 1), index++);
+    }
+
+    // indexOf with wrong type of element, just returns -1. Proof:
+    final dartList = <int>[1, 2, 3];
+    expect((dartList as List<Object>).indexOf("abc"), -1); // ignore: unnecessary_cast
+
+    // Realm list behaves differently
```
blagoev, line 751: `differently` than what? Should we clarify the comment here?

lib/src/realm_object.dart
```diff
@@ -352,7 +352,12 @@ extension RealmObjectInternal on RealmObject {
     return object;
   }
 
-  RealmObjectHandle get handle => _handle!;
+  RealmObjectHandle get handle {
+    final h = _handle;
+    if (h != null) return h;
+    throw RealmStateError('$this is not managed');
```
blagoev, line 358: this seems not in the correct place and most certainly would be an SDK bug.

test/list_test.dart
```diff
@@ -724,4 +724,33 @@ Future<void> main([List<String>? args]) async {
...
+    expect(() => (players as List<Object>).indexOf(1), throwsA(isA<RealmException>())); // ignore: unnecessary_cast
+
+    expect(() => players.indexOf(Person('10')), throwsA(isA<RealmStateError>()));
```
blagoev, line 754: I find it weird that we expect this exception here but it is actually thrown indirectly in the `handle` getter above. I think we better move the throwing of this error to the more exact place. So probably in the `indexOf` method itself.

blagoev: I couldn't follow exectly where do we read the `handle` which will throw

blagoev (changes requested):
I think we need to move the throwing of the Error in a more obvious place. Now it's really indirect which makes the validation logic hard to follow

lib/src/realm_object.dart
nielsenko: Previously this just did `=> _handle!` which would throw `_CastError:<Null check operator used on a null value>`. This can happen, if you try to index a managed list with an unmanaged object. I think that is a state error.

test/list_test.dart
nielsenko: It is called in
```dart
RealmLinkHandle _getObjectAsLink(RealmObject object) {
  final realmLink = _realmLib.realm_object_as_link(object.handle._pointer);
  return RealmLinkHandle._(realmLink);
}
```

nielsenko: This is a case of misuse by the user, we always `RealmStateError` in these cases, if we cannot prevent it compile time.

nirinchev: I believe the pattern of calling `_handle!` is something we use in a lot of places. We can replace it with an assert, but I agree with Lubo that those preconditions are ideally checked at the API entrypoint as that's where we can provide most information. For example, this would allow us to throw a friendlier exception like `Can't call indexOf with an argument that is an unmanaged object`. You could argue that developers can potentially discover what went wrong by evaluating the stacktrace, but historically, we've had a lot of issues created where only the exception message has been available.

nielsenko: So I can add the check earlier, but I still think it should be a `RealmStateError` since the user is mis-using the api in a way we cannot detect compile time. This is no different than say trying to listen for changes on a unmanaged or frozen list. In these cases we will also throw a `RealmStateError`

nirinchev: Agree - not opposed to the exception type.

lib/src/realm_object.dart
blagoev: If you try to index a managed list with an unmanaged object we should be checking that in the indexer itself. So the cast on a null value is something we should be seeing only while writing tests and new functionality.
So I think we should move this check for is object managed in the indexer.

nirinchev: I agree with moving the check to the indexer, but we could probably keep the assertion here for a cleaner error message.

blagoev: This will mask the error for us and could hide from us in our tests since we expect the same error in some cases. 
The other error message `_CastError:<Null check operator used on a null value>` will make us fix it on the spot cause it will slip through any expects.

blagoev (changes requested):
Lets handle the comments about the place where we throw the `RealmStateError`

desistefanova (approved):

desistefanova (changes requested):
Be sure to be merged with core [v12.7.0] coming with [PR Fix list query bug](https://github.com/realm/realm-dart/pull/906)  before to push into master.

nielsenko:
> Be sure to be merged with core [v12.7.0] coming with [PR Fix list query bug](https://github.com/realm/realm-dart/pull/906) before to push into master.

Sure

lib/src/realm_object.dart
nielsenko: I have replaced with an assert in debug. Also the `isManaged` check is done in `indexOf` now

test/list_test.dart
nielsenko: Checked in `indexOf` checks if `element` is unmanaged, and throws more detailed `RealmStateError` now.

nielsenko:
> Lets handle the comments about the place where we throw the `RealmStateError`

Done

test/list_test.dart
```diff
@@ -806,4 +806,43 @@ Future<void> main([List<String>? args]) async {
...
+
+  test('ManagedRealmList.indexOf', () {
+    final config = Configuration.local([Team.schema, Person.schema]);
+    final realm = getRealm(config);
+
+    final team = Team('sad team', players: [for (int i = 0; i < 100; ++i) Person('$i')]);
+    realm.write(() => realm.add(team));
+    final players = team.players;
+
+    // expect(players, isA<ManagedRealmList<Person>>());
```
desistefanova, line 818: Delete it if not needed.

desistefanova (approved):

lib/src/realm_object.dart
```diff
@@ -403,10 +403,18 @@ extension RealmObjectInternal on RealmObject {
...
+  RealmObjectHandle get handle => _handle ?? failUnmanaged(this);
+
   RealmAccessor get accessor => _accessor;
 }
 
+Never failUnmanaged(RealmEntity entity) {
+  final message = '$entity is not managed';
+  assert(false, message);
+  // in production we fall back to throwing state error as a precaution
+  throw RealmStateError(message);
```
blagoev, line 415: Why are we differentiating on build type. Lets throw one type of error which is the null message. If this happens in release we immediately know it is our fault. Unlike the current suggestion where it might be handled/silenced by the user or SDK code somewhere.

blagoev (changes requested):
We should let the null error bubble up and crash the app every time.

lib/src/realm_object.dart
nielsenko: No, we don't know it is our fault. It can only happen if the user tries to use an unmanaged object as a managed object.

Yes, we should probably catch the misuse earlier, and throw a `RealmStateError` with a better text there, but in the end it is state error, so I thought it prudent to fallback to throwing such.

blagoev: I am not sure I fully agree. If this one fails here it means we did not had the correct check for unmanged object in the correct place in the public API. This is our fault and we should add the correct check where needed. At that point we will throw a `RealmStateError` on the correct place with proper message.

blagoev: in the SDK we have many places where we dereference nullables and it could fail on these places the same, meaning we need to fix the conditions before we get to this point of null fail.

nielsenko: Changing to failing hard

nielsenko: Changing to `_handle!`

blagoev (approved):

PR #912, opened 2022-09-15
kn/dart sdk 49968

nielsenko:
- Introduce crash on purpose to highlight https://github.com/dart-lang/sdk/issues/49968

This is not intended to ever be merged, but only exists to exercise a branch on CI that provokes the above


nielsenko:
Issue seems well understood by now

PR #916, opened 2022-09-21
kn/open async suggestions

nielsenko:
ds/open_realm_async suggestions

### DIFF
--- a/test/realm_test.dart
+++ b/test/realm_test.dart
@@ -427,7 +427,7 @@ Future<void> main([List<String>? args]) async {
       teamTwo.players.addAll(newPlayers);
     });
 
-    //Ensule teams exist in realm
+    //Ensure teams exist in realm
     var teams = realm.all<Team>();
     expect(teams.length, 2);
 
@@ -788,7 +788,7 @@ Future<void> main([List<String>? args]) async {
     expect(stored.location, now.location);
     expect(stored.location.name, 'Europe/Copenhagen');
   });
-  
+
   test('Realm - open local not encrypted realm with encryption key', () {
     openEncryptedRealm(null, generateValidKey());
   });
@@ -994,7 +994,6 @@ Future<void> main([List<String>? args]) async {
     cancellationToken.cancel();
     expect(await realmIsCancelled, isTrue);
   });
-
 }
 
 List<int> generateValidKey() {
@@ -1009,8 +1008,8 @@ void openEncryptedRealm(List<int>? encryptionKey, List<int>? decryptionKey, {voi
     afterEncrypt(realm);
   }
   if (encryptionKey == decryptionKey) {
-    final decriptedRealm = getRealm(config2);
-    expect(decriptedRealm.isClosed, false);
+    final decryptedRealm = getRealm(config2);
+    expect(decryptedRealm.isClosed, false);
   } else {
     expect(
       () => getRealm(config2),


lib/src/realm_class.dart
```diff
@@ -132,29 +128,22 @@ class Realm implements Finalizable {
   /// Returns [Future<Realm>] that completes with the `realm` once the remote realm is fully synchronized or with an `error` if operation is canceled.
   /// When the configuration is [LocalConfiguration] this completes right after the local realm is opened or operation is canceled.
   static Future<Realm> open(Configuration config, {CancellationToken? cancellationToken, ProgressCallback? onProgressCallback}) async {
-    Realm realm = await CancellableFuture.fromFutureFunction<Realm>(() => _open(config, cancellationToken, onProgressCallback), cancellationToken);
-    return realm;
+    final realm = Realm(config);
```
desistefanova, line 131: I wanted to prevent this Realm to be opened before the cancellation logic to be attached. If I call cancellationToken.cancel() before to attach the releasing event that closes the Realm, then the Realm is opened successfully whiteout to be closed and the cancelled exception is thrown later.
And because of the exception the users can't get access to this realm instance and can never close it.

nielsenko: GC will close it. Realms are FInalizable, so it will be prioritezed

desistefanova: But the tests fail on teardown. :) And I prefer to be closed, not to leave any garbage to collect :)

nielsenko: No it does not fail on teardown.

lib/src/realm_class.dart
```diff
@@ -134,41 +134,28 @@ class Realm implements Finalizable {
...
+        await Future.wait<void>([
+          session.waitForDownload(cancellationToken),
+          if (onProgressCallback != null)
+            session
+                .getProgressStream(
+                  ProgressDirection.download,
+                  ProgressMode.forCurrentlyOutstandingWork,
+                )
+                .forEach(onProgressCallback)
+                .asCancellable(cancellationToken),
```
desistefanova, line 150: I think that it is better to call subscribtion.cancel(). asCancellable would not call realmCore.sessionUnregisterConnectionStateNotifier.

lib/src/realm_class.dart
```diff
@@ -134,41 +134,28 @@ class Realm implements Finalizable {
...
+        ]);
       }
+      // handle cancellation even if we are not using flexible sync
+      return await CancellableFuture.value(realm, cancellationToken);
```
desistefanova, line 154: We have to prevent realm to be opened at the beginning of this method in case the token had been cancelled before the Realm,open was called.

desistefanova (commented):
The method in [Open realm async](https://github.com/realm/realm-dart/pull/731#top) is already rewritten based on your and @nirinchev  suggestions. Please check it and if you agree, you can close this PR.

lib/src/realm_class.dart
nielsenko: Yes it will. Why do you think that?

lib/src/realm_class.dart
nielsenko: In that case we just close it again, but you can add a test for that at the very beginning if you think. I think that is a needless optimization.

desistefanova (approved):

nielsenko:
All suggestions have been addressed

PR #918, opened 2022-09-22
Remove extra byte in realm_string_t conversion

nirinchev:
`realm_string_t` should never have been 0-terminated because it knows its own length. In fact, adding an extra empty byte will result in `\u0000` being appended at the end of the string. I don't know how to test this since the only place it's used currently is for the password, but I've verified it manually by looking at the http request body.


nielsenko (approved):

nielsenko (changes requested):
`List<int>.toCharPtr` ends up calling 
```dart
Pointer<Uint8> toUint8Ptr(Allocator allocator) {
    final nativeSize = length + 1;
    final result = allocator<Uint8>(nativeSize);
    final Uint8List native = result.asTypedList(nativeSize);
    native.setAll(0, this); // copy
    native.last = 0; // zero terminate
    return result.cast();
  }
```
i guess that need to be re-considered as well?

desistefanova (approved):

nirinchev:
`List<int>.toCharPtr` is mapping to `char*` which needs null termination otherwise we can't infer the length.

blagoev (approved):
good catch.

PR #919, opened 2022-09-22
Add support for API key provider

nirinchev:
Adds support for API key provider and authentication.

* Adds support for `User.apiKeys`. This is a client that exposes functionality for managing user API keys.
* Adds support for `Credentials.apiKey` and `Credentials.serverApiKey`.
* Exposes `User.accessToken` and `User.refreshToken`.

I had to downgrade Core due to https://github.com/realm/realm-core/issues/5890.

Fixes https://github.com/realm/realm-dart/issues/393
Fixes https://github.com/realm/realm-dart/issues/385
Fixes https://github.com/realm/realm-dart/issues/386
Fixes https://github.com/realm/realm-dart/issues/387
Fixes https://github.com/realm/realm-dart/issues/388
Fixes https://github.com/realm/realm-dart/issues/389
Fixes https://github.com/realm/realm-dart/issues/390
Fixes https://github.com/realm/realm-dart/issues/435
Fixes https://github.com/realm/realm-dart/issues/433
Fixes https://github.com/realm/realm-dart/issues/432
Fixes https://github.com/realm/realm-dart/issues/394


src/realm_dart_sync.cpp
```diff
@@ -25,7 +25,7 @@ RLM_API void realm_dart_http_request_callback(realm_userdata_t userdata, const r
     // the pointers in error are to stack values, we need to make copies and move them into the scheduler invocation
     struct request_copy_buf {
         std::string url;
-        std::string body;
+        std::vector<char> body;
```
nirinchev, line 28: @nielsenko @fealebenpae can you please look at this logic. It's somewhat convoluted and I'm not sure I understand it correctly, but as far as I understand the docs, `std::string` is not a container, so it cannot be used as a buffer the way we're using it. What I'm seeing is that before and after the `ud->scheduler->invoke` on line 52, `request_copy_buf.body.data()` is returning different addresses if `body` is `std::string`. This is causing `request_copy.body` to point to mangled data. I don't know why we need the map and the vector for the headers and don't know if the url may suffer from similar problems.

nielsenko: With the latest change.. reverting to `std::string` for body in `request_copy_buf`, but re-constructing the `request_body` from the moved `buf` inside the callback I agree.

Nice find :-)

nielsenko (approved):
LGTM

src/realm_dart_sync.cpp
blagoev: I suggest we fix it instead of changing to std::vector also we do the same in `realm_dart_sync_error_handler_callback`

test/user_test.dart
```diff
@@ -116,4 +116,299 @@ Future<void> main([List<String>? args]) async {
...
+
+  baasTest('User.apiKeys.fetch with non existent returns null', (configuration) async {
+    final app = App(configuration);
+    final user = await getIntegrationUser(app);
+
+    final key = await user.apiKeys.fetch(ObjectId());
+    expect(key, isNull);
+  });
+
+  void expectApiKey(ApiKey? fetched, ApiKey expected) {
```
blagoev, line 150:
```suggestion
  void _expectApiKey(ApiKey? fetched, ApiKey expected) {
```

blagoev (changes requested):

src/realm_dart_sync.cpp
nirinchev: How do you suggest fixing it?

lib/src/native/realm_core.dart
```diff
@@ -1206,6 +1222,11 @@ class _RealmCore {
           request.headers.add(header.key, header.value);
         }
 
+        if (body != '') {
```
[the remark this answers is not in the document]
nirinchev: Yes, I'm investigating an issue which only reproduces on CI.

lib/src/user.dart
```diff
@@ -199,4 +312,7 @@ extension UserInternal on User {
   UserHandle get handle => _handle;
 
   static User create(UserHandle handle, [App? app]) => User._(handle, app);
+
+  static ApiKey createApiKey({required ObjectId id, required String name, required String? value, required bool isEnabled}) =>
```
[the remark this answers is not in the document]
nirinchev: I generally dislike having unnamed similar arguments, but since this is not a public API, I guess it's fine. I'll unname them.

lib/src/user.dart
```diff
@@ -39,6 +39,11 @@ class User {
     return _app ??= AppInternal.create(realmCore.userGetApp(_handle));
   }
 
+  /// Gets an [ApiKeyClient] instance that exposes functionality for managing
+  /// user API keys.
+  /// [API Keys Authentication Docs](https://docs.mongodb.com/realm/authentication/api-key/)
+  late final ApiKeyClient apiKeys = ApiKeyClient._(this);
```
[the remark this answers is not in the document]
nirinchev: I prefer `apiKeys` because it reads nicer - `user.apiKeys.create`. If we name that `apiKeysClient`, the same API will read like we're creating a client, so we'll need to then suffix all the APIs with `ApiKey` - like `createApiKey`, etc.

lib/src/user.dart
```diff
@@ -183,6 +200,102 @@ class UserProfile {
...
+  Future<void> _handle404Void(Future<void> future, {ObjectId? id}) async {
+    try {
+      await future;
+    } on AppException catch (e) {
+      if (e.statusCode == 404) {
+        if (id != null) {
+          throw AppInternal.createException("Failed to execute operation because ApiKey with Id: {id} doesn't exist.", e.linkToServerLogs, 404);
+        }
+
+        // If the id is null, we just ignore the exception
```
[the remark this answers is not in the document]
nirinchev: `delete` doesn't provide an id argument - the idea is that deleting a key that was already deleted is a no-op.

test/user_test.dart
nirinchev: This is a local function, not a class-level one, so prefixing it with an underscore is a lint violation.

src/realm_dart_sync.cpp
blagoev: I was going to suggest a fix today. Will write back shortly.

test/user_test.dart
blagoev: oh ok. missed that one.

lib/src/user.dart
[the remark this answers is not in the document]
blagoev: that's a good point. my point was that `apiKeys` reads like an array. `user.apiKeys`.  so one might expect `length` to exists `user.apiKeys.length`.

lib/src/user.dart
```diff
@@ -199,4 +283,39 @@ extension UserInternal on User {
...
 
   static User create(UserHandle handle, [App? app]) => User._(handle, app);
+
+  static ApiKey createApiKey(ObjectId id, String name, String? value, bool isEnabled) => ApiKey._(id, name, value, isEnabled);
+}
+
+extension on Future<void> {
+  Future<void> handle404({ObjectId? id}) async {
+    try {
+      await this;
```
blagoev, line 293: I think this is different than my suggestion. What bothers me is we are awaiting the future here instead of just attaching an error handler to it. 
So if we didn't need to handle 404 here we would be returning an unawaited Future. But now we do await before returning it.
```dart
return realmCore.deleteApiKey(_user, objectId);
```
versus
```dart
return await realmCore.deleteApiKey(_user, objectId);
```

blagoev: I suggest we do with my previous code suggestion where we attach onError handlers and still return the Future unawaited

blagoev: extension on Future<ApiKey>

blagoev (changes requested):

src/realm_dart_sync.cpp
nielsenko: Question.. for a while you reverted the `vector<char>` change.. why did you bring it back in b1e6f34a163a72d0996750b56442a61baf6ba2ae?

nirinchev: I was investigating failures on CI which turned out to be due to https://github.com/realm/realm-core/issues/5890. Once I found that out, I went back to strings.

blagoev: I did a fix PR here https://github.com/realm/realm-dart/pull/924.

lib/src/user.dart
nirinchev: Maybe I am missing something, but those seem functionally the same, why does it matter whether we use await or `onError`?

src/realm_dart_sync.cpp
nirinchev: The fix in #924 seems to be exactly what I'm doing here - how is it different?

lib/src/user.dart
nielsenko: https://dart.dev/guides/language/effective-dart/usage#prefer-asyncawait-over-using-raw-futures

blagoev: The difference is obvious . if you `await` you wait for the result of the future. If you don't await, you exit the method returning a future which can be awaited or chained even more before being awaited. If you notice we don't do any awaits in the SDK we return the futures unawaited in all other places. 
If you compare my suggestions with this one you will notice how clean my suggestion is. It is just chaining multiple futures and returning the last without awaiting it.

src/realm_dart_sync.cpp
blagoev: It is different since it uses std::string instead of manually handling strings in a std::vactor. It also makes explicit where exactly we copy the values and take pointers on them. If in the future someone needs to add another string to that buffer it is not error prone to do. 
This hear does not fix the original issue of taking pointers to freed memory.

nirinchev: I'm not using vectors in this PR? As you can see this comment is outdated because it was made on code that changed. The current implementation is using std::strings with the only difference being that I'm std::move-ing the original request, which is not a problem in and of itself, but I'd be happy to just copy it instead.

lib/src/user.dart
nirinchev: Your suggestion is functionally the same as the code that uses async/await. Whether it's cleaner or not is a matter of personal preference, but the compiled code should be virtually the same.

```dart
Future<void> doSomething() { ... }

Future<void> doSomethingWithAwait() {
  try {
    await doSomething();
  } catch (e) {
    print(e);
  }
}

Future<void> doSomethingWithOnError() {
  return doSomething().onError((e) {
    print(e);
  });
}
```

These two are exactly the same. They will return a future that will either complete as soon as `doSomething` completes or will print a message when `doSomething` fails.

src/realm_dart_sync.cpp
blagoev: hm, I did review on a vector version. Let me check the current code. We should try to sync somehow reviews with newer code changes.

blagoev: ok this one is a bit different. If you copy the request and then std::move it it will be the same as #924. I suggest we go with #924 since it already split the fix into a separate entity + it is fixing up the changelog.

lib/src/user.dart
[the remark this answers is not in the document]
blagoev: then the question might be why do we require it for `enable` and `disable` and not for `fetch/fetchAll` and `delete`. Why not make `enable/disable` no-op on non existing keys? Right now they will throw an exception

lib/src/user.dart
blagoev: There is a subtle difference between the two versions. In the first one you wait for the result and in the second you don't. We are using the second version in the SDK. What I am suggesting is to keep the consistency.

src/realm_dart_sync.cpp
```diff
@@ -30,22 +30,25 @@ RLM_API void realm_dart_http_request_callback(realm_userdata_t userdata, const r
...
     buf.headers_vector.reserve(request.num_headers);
     for (size_t i = 0; i < request.num_headers; i++) {
         auto [it, _] = buf.headers.emplace(request.headers[i].name, request.headers[i].value);
         buf.headers_vector.push_back({ it->first.c_str(), it->second.c_str() });
     }
-    request_copy.headers = buf.headers_vector.data();
 
     auto ud = reinterpret_cast<realm_dart_userdata_async_t>(userdata);
-    ud->scheduler->invoke([ud, request_copy = std::move(request_copy), buf = std::move(buf), request_context]() {
+    ud->scheduler->invoke([ud, request = std::move(request), buf = std::move(buf), request_context]() {
```
blagoev, line 43: I don't like we are invalidating a `const` argument here. I suggest we go with #924 where we copy the argument and then we move the local copy

lib/src/user.dart
nielsenko: For the fun of it I tried to decompile the generated code for the two functions. 

I only now how to do it for jit snapshots, and it would be better to compare the aot snapshots, but for the jit snapshots there is actually a difference in the amount of code generated, the `doSomethingWithOnError` being shorter.

I'm not sure if it has any impact on performance, or if it even holds true for aot compiled code (I'm trying to figure out how they do that - I know the dart sdk team can), but anyway.

As I think this is an interesting question, I'll try to clarify with the Dart SDK team.

For now, I don't think we should deviate from the [effective dart](https://dart.dev/guides/language/effective-dart/usage#prefer-asyncawait-over-using-raw-futures) recommendations, but @blagoev may have a point here, so I reserve the right to change my opinion later :-)

nielsenko: BTW, I asked for how to decompile AOT code here: https://github.com/dart-lang/sdk/issues/50059

nirinchev: > In the first one you wait for the result and in the second you don't

I'm not sure what that means. I dropped a note in the roundtable to discuss it as it'll probably be faster.

nirinchev: Okay, we decided to prefer `await` when we need to do continuations.

lib/src/user.dart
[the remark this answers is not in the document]
nirinchev: If you delete something that doesn't exist, then the end result is the same as if it existed - the thing doesn't exist. With `enable`/`disable`, there's an expectation that the thing exists after the call - with `enable`, you should even be able to login with the API key. So in that situation, throwing seems better since silently ignoring the error would mislead the developer into thinking everything is fine.

src/realm_dart_sync.cpp
nirinchev: Yeah, let's merge #924 as there's a lot of commentary by Mathias. I'll rebase this after.

src/realm_dart_sync.cpp
blagoev: Resolving this in favor of #924

desistefanova (changes requested):
Added some comments.
Changelog is missing.

test/credentials_test.dart
```diff
@@ -177,18 +167,14 @@ Future<void> main([List<String>? args]) async {
...
   });
 
   baasTest('Email/Password - retry custom confirmation for not registered user', (configuration) async {
     final app = App(configuration);
     final authProvider = EmailPasswordAuthProvider(app);
     String username = "${generateRandomString(5)}@realm.io";
-    expect(() async {
-      await authProvider.retryCustomConfirmationFunction(username);
-    }, throws<RealmException>("user not found"));
+    expectLater(() => authProvider.retryCustomConfirmationFunction(username), throws<AppException>("user not found"));
```
desistefanova, line 177:
```suggestion
    await expectLater(() => authProvider.retryCustomConfirmationFunction(username), throws<AppException>("user not found"));
```

test/credentials_test.dart
```diff
@@ -226,9 +212,7 @@ Future<void> main([List<String>? args]) async {
     final app = App(configuration);
     final authProvider = EmailPasswordAuthProvider(app);
     String username = "${generateRandomString(5)}@realm.io";
-    expect(() async {
-      await authProvider.resetPassword(username);
-    }, throws<RealmException>("user not found"));
+    expectLater(() => authProvider.resetPassword(username), throws<AppException>("user not found"));
```
desistefanova, line 215:
```suggestion
    await expectLater(() => authProvider.resetPassword(username), throws<AppException>("user not found"));
```

desistefanova (changes requested):
Some awaits were missing.

test/user_test.dart
```diff
@@ -116,4 +116,299 @@ Future<void> main([List<String>? args]) async {
...
+    expect(apiKeys.single, original);
+  });
+
+  baasTest('User.apiKeys.fetchAll with multiple keys returns all', (configuration) async {
+    final app = App(configuration);
+    final user = await getIntegrationUser(app);
+
+    final original = <ApiKey>[];
+    for (var i = 0; i < 5; i++) {
+      original.add(await user.apiKeys.create('my-api-key-$i'));
```
desistefanova, line 194: What if we created keys with same name?

test/user_test.dart
```diff
@@ -116,4 +116,299 @@ Future<void> main([List<String>? args]) async {
...
+
+  baasTest("User.apiKeys can't login with deleted key", (configuration) async {
+    final app = App(configuration);
+    final user = await getIntegrationUser(app);
+
+    final key = await user.apiKeys.create('my-key');
+
+    await user.apiKeys.delete(key.id);
+
+    final credentials = Credentials.apiKey(key.value!);
```
desistefanova, line 405: Isn't it strange for the API that `Credentials.apiKey(key.value!)` requires apiKey but we have to pass its value which is possible to be null?

desistefanova: We don't have a test for login with an `apiKey` after it has been received with `fetrch`. Will we have the `value` populated then? I can check it, it is just a reminder.

nirinchev: No, you can't fetch an API key value after it was created. The idea is that you create an API key and you store its value somewhere, then you use it to login. It's not expected that you have the full API key at this point, only the value.

test/user_test.dart
nirinchev: Good point, I added a test for that.

test/user_test.dart
desistefanova: Of course, this is for security reasons. I didn't think about that.

test/user_test.dart
```diff
@@ -116,4 +116,299 @@ Future<void> main([List<String>? args]) async {
...
+
+    await user.apiKeys.enable(key.id);
+
+    final apiKeyUser = await app.logIn(credentials);
+    expect(apiKeyUser.provider, AuthProviderType.apiKey);
+    expect(apiKeyUser.id, user.id);
+    expect(apiKeyUser.refreshToken, isNot(user.refreshToken));
+  });
+
+  baasTest("User.apiKeys can't login with disabled key", (configuration) async {
```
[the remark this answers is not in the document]
nirinchev: Yeah, I originally wanted to separate them so they validate each their own thing, but on a second thought, I don't think it matters that much, so removed the "disabled key" one.

lib/src/user.dart
```diff
@@ -76,6 +81,18 @@ class User {
...
 
+  /// Gets the refresh token for this [User]. This is the user's credential for
+  /// accessing Atlas App Services and should be treated as sensitive data.
+  String get refreshToken {
+    return realmCore.userGetRefreshToken(this);
+  }
+
+  /// Gets the access token for this [User]. This is the user's credential for
+  /// accessing Atlas App Services and should be treated as sensitive data.
+  String get accessToken {
```
desistefanova, line 92: It will be good to tests also accessToken in the expected results of some of the tests.

desistefanova (commented):
I didn't find any test for serverApiKey. It would be nice to have. I am not sure if it is possible to generate any key. Maybe we can do a manual test, just to have it.

lib/src/user.dart
nirinchev: Using the access token is a bit of a pain - I can test that it's not empty, but anything more than that will be more complicated than I'd want it to be.

desistefanova: I think it is enough, we will be sure that the get method works.

lib/src/cli/atlas_apps/baas_client.dart
```diff
@@ -353,6 +355,16 @@ class BaasClient {
     }
   }
 
+  Future<String> createApiKey(BaasApp app, String name, bool enabled) async {
+    final dynamic result = await _post('groups/$_groupId/apps/${app.appId}/api_keys', '{ "key":"", "name":"$name", "disabled": false }');
+    if (!enabled) {
+      // We need to disable the key with another API call - the argument in the create call is ignored.
```
desistefanova, line 361: Is it a bug in Atlas API?

test/user_test.dart
```diff
@@ -116,4 +116,332 @@ Future<void> main([List<String>? args]) async {
...
+
+  baasTest("Credentials.serverApiKey can login user", (configuration) async {
+    final app = App(configuration);
+
+    final apiKey = await createServerApiKey(app, ObjectId().toString());
+    final credentials = Credentials.serverApiKey(apiKey);
+
+    final apiKeyUser = await app.logIn(credentials);
+
+    // This is due to the way Core stores the user provider: https://github.com/realm/realm-core/blob/7b087008a19773f1bb92324d410f9e70db17fca9/src/realm/object-store/sync/app_credentials.cpp#L33-L34
```
desistefanova, line 417: I think it is correct to create an issue for Core.

test/user_test.dart
```diff
@@ -116,4 +116,332 @@ Future<void> main([List<String>? args]) async {
...
+    expect(apiKeyUser.provider, AuthProviderType.apiKey);
```
desistefanova, line 418:
```suggestion
expect(apiKeyUser.provider, AuthProviderType.apiKey);
expect(apiKeyUser.state, UserState.loggedIn);
```

test/user_test.dart
nirinchev: I asked on #realm-core-team - I was hoping to get a response, but can file a ticket as well.

lib/src/cli/atlas_apps/baas_client.dart
nirinchev: Maybe it's intentional - I simplified the original call though and removed that comment.

test/user_test.dart
nirinchev: Filed https://github.com/realm/realm-core/issues/5914.

CHANGELOG.md
```diff
@@ -38,6 +38,11 @@
   });
   ```
 
+* Added support for API key authentication. (Issue [#432](https://github.com/realm/realm-dart/issues/432))
```
desistefanova, line 41: This is handled in the same PR
```suggestion
* Added support for API key authentication. (Issue [#432](https://github.com/realm/realm-dart/issues/432)), (PR [#919](https://github.com/realm/realm-dart/pull/919))
```

CHANGELOG.md
```diff
@@ -38,6 +38,11 @@
...
+  * Expose `User.apiKeys` client - this client can be used to create, fetch, and delete API keys.
+  * Expose `Credentials.apiKey` and `Credentials.serverApiKey` that enable authentication with API keys.
+* Exposed `User.accessToken` and `User.refreshToken` - these tokens can be used to authenticate against the server when calling HTTP API outside of the dart/flutter SDK. For example, if you want to use the GraphQL or the data access API. (PR [#919](https://github.com/realm/realm-dart/pull/919))
```
desistefanova, line 44:
```suggestion
* Exposed `User.accessToken` and `User.refreshToken` - these tokens can be used to authenticate against the server when calling HTTP API outside of the Dart/Flutter SDK. For example, if you want to use the GraphQL or the data access API. (PR [#919](https://github.com/realm/realm-dart/pull/919))
```

desistefanova (approved):

CHANGELOG.md
desistefanova: I see. OK.

nirinchev:
@desistefanova I've decided not to expose server API key since it's the same as user api key - see @ironage's comment here: https://github.com/realm/realm-core/issues/5914#issuecomment-1263949337. LMK if that makes sense to you. I can confirm that the server api key tests pass with the api key provider, so seems superficial indeed.

desistefanova:
> @desistefanova I've decided not to expose server API key since it's the same as user api key - see @ironage's comment here: [realm/realm-core#5914 (comment)](https://github.com/realm/realm-core/issues/5914#issuecomment-1263949337). LMK if that makes sense to you. I can confirm that the server api key tests pass with the api key provider, so seems superficial indeed.
> 
> @blagoev I merged this with master, so it's no longer touching the native implementation.

But it is possible in the Atlas web UI to create API key that doesn't belong to any user. How the developers will be able to use this key?

nirinchev:
They can use `Credentials.apiKey` - apparently there's no difference in the two credential types. You can see the last couple of tests in `user_test.dart` are logging in a user with a server API key.

desistefanova (approved):

blagoev (approved):
Not sure about the `expectLater` calls. Should we always await the action or that is not mandatory?

test/user_test.dart
```diff
@@ -116,4 +116,319 @@ Future<void> main([List<String>? args]) async {
...
+    final user = await getIntegrationUser(app);
+
+    final key = await user.apiKeys.create('my-key');
+
+    await user.apiKeys.delete(key.id);
+
+    final credentials = Credentials.apiKey(key.value!);
+
+    await expectLater(
+        () => app.logIn(credentials),
```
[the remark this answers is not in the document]
nirinchev: Resolved offline - it's fine to leave it like this since the `expectLater` will await the future returned by `logIn`.

PR #920, opened 2022-09-23
Support encryption key in configuration

### LINKED ISSUE #88: Config: Support encryption keys
lib/src/configuration.dart

// /// The key used to encrypt the entire [Realm].
// ///
// /// A full 64byte (512bit) key for AES-256 encryption.
// /// Once set, must be specified each time the file is used.
// final List<int>? encryptionKey;

desistefanova:
Support encryption key used to encrypt the entire [Realm].
Its length is forced to be 64bytes.
If null encryption is not enabled.
Fixes #88


lib/src/native/realm_core.dart
```diff
@@ -225,7 +225,9 @@ class _RealmCore {
       } else if (config is DisconnectedSyncConfiguration) {
         _realmLib.realm_config_set_force_sync_history(configPtr, true);
       }
-
+      if (config.encryptionKey != null && config is! InMemoryConfiguration) {
```
nirinchev, line 228: How can we set `config.encryptionKey` on an `InMemoryConfiguration`? I'd rather have this be an assert rather than silently ignoring it.

lib/src/configuration.dart
```diff
@@ -57,6 +57,9 @@ typedef MigrationCallback = void Function(Migration migration, int oldSchemaVers
 /// Configuration used to create a [Realm] instance
 /// {@category Configuration}
 abstract class Configuration implements Finalizable {
+
+  static const int encryptionKeySize = 64;
```
nirinchev, line 61: I don't think this needs to be public.

test/realm_test.dart
```diff
@@ -824,6 +824,71 @@ Future<void> main([List<String>? args]) async {
     expect(stored.location, now.location);
     expect(stored.location.name, 'Europe/Copenhagen');
   });
+
+  test('Realm open with wrong encryption key', () {
```
nirinchev, line 828: These are all testing that a Realm that is already opened with 1 encryption key cannot be opened. But we're not testing a scenario where we open, then close, then try to open again. You could probably parameterize the tests to avoid having to copy-paste them all.

nirinchev (approved):
Looks great, just a few suggestions and would like to see more tests.

lib/src/configuration.dart
desistefanova: I use it inside _RealmCore and and in the tests also.

test/realm_test.dart
desistefanova: This test covers this scenario "Realm - open local encrypted realm with encryption key". Yes, I could think about parameterizing.

lib/src/configuration.dart
nirinchev: Sure, but I don't think it justifies adding an extra public member.

test/realm_test.dart
nirinchev: The issue is that it doesn't cover the case where the Realm is not open when you provide the invalid encryption key.

lib/src/configuration.dart
```diff
@@ -137,10 +145,11 @@ abstract class Configuration implements Finalizable {
         schemaVersion: schemaVersion,
         fifoFilesFallbackPath: fifoFilesFallbackPath,
         path: path,
+          encryptionKey: encryptionKey,
```
nielsenko, line 148: formatting seems a bit off

nielsenko (approved):

test/realm_test.dart
desistefanova: OK, I see. It is another special case :). I added the test.

lib/src/configuration.dart
desistefanova: I will think where I can move it.

test/configuration_test.dart
```diff
@@ -527,7 +527,29 @@ Future<void> main([List<String>? args]) async {
     realm.close();
 
     final disconnectedSyncConfig = Configuration.disconnectedSync(schema, path: realmPath);
-    final disconnectedRealm = Realm(disconnectedSyncConfig);
+    final disconnectedRealm = getRealm(disconnectedSyncConfig);
     expect(disconnectedRealm.find<Task>(oid), isNotNull);
   });
+
+  test('Configuration set short encryption key', () {
```
nirinchev, line 534: What would happen if I create an encryption key containing numbers that don't fit into a byte? E.g. `List<int>.generate(encryptionKeySize, (i) => random.nextInt(1024))`?

nirinchev (approved):
One final question from me, otherwise it looks good.

lib/src/configuration.dart
```diff
@@ -92,9 +92,10 @@ abstract class Configuration implements Finalizable {
     this.fifoFilesFallbackPath,
     this.encryptionKey,
   }) {
-    if (encryptionKey != null && encryptionKey!.length != realmCore.encryptionKeySize) {
-      throw RealmException("EncryptionKey must be 64 bytes");
+    if (encryptionKey != null && encryptionKey!.isNotEmpty && encryptionKey?.length != realmCore.encryptionKeySize) {
+      throw RealmException("Wrong encryption key size (must be 0 or ${realmCore.encryptionKeySize})");
```
nirinchev, line 96: Why do we accept empty keys here?

test/configuration_test.dart
desistefanova: It is a good point. I added some tests with generated list with exceeding byte size items. It seems to be working. I suppose they are truncated and you are even able to open realm twice with the same key. I'm not sure whether we to throw an exception or to leave it like this.

lib/src/configuration.dart
desistefanova: They do the same in C-API.
I just do the same validation in the Configuration constructor, since we call the core function later on creating a realm with this config.
```
 if (key_size != 0 && key_size != 64) {
     throw std::logic_error{"Wrong encryption key size (must be 0 or 64)"};
 }
```

test/configuration_test.dart
nirinchev: Hm... interesting. I think we may want to validate it to make sure people are not supplying nonsensical data, but perhaps I'm being overly cautious.

desistefanova: I think it is not a good idea to add `Uint8List` in the external API, since it requires `typed_data.dart`.

lib/src/configuration.dart
desistefanova: OK. I agree.

test/configuration_test.dart
nirinchev: Oh, I didn't mean that - more like, in the configuration constructor, do something like:

```dart
  Configuration._(
    this.schemaObjects, {
    String? path,
    this.fifoFilesFallbackPath,
    this.encryptionKey,
  }) {
    _validateEncryptionKey(encryptionKey);

    this.path = path ?? _path.join(_path.dirname(_defaultPath), _path.basename(defaultRealmName));
  }

  void _validateEncryptionKey(List<int>? key) {
    if (key == null) {
      return;
    }

    if (key.length != realmCore.encryptionKeySize) {
      throw RealmException("Wrong encryption key size (must be ${realmCore.encryptionKeySize}, but was ${key.length})");
    }

    final notAByteElement = key.firstWhereOrNull((e) => e > 255);
    if (notAByteElement != null) {
      throw RealmException("Encryption key must be a list of bytes, but found $notAByteElement at index ${key.indexOf(notAByteElement)}");
    }
  }
```

desistefanova: Done.

lib/src/native/realm_core.dart
```diff
@@ -225,7 +227,10 @@ class _RealmCore {
       } else if (config is DisconnectedSyncConfiguration) {
         _realmLib.realm_config_set_force_sync_history(configPtr, true);
       }
-
+      if (config.encryptionKey != null) {
+        assert(config is! InMemoryConfiguration, "Encryption keys are not allowed for InMemoryConfiguration");
```
blagoev, line 231: This check is debug only. It will not happen in release builds. Is this the intention?

desistefanova: Yes, it is by purpose. Actually, there is no way the user to do that, because there is no `encryptionKey ` parameter in `Configuration.inMemory` method.

blagoev: I suggest we remove this then. Core will also ignore this if its an inmemory one

nirinchev: I think asserting is valuable here as it clarifies the behavior. The type system should prevent setting this value, but someone 2 years from now may decide to refactor the configurations and accidentally leak the encryption key into the in memory config and hit that assert.

blagoev: how can you leak a non existing property. Also if someone does that it will still not work in Core. I think we don't need to litter the code with such extra sanity checks which can never happen cause of the API shape.  
Having asserts like that makes sense only if there is a possibility in the API to pass an invalid argument into a function.

blagoev: Yes that is my point. If we can not set through the API, then the `assert` is meaningless. unlike a sanity check for a value which might be returned by a function. And even if we set it in the future unintentionally by a refactoring we what we are doing is duplicating code that is already in Core.

desistefanova: I will remove the assert.

richard457:
I am writing to request clarification on the encryption key when passed, when added to realm later will it encrypt the existing realm?

richard457:
also the example from docs shows encrypting local realm
```dart
// Generate encryption key
--
final key = List<int>.generate(64, (i) => Random().nextInt(256));
 
final encryptedConfig = Configuration.local([Car.schema],
// Include the encryption key in the configuration
encryptionKey: key);
final encryptedRealm = Realm(encryptedConfig);
```
does it work on flexible sync as well @desistefanova

nirinchev:
Yes, it works with flexible sync and no - you cannot change the encryption of a file after it was created. You can use `Realm.writeCopy` to create an encrypted copy of the file, but that only works for local files. For synchronized realms, you can just recreate the file and re-fetch the data.

PR #924, opened 2022-09-26
fix copy of structs through the scheduler.

blagoev:


src/realm_dart_sync.cpp
```diff
@@ -33,19 +33,19 @@ RLM_API void realm_dart_http_request_callback(realm_userdata_t userdata, const r
     realm_http_request_t request_copy = request; // copy struct
```
nirinchev, line 33: You don't need to copy the request here - you can `std::move` the request and create the copy inside the scheduler callback.

blagoev: I am not sure about that. The request is `const` arg so we should not `std::move` it

CHANGELOG.md
```diff
@@ -2,14 +2,17 @@
 
 **This project is in the Beta stage. The API should be quite stable, but occasional breaking changes may be made.**
 
+### Breaking Changes
+* Fixed an issue that would cause passwords sent to the server (e.g. `Credentials.EmailPassword` or `EmailPasswordAuthProvider.registerUser`) to contain an extra empty byte at the end. (PR [#918](https://github.com/realm/realm-dart/pull/918)).
+  Notice: Any existing email users might need to be recreated because of this breaking change.
```
blagoev, line 7: @nirinchev Added this notice here. Any good?

src/realm_dart_sync.cpp
nirinchev: `std::move`-ing doesn't guarantee that a move will occur, it only tells the compiler it _may_ move it.

CHANGELOG.md
nirinchev: Sure, sounds good

src/realm_dart_sync.cpp
blagoev: yes and accessing the old reference is undefined behavior. Hence I am hesitant to do what you suggest

nirinchev: We're not accessing the old reference though, so I'm not sure I see the problem. If you do prefer to eagerly copy here, then we don't need to copy it again on line 45.

blagoev:
@nirinchev seems like this change is not good. Will convert to draft until it is ready

src/realm_dart_sync.cpp
blagoev: My point is a `const` arg should not be modified. And by moving it we will invalidate it.

blagoev: on line 45 we do need to copy it again since it's not modifiable since its an rvalue.

src/realm_dart_sync.cpp
```diff
@@ -33,20 +33,20 @@ RLM_API void realm_dart_http_request_callback(realm_userdata_t userdata, const r
...
     for (size_t i = 0; i < request.num_headers; i++) {
         auto [it, _] = buf.headers.emplace(request.headers[i].name, request.headers[i].value);
         buf.headers_vector.push_back({ it->first.c_str(), it->second.c_str() });
     }
-    request_copy.headers = buf.headers_vector.data();
 
     auto ud = reinterpret_cast<realm_dart_userdata_async_t>(userdata);
     ud->scheduler->invoke([ud, request_copy = std::move(request_copy), buf = std::move(buf), request_context]() {
-        (reinterpret_cast<realm_http_request_func_t>(ud->dart_callback)(ud->handle, request_copy, request_context));
+        realm_http_request_t request = std::move(request_copy);
```
nirinchev, line 45: Why is this copy necessary?

blagoev:
@nirinchev I fixed the issue and opening this again. I believe the CI fails are not related to this change

src/realm_dart_sync.cpp
blagoev: on line 45 we do need to copy it again since it's not modifiable since its an rvalue.

blagoev: In other words you get this error if you try to modify `request_copy.url`
`error C3490: 'url' cannot be modified because it is being accessed through a const object`

blagoev: Btw this ` realm_http_request_t request = std::move(request_copy);` is not a copy

src/realm_dart_sync.cpp
blagoev: pardon me, on line 45 its not a copy. It's a move so as efficient as it can be.

src/realm_dart_sync.cpp
nirinchev: I'm not a c++ expert, but my understanding is that you cannot std::move a const object. So if the error you get is that you're accessing a const object, then `std::move`-ing it will just do a copy.

CHANGELOG.md
```diff
@@ -37,20 +40,24 @@
...
-  As an example
+  Example
   ```dart
-  expect(realm.query<Person>('FALSEPREDICATE').query('TRUEPREDICATE'), isEmpty);
+  expect(realm.query<Person>('FALSEPREDICATE').query('TRUEPREDICATE'), isEmpty); //<-- Fails if a Persion object exists
   ```
-  would fail if any Persons exists
-* Fixed an issue that would cause passwords sent to the server (e.g. `Credentials.EmailPassword` or `EmailPasswordAuthProvider.registerUser`) to contain an extra empty byte at the end. (PR [#918](https://github.com/realm/realm-dart/pull/918))
-* Added support for realm list of nullable primitive types, ie. `RealmList<int?>`. ([#163](https://github.com/realm/realm-dart/issues/163))
+* Fixed copying of native structs for session errors and http requests. ([#924](https://github.com/realm/realm-dart/issues/924))
```
desistefanova, line 60: Maybe it will be nice to explain here in the Changelog what the fixed problem was. Did we have an error C3490 or something else? Do we have some issue that had to be attached? Reading the PR I don't understand why we need this change unles I following the comments.

desistefanova, line 60: Maybe it will be nice to explain here in the Changelog what the fixed problem was. Did we have an error C3490 or something else? Do we have some issue that had to be attached? This link ([#924](https://github.com/realm/realm-dart/issues/924)) point to the same PR. Reading the PR I don't understand why we need this change unles I following the comments.

src/realm_dart_sync.cpp
blagoev: std::move is not a copy its more of a const_cast

CHANGELOG.md
blagoev: its mostly an internal fix. not sure how can we improve the messaging here. It is fine to link to the PR in the changelog and we do it in many places. We don't have an issue for this fix.

blagoev: Updated the PR description

desistefanova: OK. I thought the link was broken because it pointed to the issues 'realm-dart/issues/924' instead of to the pulls 'realm-dart/pull/924'. It is still working, but should be 'pull'.

desistefanova (approved):

CHANGELOG.md
blagoev: that's good point. will fix

src/realm_dart_sync.cpp
nielsenko: Hmm.. to my understanding `std::move` on a const object will copy it. There was a discussion on stackoverflow about the same https://stackoverflow.com/questions/28595117/why-can-we-use-stdmove-on-a-const-object. Perhaps @RedBeard0531 can way in?

nielsenko: I see @nirinchev already posted the exact same link, sorry 🤦

.vscode/settings.json
```diff
@@ -63,6 +63,85 @@
     "xstring": "cpp",
     "xutility": "cpp",
     "iosfwd": "cpp",
-    "sstream": "cpp"
+    "sstream": "cpp",
```
RedBeard0531, line 66: You may want to try a plugin like [this](https://marketplace.visualstudio.com/items?itemName=chrislajoie.vscode-modelines) which can read modelines, rather than trying to keep a list of headers. I can't vouch for this specific plugin, but I use [something similar](https://github.com/sfiera/vim-emacsmodeline) in vim to deal with this and it works well.

src/realm_dart_sync.cpp
RedBeard0531: You can just make the argument not be `const`. Top-level `const` on arguments is ignored when deciding the function's type, so `void(foo)` and `void(const foo)` have the same type and can both be assigned to a `void(*)(foo)` function pointer. The `const` is only meaningful inside the implementation of the function as it makes the parameter value behave like a const object. In this case, that is probably something you don't want.

> std::move-ing doesn't guarantee that a move will occur, it only tells the compiler it may move it.

Sort of. `std::move` makes its result a (possibly const) rvalue. That then can affect overload resolution specifically to choose overloads that then move from the object. All of this is specifically defined, and no part of it is up to the compiler (except for the poorly named copy-elision optimization where it is allowed to completely skip both copies and moves in some cases).

> yes and accessing the old reference is undefined behavior.

Not quite, although it is perhaps a good rule of thumb to pretend this is the case most of the time. A) it isn't the `std::move` that causes this affect, it is whether a move actually occurs. B) For all standard library types, you are guaranteed that a moved-from object will be left in a "valid but unspecified state". This means that you can call any methods that don't have preconditions. For user-defined types, unless otherwise stated, it is best to assume that you can only destroy the object or assign over it. "Named assignments" such as `clear()` or `reset()` are also generally considered acceptable.

(As I said, only using `std::move` on object you don't touch again is a good safe rule of thumb if you don't want to think too hard. But this thread was already getting into the weeds, so I wanted to make sure there was a precise statement of what the real rules are)

src/realm_dart_sync.cpp
RedBeard0531: The real answer here is that you want to add `mutable` to the lambda. Without it, all by-value captures, including `request_copy` are considered as `const` inside the body of the lambda. So `request` is silently making a copy, even though there is a `std::move`. You probably should also remove the const from the original `request` parameter, and the `request_copy` variable and just capture `request` as ~`request = std::move(request)`~~ simply `request` (see next paragraph). And if it the lambda is marked mutable, you can just modify it in place without making another copy. Since the callback will only be called once this is fine.

Oh. I just noticed that `realm_http_request_t` is the C API type. For C types (or any other "Trivially Copiable" type in C++) there is no difference between move and copy. So all of this discussion of move vs copy is moot, and you might as well just copy since it is syntactically lighter. I'm leaving what I've already written about `std::move` in the hope that it is informative.

> std::move is not a copy its more of a const_cast

To be pedantic, `const_cast` can only add or remove `const` and `volatile` qualifiers. `std::move` is specifically a `static_cast` that adds turns an lvalue into an rvalue. It does NOT remove the `const`. But it also doesn't copy (although the thing you pass the result to may make a copy of the referenced object).

> my understanding is that you cannot std::move a const object.

You _can_ `std::move` a const object, but you probably `shouldn't`. It produces a `const T&&` which is a bit of a weird thing. While a type _could_ have a special constructor just for that, in general it will select the copy constructor during overload resolution and make a copy. So the `move` isn't harmful (in that it doesn't change the behavior of the program, although it may make more template instantiations and potentially slow down compiles or bloat code size), but it is misleading. We could consider making our own `realm::move` that will cause a compiler error on const objects so that you know to either remove the `const` or remove the no-op `move`. The reason `std::move` allows passing `const` object is both for consistency, and because it is useful in templates where there are some cases where you want to move from non-const objects but obviously you can't from const ones, but you don't know what you will be passed and want to work with either. But that isn't something we do often.

RedBeard0531 (changes requested):

src/realm_dart_sync.cpp
nielsenko: I knew there be dragons here.. Thanks you so much @RedBeard0531 🙏 

To be pragmatic, could you perhaps write a version of 
```c++
RLM_API void realm_dart_http_request_callback(realm_userdata_t userdata, const realm_http_request_t request, void* request_context)
```
as you think it should be written? So we have a hands on example of how theory manifest in practice?

nirinchev: Based on my understanding, would something like this be fine:

```cpp
    ud->scheduler->invoke([ud, request, buf = std::move(buf), request_context]() {
        realm_http_request_t request_copy = request;
        request_copy.headers = buf.headers_vector.data();
        request_copy.url = buf.url.c_str();
        request_copy.body = buf.body.data();

        (reinterpret_cast<realm_http_request_func_t>(ud->dart_callback)(ud->handle, request_copy, request_context));
    });
```

Notably, we're capturing the original `request` and then creating the copy inside the callback.

RedBeard0531: Also, if you add `mutable` to the lambda, you can get rid of the `realm_copy` variable.

BTW, if you are OK writing this part in C++, why not just implement the C++ `GenericNetworkTransport` API directly?

src/realm_dart_sync.cpp
blagoev: Thanks @RedBeard0531. Good insights. I was with the impression that `const args` will help the compiler in someway optimize the function invocation. I will remove the const from the arg there.

.vscode/settings.json
blagoev: I am unsure if this will help. This is list is kept by VS Code for some reason and is constantly updated. I suspect it is the cmake extension that does this. Its not something we are doing manually.

src/realm_dart_sync.cpp
blagoev: Will fix

blagoev: Btw the C-API definition of the http request is this one 
```cpp
 /**
 * Callback function used by Core to make a HTTP request.
 *
 * Complete the request by calling realm_http_transport_complete_request(),
 * passing in the request_context pointer here and the received response.
 * Network request are expected to be asynchronous and can be completed on any thread.
 *
 * @param request The request to send.
 * @param request_context Internal state pointer of Core, needed by realm_http_transport_complete_request().
 */
typedef void (*realm_http_request_func_t)(realm_userdata_t userdata, const realm_http_request_t request,
                                          void* request_context);
```

Shouldn't that be a `const realm_http_request_t& request` instead. Why do they insist on copying the request object while invoking the function.

This is how they invoke it https://github.com/realm/realm-core/blob/master/src/realm/object-store/c_api/http.cpp#L60-L63. 

So why not pass the second argument by reference?

blagoev: @RedBeard0531 I did the fix. Could you please, check the PR again. Also this message above :)

src/realm_dart_sync.cpp
```diff
@@ -21,32 +21,29 @@
...
     for (size_t i = 0; i < request.num_headers; i++) {
         auto [it, _] = buf.headers.emplace(request.headers[i].name, request.headers[i].value);
         buf.headers_vector.push_back({ it->first.c_str(), it->second.c_str() });
     }
-    request_copy.headers = buf.headers_vector.data();
 
     auto ud = reinterpret_cast<realm_dart_userdata_async_t>(userdata);
-    ud->scheduler->invoke([ud, request_copy = std::move(request_copy), buf = std::move(buf), request_context]() {
-        (reinterpret_cast<realm_http_request_func_t>(ud->dart_callback)(ud->handle, request_copy, request_context));
+    ud->scheduler->invoke([ud, request = std::move(request), buf = std::move(buf), request_context]() mutable {
```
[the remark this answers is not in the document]
blagoev: That is true for `buf` as well. I kept it for consistency. Otherwise the who reads it will wonder why we are not moving request but we are moving buf. Also for c++ I prefer being explicit.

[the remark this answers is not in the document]
nirinchev: If it doesn't do anything, then let's remove it from `buf` as well - no point in pretending we move something.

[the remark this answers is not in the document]
blagoev: I prefer to leave that so it is more understandable what the intention is. Otherwise you really need to understand there is no difference in this particular case.

src/realm_dart_sync.cpp
```diff
@@ -69,20 +66,18 @@ RLM_API void realm_dart_sync_error_handler_callback(realm_userdata_t userdata, r
...
     buf.user_info_vector.reserve(error.user_info_length);
     for (size_t i = 0; i < error.user_info_length; i++) {
         auto [it, _] = buf.user_info_map.emplace(error.user_info_map[i].key, error.user_info_map[i].value);
         buf.user_info_vector.push_back({ it->first.c_str(), it->second.c_str() });
     }
-    error.user_info_map = buf.user_info_vector.data();
 
     auto ud = reinterpret_cast<realm_dart_userdata_async_t>(userdata);
-    ud->scheduler->invoke([ud, session = *session, error = std::move(error), buf = std::move(buf)]() {
+    ud->scheduler->invoke([ud, session = *session, error = std::move(error), buf = std::move(buf)]() mutable {
```
RedBeard0531, line 77: I think the std::move of error is meaningless here as well. In fact all comments carry over.

I am amused that this code is basically identical to the above, but unfortunately, due to the structure of the c api, it is just different enough that I don't think there is a clean way to avoid the code duplication.

src/realm_dart_sync.cpp
```diff
@@ -21,32 +21,29 @@
...
 #include "realm_dart_sync.h"
 
-RLM_API void realm_dart_http_request_callback(realm_userdata_t userdata, const realm_http_request_t request, void* request_context) {
-    // the pointers in error are to stack values, we need to make copies and move them into the scheduler invocation
+RLM_API void realm_dart_http_request_callback(realm_userdata_t userdata, realm_http_request_t request, void* request_context) {
+    // the pointers in request are to stack values, we need to make copies and move them into the scheduler invocation
     struct request_copy_buf {
         std::string url;
         std::string body;
         std::map<std::string, std::string> headers;
```
RedBeard0531, line 29: FYI, it is probably better to use a `std::deque<std::pair<std::string, std::string>>` here. You could also use a `std::vector<...>` as long as you `reserve()` the correct size up front, or do two passes, one to push everything, and another to push collect the addresses of the strings. You just need to avoid a growth event after collecting the addresses.

src/realm_dart_sync.cpp
```diff
@@ -21,32 +21,29 @@
...
         auto [it, _] = buf.headers.emplace(request.headers[i].name, request.headers[i].value);
         buf.headers_vector.push_back({ it->first.c_str(), it->second.c_str() });
     }
-    request_copy.headers = buf.headers_vector.data();
 
     auto ud = reinterpret_cast<realm_dart_userdata_async_t>(userdata);
-    ud->scheduler->invoke([ud, request_copy = std::move(request_copy), buf = std::move(buf), request_context]() {
-        (reinterpret_cast<realm_http_request_func_t>(ud->dart_callback)(ud->handle, request_copy, request_context));
+    ud->scheduler->invoke([ud, request = std::move(request), buf = std::move(buf), request_context]() mutable {
+        request.url = buf.url.c_str();
```
RedBeard0531, line 43: Please add a comment stating that you are only updating the addresses due to moves, and that all other fields of the request, such as sizes do not need to be modified.

Alternatively, since you know the sizes of everything, You might as well just recreate the whole `request` from scratch rather than moving it around. Your call though.

src/realm_dart_sync.cpp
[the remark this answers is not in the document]
RedBeard0531: Fair enough. I am not maintaining this code, so if the people who will be think this will be more maintainable, I won't stand in your way.

RedBeard0531:
All comments are optional suggestions. The code looks correct to me now, at least as far as I can see in a quick review.

src/realm_dart_sync.cpp
blagoev: ok I spoke with Yavor about this. seems no point on doing that. there will be not much gain

src/realm_dart_sync.cpp
```diff
@@ -26,24 +27,25 @@ RLM_API void realm_dart_http_request_callback(realm_userdata_t userdata, realm_h
...
     buf.url = request.url;
     buf.body = std::string(request.body, request.body_size);
-    buf.headers_vector.reserve(request.num_headers);
+    buf.headers_values.reserve(request.num_headers);
+    buf.headers.reserve(request.num_headers);
     for (size_t i = 0; i < request.num_headers; i++) {
-        auto [it, _] = buf.headers.emplace(request.headers[i].name, request.headers[i].value);
-        buf.headers_vector.push_back({ it->first.c_str(), it->second.c_str() });
+        auto& ref = buf.headers_values.emplace_back(request.headers[i].name, request.headers[i].value);
+        buf.headers.push_back({ ref.first.c_str(), ref.second.c_str() });
```
RedBeard0531, line 40:
```suggestion
        auto& [name, value] = buf.headers_values.emplace_back(request.headers[i].name, request.headers[i].value);
        buf.headers.push_back({ name.c_str(), value.c_str() });
```

I find this clearer. Your choice if you want to make this change. This was mostly an FYI in case you didn't know it was possible.

blagoev: nice! I will use it

PR #925, opened 2022-09-26
Client reset

### LINKED ISSUE #353: Sync Client Config: resync mode

### LINKED ISSUE #810: Client Reset
callbacks before and after, complex tests

desistefanova:
Implements [Client Resets](https://www.mongodb.com/docs/atlas/app-services/sync/error-handling/client-resets/) support.
Added argument `clientResetHandler` when creating `FlexibleSyncConfiguration`. There are 4 possible handles that defines the mode of processing data in case of client reset:

- RecoverOrDiscardUnsyncedChangesHandler
- RecoverUnsyncedChangesHandler
- DiscardUnsyncedChangesHandler
- ManualRecoveryHandler

The default mode is RecoverOrDiscard.
Tests are added.

Fixes #353 
Fixes #810


lib/src/configuration.dart
```diff
@@ -162,7 +162,8 @@ abstract class Configuration implements Finalizable {
     String? fifoFilesFallbackPath,
     String? path,
     SyncErrorHandler syncErrorHandler = defaultSyncErrorHandler,
-    SyncClientResetErrorHandler syncClientResetErrorHandler = const ManualSyncClientResetHandler(_defaultSyncClientResetHandler),
+    SyncClientResetErrorHandler syncClientResetErrorHandler = const RecoverOrDiscardSyncClientResetHandler(_defaultSyncClientResetHandler),
+    ClientResyncMode clientResyncMode = ClientResyncMode.recoverOrDiscard,
```
nirinchev, line 166: This should not be exposed.

nirinchev (changes requested):

lib/src/configuration.dart
desistefanova: Oh, right. What about having only a getter of `clientResyncMode`?

desistefanova: OK, I will change it.

blagoev (approved):

lib/src/configuration.dart
```diff
@@ -369,8 +369,12 @@ class RealmSchema extends Iterable<SchemaObject> {
...
 /// The signature of a callback that will be invoked if a client reset error occurs for this [Realm].
 ///
-/// Currently, Flexible Sync supports only the [ManualSyncClientResetHandler].
-class SyncClientResetErrorHandler {
+/// A Client Resync is triggered if the device and server cannot agree
+/// on a common shared history for the Realm file,
+/// thus making it impossible for the device to upload or receive any changes.
+/// This can happen if the server is rolled back or restored from backup.
+/// {@category Sync}
+abstract class SyncClientResetErrorHandler {
```
nirinchev, line 377:
```suggestion
abstract class ClientResetHandlerBase {
```

I don't think we need to prefix that with `Sync` because there's no other component that can cause client resets. I also don't think we need the `error` in there because it's somewhat evident.

lib/src/configuration.dart
```diff
@@ -379,4 +383,69 @@ class SyncClientResetErrorHandler {
 }
 
 /// A client reset strategy where the user needs to fully take care of a client reset.
-typedef ManualSyncClientResetHandler = SyncClientResetErrorHandler;
+///
+/// If you set [ManualSyncClientResetHandler] callback as `syncClientResetErrorHandler` argument of [Configuration.flexibleSync],
+/// that will enable full control of moving unsynced changes to synced realm.
+/// {@category Sync}
+class ManualSyncClientResetHandler extends SyncClientResetErrorHandler {
```
nirinchev, line 390:
```suggestion
class ManualRecoveryHandler extends SyncClientResetErrorHandler {
```

lib/src/configuration.dart
```diff
@@ -379,4 +383,69 @@ class SyncClientResetErrorHandler {
...
+}
+
+/// A client reset strategy where all the not yet synchronized data is automatically
+/// discarded and a fresh copy of the synchronized Realm is obtained.
+///
+/// If you set [DiscardLocalSyncClientResetHandler] callback as `syncClientResetErrorHandler` argument of [Configuration.flexibleSync],
+/// the local Realm will be discarded and replaced with the server side Realm.
+/// All local changes will be lost.
+/// {@category Sync}
+class DiscardLocalSyncClientResetHandler extends SyncClientResetErrorHandler {
```
nirinchev, line 401:
```suggestion
class DiscardUnsyncedChangesHandler extends SyncClientResetErrorHandler {
```

lib/src/configuration.dart
```diff
@@ -379,4 +383,69 @@ class SyncClientResetErrorHandler {
...
+  const DiscardLocalSyncClientResetHandler(super.callback);
+}
+
+///A client reset strategy that attempts to automatically recover any unsynchronized changes.
+///
+/// If you set [RecoverSyncClientResetHandler] callback as `syncClientResetErrorHandler` argument of [Configuration.flexibleSync],
+/// Realm will compare the local Realm with the Realm on the server and automatically transfer
+/// any changes from the local Realm that makes sense to the Realm provided by the server.
+/// {@category Sync}
+class RecoverSyncClientResetHandler extends SyncClientResetErrorHandler {
```
nirinchev, line 411:
```suggestion
class RecoverUnsyncedChangesHandler extends SyncClientResetErrorHandler {
```

lib/src/configuration.dart
```diff
@@ -379,4 +383,69 @@ class SyncClientResetErrorHandler {
...
+/// A client reset strategy that attempts to automatically recover any unsynchronized changes.
+/// If that fails, this handler fallsback to the discard unsynced changes strategy.
+///
+/// If you set [RecoverOrDiscardSyncClientResetHandler] callback as `syncClientResetErrorHandler` argument of [Configuration.flexibleSync],
+/// Realm will compare the local Realm with the Realm on the server and automatically transfer
+/// any changes from the local Realm that makes sense to the Realm provided by the server.
+/// If that fails, the local changes will be discarded.
+/// This is the default mode for fully synchronized Realms.
+/// {@category Sync}
+class RecoverOrDiscardSyncClientResetHandler extends SyncClientResetErrorHandler {
```
nirinchev, line 424:
```suggestion
class RecoverOrDiscardUnsyncedChangesHandler extends SyncClientResetErrorHandler {
```

nirinchev (changes requested):
Some names need to be updated to be aligned with the other SDKs. I tried to find the meetings notes from the meeting where the names were decided, but couldn't. In any case, I just copied the names from .NET as those should be the same as what's in other SDKs.

lib/src/configuration.dart
blagoev: I think having the `Error` there is more clear that it is an error. Actually this will be `ClientResetExceptionHandler` with the rename we are planning to do. Why do we put `Base` on each base class. We could skip that imo. That's not a common pattern.

lib/src/configuration.dart
blagoev: ManualClientResetExceptionHandler?

lib/src/configuration.dart
desistefanova: I decided to keep this name and to follow this naming convention, in order to avoid breaking changes.

lib/src/configuration.dart
blagoev: DiscardUnsyncedChangesClientResetErrorHandler? . will be  DiscardUnsyncedChangesClientResetExceptionHandler?

lib/src/configuration.dart
nirinchev: Can the last person close the door? 😜 Seriously though, these are the names that we aligned on across other SDKs and we should not deviate from them.

lib/src/configuration.dart
nirinchev: It's common for .NET and I've seen us do it in dart, but am fine with dropping the base. I don't think we need to have exception/error in the name, because it's not necessarily an exceptional situation. You handle a client reset request from the server, which can have multiple reasons to occur.

lib/src/configuration.dart
blagoev: :) ok if that was aligned then we don't need to discuss it again

lib/src/configuration.dart
blagoev: ok lets not have exception or error in the name

lib/src/configuration.dart
```diff
@@ -54,6 +54,11 @@ typedef InitialDataCallback = void Function(Realm realm);
 /// The `oldSchemaVersion` argument indicates the version from which the Realm migrates while
 typedef MigrationCallback = void Function(Migration migration, int oldSchemaVersion);
 
+/// The signature of a callback that will be triggered when a Client Reset error happens in a synchronized Realm.
+///
+/// The `clientResetError` argument holds useful data to be used when trying to manually recover from a client reset.
+typedef ClientResetCallback = void Function(SyncClientResetError clientResetError);
```
blagoev, line 60: Why do we call the type `SyncClientResetError` now. Should we rename to ClientResetError handler? @nirinchev ?

blagoev (changes requested):
Suggest we don't use named parameters here

lib/src/configuration.dart
```diff
@@ -394,14 +399,88 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+
+/// A client reset strategy where all the not yet synchronized data is automatically
+/// discarded and a fresh copy of the synchronized Realm is obtained.
+///
+/// If you set [DiscardUnsyncedChangesHandler] callback as `clientResetHandler` argument of [Configuration.flexibleSync],
+/// the local Realm will be discarded and replaced with the server side Realm.
+/// All local changes will be lost.
+/// {@category Sync}
+class DiscardUnsyncedChangesHandler extends ClientResetHandler {
+  const DiscardUnsyncedChangesHandler({required ClientResetCallback onReset}) : super(onReset);
```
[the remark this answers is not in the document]
nirinchev: I converted this PR to draft as there are more changes incoming. It was my suggestion to use named args because we'll have more callbacks there. The final signature will look something like `DiscardUnsyncedChangesHandler({BeforeCallback? beforeCallback, AfterCallback? afterCallback, ClientResetCallback? manualFallback})`. Having a function where you pass 3 optional callbacks without named args is going to look very confusing.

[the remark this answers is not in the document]
desistefanova: Shall I merge https://github.com/realm/realm-dart/pull/970 to this PR?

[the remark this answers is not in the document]
nirinchev: No, #970 is not ready at all yet. I'd like to get at least 1 test running to verify that things work as they should.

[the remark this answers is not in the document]
desistefanova: OK. This PR was only for setting the default value of resync mode. Maybe it is better to combine both tasks in one PR.

lib/src/configuration.dart
desistefanova: I will rename it

lib/src/configuration.dart
```diff
@@ -394,14 +399,88 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+  @override
+  ClientResyncModeInternal get _mode => ClientResyncModeInternal.manual;
+}
+
+/// A client reset strategy where all the not yet synchronized data is automatically
+/// discarded and a fresh copy of the synchronized Realm is obtained.
+///
+/// If you set [DiscardUnsyncedChangesHandler] callback as `clientResetHandler` argument of [Configuration.flexibleSync],
+/// the local Realm will be discarded and replaced with the server side Realm.
+/// All local changes will be lost.
```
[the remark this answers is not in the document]
desistefanova: This comment is outdated. Could you check the code now and clarify if they must be optional or not according to you?

lib/src/cli/atlas_apps/baas_client.dart
```diff
@@ -510,6 +530,24 @@ class BaasClient {
...
+    final app = BaasApp(appId, clientAppId, name);
+
+    final dynamic services = await _get('groups/$_groupId/apps/$appId/services');
+    dynamic service = services.firstWhere((dynamic s) => s["name"] == "BackingDB", orElse: () => throw Exception("Func 'confirmFunc' not found"));
+    final mongoServiceId = service['_id'] as String;
+    final dynamic configDocs = await _get('groups/$_groupId/apps/$appId/services/$mongoServiceId/config');
+    final dynamic flexibleSync = configDocs['flexible_sync'];
+    flexibleSync["is_recovery_mode_disabled"] = !enable;
+    configDocs['flexible_sync'] = flexibleSync;
+    String data = jsonEncode(<String, dynamic>{'clusterName': configDocs['clusterName'], 'flexible_sync': configDocs['flexible_sync']});
```
nirinchev, line 548: Why not json encode `configDocs`?

test/client_reset_test.dart
```diff
@@ -0,0 +1,256 @@
+////////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2021 Realm Inc.
```
nirinchev, line 3:
```suggestion
// Copyright 2022 Realm Inc.
```

test/client_reset_test.dart
```diff
@@ -0,0 +1,256 @@
...
+  for (Type clientResetHandlerType in [
+    RecoverOrDiscardUnsyncedChangesHandler,
+    RecoverUnsyncedChangesHandler,
+    DiscardUnsyncedChangesHandler,
+  ]) {
+    baasTest('$clientResetHandlerType.manualResetFallback invoked when throw an error on Before Callback', (appConfig) async {
+      final app = App(appConfig);
+      final user = await getIntegrationUser(app);
+
+      final onManualResetFallback = Completer<void>();
```
nirinchev, line 96: Can we change this to `Completer<ClientResetError>` and then assert on the error we get from the manual reset fallback?

test/client_reset_test.dart
```diff
@@ -0,0 +1,256 @@
...
+    final realm = await Realm.open(config);
+    await realm.syncSession.waitForUpload();
+
+    await triggerClientReset(realm);
+
+    final error = await resetCompleter.future;
+    expect(error.message, contains('Bad client file identifier'));
+  });
+
+  for (Type clientResetHandlerType in [
```
nirinchev, line 87: These for loops are repeated several times - can we move all the parameterized tests in a single for loop?

test/client_reset_test.dart
```diff
@@ -0,0 +1,256 @@
...
+    RecoverOrDiscardUnsyncedChangesHandler,
+    RecoverUnsyncedChangesHandler,
+    DiscardUnsyncedChangesHandler,
+  ]) {
+    baasTest('$clientResetHandlerType.manualResetFallback invoked when throw an error on After Callbacks', (appConfig) async {
+      final app = App(appConfig);
+      final user = await getIntegrationUser(app);
+
+      final onManualResetFallback = Completer<void>();
```
nirinchev, line 122: Similar to the previous test, let's assert on the exception we get in the manual reset fallback.

test/client_reset_test.dart
```diff
@@ -0,0 +1,256 @@
...
+      await onManualResetFallback.future;
+    });
+  }
+
+  for (Type clientResetHandlerType in [
+    RecoverOrDiscardUnsyncedChangesHandler,
+    RecoverUnsyncedChangesHandler,
+    DiscardUnsyncedChangesHandler,
+  ]) {
+    baasTest('$clientResetHandlerType.Befor and After callbacks are invoked', (appConfig) async {
```
nirinchev, line 149:
```suggestion
    baasTest('$clientResetHandlerType.Before and After callbacks are invoked', (appConfig) async {
```

test/client_reset_test.dart
```diff
@@ -0,0 +1,256 @@
...
+          afterDiscardCallback: (Realm beforeFrozen, Realm after) {
+            onAfterCompleter.complete();
+            discard = true;
+          },
+        ));
+
+    final realm = await Realm.open(config);
+    await realm.syncSession.waitForUpload();
+
+    await disableAutomaticRecoveryEnabled();
```
nirinchev, line 202: This name is a bit of a mouthful - can we shorten to `disableAutomaticRecovery`?

lib/src/cli/atlas_apps/baas_client.dart
desistefanova: It doesn't accept all the properties of the config to be set in the request. It insists only clusterName, flexible_sync, sync, permissions and one more thing to be passed.

nirinchev: I see, in that case, there should be no need to update it on line 547. We can just do 

```dart
final data = jsonEncode(<String, dynamic>{'clusterName': configDocs['clusterName'], 'flexible_sync': flexibleSync});
```

test/client_reset_test.dart
```diff
@@ -0,0 +1,345 @@
...
+    final subscription = realm.all<Task>().changes.listen((event) {
+      notifications.add(event);
+    });
+
+    await triggerClientReset(realm);
+    await onBeforeCompleter.future;
+    await onAfterCompleter.future;
+    expect(beforeResetCallbackOccured, 1);
+    expect(afterDiscardCallbackOccured, 1);
+    expect(manualResetFallbackOccured, 0);
```
nirinchev, line 294: There's nothing updating that variable, so it will always be 0.

test/client_reset_test.dart
```diff
@@ -0,0 +1,345 @@
...
+
+    await waitForCondition(() => notifications.length == 1, retryDelay: Duration(milliseconds: 10));
```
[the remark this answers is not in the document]
nirinchev: I believe Core will always emit an empty notification the first time a notification listener is attached. So we should probably rework this to something like:

```dart
final subscription = realm.all<Task>().changes.listen((event) {
  notifications.add(event);
});

await waitForCondition(() => notifications.length == 1);

await triggerClientReset(realm);
// await...

await waitForCondition(() => notifications.length == 2);
```

src/realm_dart_sync.cpp
```diff
@@ -147,20 +147,26 @@ RLM_API bool realm_dart_sync_before_reset_handler_callback(realm_userdata_t user
...
     auto ud = reinterpret_cast<realm_dart_userdata_async_t>(userdata);
-    ud->scheduler->invoke([ud, &realm, &cv, &m, &complete, &success]() {
-        std::unique_lock<std::mutex> lk(m);
-        success = (reinterpret_cast<realm_sync_before_client_reset_func_t>(ud->dart_callback))(ud->handle, realm);
+    std::unique_lock<std::mutex> lk(m);
+    auto unlockFunc = [&complete, &success, &lk, &cv](bool result) {
         complete = true;
+        success = result;
         lk.unlock();
         cv.notify_one();
```
nirinchev, line 155:
```suggestion
    auto unlockFunc = [&complete, &success, &m, &cv](bool result) {
        std::unique_lock<std::mutex> lk(m);
        complete = true;
        success = result;
        lk.unlock();
        cv.notify_one();
```

src/realm_dart_sync.cpp
```diff
@@ -147,20 +147,26 @@ RLM_API bool realm_dart_sync_before_reset_handler_callback(realm_userdata_t user
...
-    });
-
-    std::unique_lock<std::mutex> lk(m);
+    };
+    ud->scheduler->invoke([ud, &realm, &cv, &m, &complete, &success, &unlockFunc]() {
+        (reinterpret_cast<realm_sync_before_client_reset_lock_func_t>(ud->dart_callback))(ud->handle, realm, &unlockFunc);
+     });
     cv.wait(lk, [&complete] {return complete; });
```
nirinchev, line 160:
```suggestion
    std::unique_lock<std::mutex> lk(m);
    cv.wait(lk, [&complete] {return complete; });
```

src/realm_dart_sync.cpp
```diff
@@ -22,6 +22,32 @@
 #include "realm_dart.hpp"
 #include "realm_dart_sync.h"
 
+class WaitingCallback
```
nirinchev, line 25: Why do we need this class?

lib/src/native/realm_core.dart
```diff
@@ -476,15 +492,56 @@ class _RealmCore {
...
+          RealmInternal.getUnowned(syncConfig, RealmHandle._unowned(_realmLib.realm_from_thread_safe_reference(afterReference, scheduler.handle._pointer)));
+      _continueWhenComplete(() => afterResetCallback(beforeRealm, afterRealm), unlockCallbackFunc);
+    }
+    else {
+      _invokeNativeFunction(unlockCallbackFunc, true);
     }
+  }
 
-    syncConfig.syncErrorHandler(syncError);
+  static void _invokeNativeFunction(Pointer<Void> nativeFunction, bool result, [Object? error]) {
```
nirinchev, line 531: We can remove the `result` argument and infer it from `error != null`.

nirinchev (commented):
Will review tests a little later. I'm sure the C++ code can be reshuffled a little, but I can't concentrate right now to provide meaningful suggestions. The overall logic seems fine to me though.

src/realm_dart_sync.cpp
desistefanova: I wanted to create a method that is not visible in the external realm.h API. Could be better solution.

nirinchev: A freestanding function would not be visible in realm.h.

desistefanova: I wanted to say realm_dart_sync.h, not realm.h. Anyway, my mistake probably was that I had defined the function at the end of the file. If it is defined before to be called it builds and it is not needed to be exposed in realm_dart_sync.h. Thanks, I will remove the class.

common/lib/src/realm_types.dart
```diff
@@ -184,14 +184,15 @@ class SyncClientError extends SyncError {
 
 /// An error type that describes a client reset error condition.
 /// {@category Sync}
-class SyncClientResetError extends SyncError {
+class ClientResetError extends SyncError {
   /// If true the received error is fatal.
   final bool isFatal = true;
+  final Object? config;
```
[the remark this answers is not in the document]
nirinchev: Why not move `ClientResetError` to the realm/realm_dart package? As far as I know, the main reason for us to have the `realm_common` package is for the generator and the SDK to share code, but this error is not something the generator cares about.

src/realm_dart_sync.cpp
```diff
@@ -138,3 +137,57 @@ RLM_API void realm_dart_sync_on_subscription_state_changed_callback(realm_userda
...
+    realm::util::UniqueFunction<void(realm::util::UniqueFunction<void(bool)>*)> userCallback =
+        [&ud, &before_realm, &after_realm, did_recover](realm::util::UniqueFunction<void(bool)>* unlockFunc) {
+        ud->scheduler->invoke([ud, &before_realm, &after_realm, did_recover, unlockFunc]() {
+            (reinterpret_cast<realm_sync_after_client_reset_begin_func_t>(ud->dart_callback))(ud->handle, before_realm, after_realm, did_recover, unlockFunc);
+        });
+    };
+    return wait_dart_before_continue(&userCallback);
+}
+
+RLM_API void realm_dart_invoke_navite_with_result(bool success, void* unlockFunc)
```
nielsenko, line 187:
```suggestion
RLM_API void realm_dart_invoke_native_with_result(bool success, void* unlockFunc)
```

src/realm_dart_sync.h
```diff
@@ -35,3 +38,9 @@ RLM_API void realm_dart_sync_connection_state_changed_callback(realm_userdata_t
                                                                realm_sync_connection_state_e new_state);
 
 RLM_API void realm_dart_sync_on_subscription_state_changed_callback(realm_userdata_t userdata, realm_flx_sync_subscription_set_state_e state);
+
+RLM_API bool realm_dart_sync_before_reset_handler_callback(realm_userdata_t userdata, realm_t* realm);
+
+RLM_API bool realm_dart_sync_after_reset_handler_callback(realm_userdata_t userdata, realm_t* before_realm, realm_thread_safe_reference_t* after_realm, bool did_recover);
+
+RLM_API void realm_dart_invoke_navite_with_result(bool success, void* unlockFunc);
```
nielsenko, line 46:
```suggestion
RLM_API void realm_dart_invoke_native_with_result(bool success, void* unlockFunc);
```
Need to run ffigen again

src/realm_dart_sync.cpp
```diff
@@ -138,3 +137,57 @@ RLM_API void realm_dart_sync_on_subscription_state_changed_callback(realm_userda
...
+        std::unique_lock<std::mutex> lk(m);
+        complete = true;
+        success = result;
+        lk.unlock();
+        cv.notify_one();
+    };
+    (*userCallback)(&unlockFunc);
+
+    std::unique_lock<std::mutex> lk(m);
+    cv.wait(lk, [&complete] {return complete; });
```
nielsenko, line 158:
```suggestion
    realm::util::UniqueFunction unlockFunc = [&](bool result) {
        std::unique_lock lk(m);
        complete = true;
        success = result;
        lk.unlock();
        cv.notify_one();
    };
    (*userCallback)(&unlockFunc);

    std::unique_lock lk(m);
    cv.wait(lk, [&] { return complete; });
```
Let the compiler do the work. This is a "non-escaping" lambda - just use `[&]`

src/realm_dart_sync.cpp
```diff
@@ -138,3 +137,57 @@ RLM_API void realm_dart_sync_on_subscription_state_changed_callback(realm_userda
...
+
+RLM_API bool realm_dart_sync_before_reset_handler_callback(realm_userdata_t userdata, realm_t* realm)
+{
+    auto ud = reinterpret_cast<realm_dart_userdata_async_t>(userdata);
+    realm::util::UniqueFunction<void(realm::util::UniqueFunction<void(bool)>*)> userCallback =
+        [&ud, &realm](realm::util::UniqueFunction<void(bool)>* unlockFunc) {
+        ud->scheduler->invoke([ud, &realm, unlockFunc]() {
+            (reinterpret_cast<realm_sync_before_client_reset_begin_func_t>(ud->dart_callback))(ud->handle, realm, unlockFunc);
+        });
+    };
```
nielsenko, line 171:
```suggestion
    realm::util::UniqueFunction userCallback =
        [ud, realm](realm::util::UniqueFunction<void(bool)>* unlockFunc) {
        ud->scheduler->invoke([ud, realm, unlockFunc]() {
            (reinterpret_cast<realm_sync_before_client_reset_begin_func_t>(ud->dart_callback))(ud->handle, realm, unlockFunc);
        });
    };
```
Don't pass argument by reference to escaping lambdas.. and let the compiler to the type inference.

src/realm_dart_sync.cpp
```diff
@@ -138,3 +137,57 @@ RLM_API void realm_dart_sync_on_subscription_state_changed_callback(realm_userda
...
+
+RLM_API bool realm_dart_sync_after_reset_handler_callback(realm_userdata_t userdata, realm_t* before_realm, realm_thread_safe_reference_t* after_realm, bool did_recover)
+{
+    auto ud = reinterpret_cast<realm_dart_userdata_async_t>(userdata);
+    realm::util::UniqueFunction<void(realm::util::UniqueFunction<void(bool)>*)> userCallback =
+        [&ud, &before_realm, &after_realm, did_recover](realm::util::UniqueFunction<void(bool)>* unlockFunc) {
+        ud->scheduler->invoke([ud, &before_realm, &after_realm, did_recover, unlockFunc]() {
+            (reinterpret_cast<realm_sync_after_client_reset_begin_func_t>(ud->dart_callback))(ud->handle, before_realm, after_realm, did_recover, unlockFunc);
+        });
+    };
```
nielsenko, line 183:
```suggestion
   realm::util::UniqueFunction userCallback =
        [ud, before_realm, after_realm, did_recover](realm::util::UniqueFunction<void(bool)>* unlockFunc) {
        ud->scheduler->invoke([ud, before_realm, after_realm, did_recover, unlockFunc]() {
            (reinterpret_cast<realm_sync_after_client_reset_begin_func_t>(ud->dart_callback))(ud->handle, before_realm, after_realm, did_recover, unlockFunc);
        });
 ```
 Same

**Review by nielsenko — CHANGES_REQUESTED** [review]
I have some suggestions for the c++ code. I have gathered them in the branch `kn/resync_mode_suggestions` for your convinience.

File: src/realm_dart_sync.cpp [hunk 29]
```diff
@@ -138,3 +137,57 @@ RLM_API void realm_dart_sync_on_subscription_state_changed_callback(realm_userda
         (reinterpret_cast<realm_sync_on_subscription_state_changed_t>(ud->dart_callback))(ud->handle, state);
     });
 }
+
+bool wait_dart_before_continue(realm::util::UniqueFunction<void(realm::util::UniqueFunction<void(bool)>*)>* userCallback)
+{
+    std::condition_variable cv;
+    std::mutex m;
```
blagoev, line 144: lets use full variable names

blagoev, line 144:
```suggestion
    std::mutex mutex;
```

src/realm_dart_sync.cpp
```diff
@@ -138,3 +137,57 @@ RLM_API void realm_dart_sync_on_subscription_state_changed_callback(realm_userda
         (reinterpret_cast<realm_sync_on_subscription_state_changed_t>(ud->dart_callback))(ud->handle, state);
     });
 }
+
+bool wait_dart_before_continue(realm::util::UniqueFunction<void(realm::util::UniqueFunction<void(bool)>*)>* userCallback)
+{
+    std::condition_variable cv;
```
blagoev, line 143:
```suggestion
    std::condition_variable condition;
```

src/realm_dart_sync.cpp
```diff
@@ -138,3 +137,57 @@ RLM_API void realm_dart_sync_on_subscription_state_changed_callback(realm_userda
...
+    realm::util::UniqueFunction<void(bool)> unlockFunc = [&complete, &success, &cv, &m](bool result) {
+        std::unique_lock<std::mutex> lk(m);
+        complete = true;
+        success = result;
+        lk.unlock();
+        cv.notify_one();
+    };
+    (*userCallback)(&unlockFunc);
+
+    std::unique_lock<std::mutex> lk(m);
```
blagoev, line 157:
```suggestion
    std::unique_lock<std::mutex> lock(m);
```

blagoev (changes requested):
I have more comments, but decided that I have too many here. Will continue to review

src/realm_dart_sync.cpp
```diff
@@ -138,3 +137,57 @@ RLM_API void realm_dart_sync_on_subscription_state_changed_callback(realm_userda
...
+    cv.wait(lk, [&complete] {return complete; });
+
+    return success;
+}
+
+RLM_API bool realm_dart_sync_before_reset_handler_callback(realm_userdata_t userdata, realm_t* realm)
+{
+    auto ud = reinterpret_cast<realm_dart_userdata_async_t>(userdata);
+    realm::util::UniqueFunction<void(realm::util::UniqueFunction<void(bool)>*)> userCallback =
+        [&ud, &realm](realm::util::UniqueFunction<void(bool)>* unlockFunc) {
```
blagoev, line 167: realm is a pointer no need to capture a reference to it

test/test.dart
```diff
@@ -685,3 +687,40 @@ Future<void> _printPlatformInfo() async {
 
   print('Current PID $pid; OS $os, $pointerSize bit, CPU ${cpu ?? 'unknown'}');
 }
+
+Future<void> triggerClientReset(Realm realm, {bool restartSession = true}) async {
```
blagoev, line 691: do we need these methods in test.dart. Why don't we have it only in the client_reset_test.dart

blagoev (changes requested):

src/realm_dart_sync.cpp
nielsenko: For such a short function I actually prefer short variable names.

1. The function calls like `notify_one` and `unlock` stand out more.
2. It is easy to see the type anyway, as it is less than 10 lines away.

I also prefer `i` in a short `for` loop.

lib/src/configuration.dart
```diff
@@ -395,21 +418,161 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+abstract class ClientResetHandler {
+  /// Defines what should happen in case of a Client Resync
+  ClientResyncModeInternal get _mode;
+
+  BeforeResetCallback? get _beforeResetCallback => null;
+  AfterResetCallback? get _afterDiscardCallback => null;
+  AfterResetCallback? get _afterRecoveryCallback => null;
+
+  /// The callback that handles the [ClientResetError].
+  final ClientResetCallback? onManualReset;
```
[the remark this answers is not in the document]
nielsenko: Maybe go with the `onXYZ` naming everywhere? This would be consistent with how flutter names callback instances like `onPressed` etc?

lib/src/configuration.dart
```diff
@@ -395,21 +418,161 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+  BeforeResetCallback? get _beforeResetCallback => beforeResetCallback;
+
+  @override
+  AfterResetCallback? get _afterDiscardCallback => afterResetCallback;
+}
+
+/// A client reset strategy that attempts to automatically recover any unsynchronized changes.
+///
+/// If you set [RecoverUnsyncedChangesHandler] callback as `clientResetHandler` argument of [Configuration.flexibleSync],
+/// Realm will compare the local Realm with the Realm on the server and automatically transfer
```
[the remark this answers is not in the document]
nielsenko: In the above sentence that would make 3 hyperlinks to the same. I think that would look a bit silly.

lib/src/cli/atlas_apps/baas_client.dart
```diff
@@ -510,6 +530,23 @@ class BaasClient {
     final result = input.replaceRange(4, input.length - 4, '');
     return result;
   }
+
+  Future<void> setAutomaticRecoveryEnabled(String name, bool enable) async {
+    final dynamic docs = await _get('groups/$_groupId/apps');
```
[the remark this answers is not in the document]
nirinchev: Our lint rules enforce explicitly declaring dynamic variables.

lib/src/cli/atlas_apps/baas_client.dart
```diff
@@ -133,7 +148,10 @@ class BaasClient {
   }
 
   Future<void> _createAppIfNotExists(Map<String, BaasApp> existingApps, String appName, {String? confirmationType}) async {
-    if (!existingApps.containsKey(appName)) {
+    final existingApp = existingApps[appName];
+    if (existingApp != null) {
+      print('Found existing app ${existingApp.clientAppId}');
```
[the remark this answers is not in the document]
nirinchev: There's one statement in the else clause - early return here would make the code longer for no obvious clarity.

lib/src/configuration.dart
```diff
@@ -395,21 +418,161 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+/// All local changes will be lost.
+/// {@category Sync}
+class DiscardUnsyncedChangesHandler extends ClientResetHandler {
+  /// The callback that will be executed just before the client reset happens.
+  final BeforeResetCallback? beforeResetCallback;
+
+  /// The callback that will be executed just after the client reset happens.
+  final AfterResetCallback? afterResetCallback;
+
+  const DiscardUnsyncedChangesHandler({this.beforeResetCallback, this.afterResetCallback, ClientResetCallback? manualResetFallback})
```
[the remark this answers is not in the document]
nirinchev: I think `manualRecoveryFallbackCallback` is way too long. It's called `fallback` to indicate that this callback is executed when recovery fails. I don't think we should have `recovery` in the name either as we there's no reason to believe recovery would be possible in that situation.

lib/src/configuration.dart
```diff
@@ -395,21 +418,161 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+  /// The callback that will be executed just after the client reset happens if the local changes
+  /// needed to be discarded.
+  final AfterResetCallback? afterDiscardCallback;
+
+  const RecoverOrDiscardUnsyncedChangesHandler(
+      {this.beforeResetCallback, this.afterRecoveryCallback, this.afterDiscardCallback, ClientResetCallback? manualResetFallback})
+      : super(manualResetFallback);
+
+  @override
+  ClientResyncModeInternal get _mode => ClientResyncModeInternal.recoverOrDiscard;
```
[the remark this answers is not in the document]
nirinchev: What do you mean? The property in the base class is abstract.

lib/src/native/realm_core.dart
```diff
@@ -479,15 +495,54 @@ class _RealmCore {
 
   static void _syncErrorHandlerCallback(Object userdata, Pointer<realm_sync_session> session, realm_sync_error error) {
     final syncConfig = userdata as FlexibleSyncConfiguration;
+    final syncError = error.toSyncError(syncConfig);
 
-    final syncError = error.toSyncError();
+    if (syncError is ClientResetError) {
+      syncConfig.clientResetHandler.onManualReset?.call(syncError);
+    } else {
```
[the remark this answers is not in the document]
nirinchev: Similarly to other comments - using early returns for short if-else clauses makes the code less readable.

CHANGELOG.md
```diff
@@ -2,12 +2,41 @@
...
 * Support notifications on all managed realm lists, including list of primitives, ie. `RealmList<int>.changes` is supported. ([#893](https://github.com/realm/realm-dart/pull/893))
 * Support named backlinks on realm models. You can now add and annotate a realm object iterator field with `@Backlink(#fieldName)`. ([#996](https://github.com/realm/realm-dart/pull/996))
+* Support [Client Resets](https://www.mongodb.com/docs/atlas/app-services/sync/error-handling/client-resets/). Flutter SDK automatically detects the need for client resets and automatically performs a client reset according to the configured client reset callbacks for the type of client reset handlers set to `FlexibleSyncConfiguration`. Added argument `clientResetHandler` to `Configuration.flexibleSync`.
+  * `ManualRecoveryHandler` - the user needs to fully take care of a client reset.
+  * `DiscardUnsyncedChangesHandler` - all the not yet synchronized data is automatically discarded and a fresh copy of the synchronized Realm is obtained.
+  * `RecoverUnsyncedChangesHandler` - automatically recover any unsynchronized changes.
```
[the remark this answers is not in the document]
nirinchev: The changelog should not replace docs IMO.

CHANGELOG.md
```diff
@@ -18,7 +47,7 @@
 * Realm Studio: 12.0.0 or later.
 
 ### Internal
-* Using Core 12.11.0. ([#988](https://github.com/realm/realm-dart/pull/988))
+* Uses Realm Core 12.11.0. ([#988](https://github.com/realm/realm-dart/pull/988))
```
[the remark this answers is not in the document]
nirinchev: I think those changelog nits are a bit of a waste of time. We should decide on the wording and update the release action to use the correct one: https://github.com/realm/realm-dart/blob/866497cc366f59813462de2b7cdc05c0339cbfee/.github/workflows/publish-release.yml#L248.

CHANGELOG.md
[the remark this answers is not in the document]
blagoev: No, but right now it is like this is the best option, which is not the default. And for all other we say more what are the drawbacks.

CHANGELOG.md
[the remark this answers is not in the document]
blagoev: yes, indeed. I am not sure where is this coming from. I just noticed we are changing it, to something we have not used before. And checked .NET is using `Using` instead which we currently do. Hence the suggestion

lib/src/cli/atlas_apps/baas_client.dart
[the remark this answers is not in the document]
blagoev: ok, I see

lib/src/configuration.dart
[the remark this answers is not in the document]
blagoev: ok

lib/src/configuration.dart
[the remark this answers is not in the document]
blagoev: its a high light not a hyperlink.

lib/src/configuration.dart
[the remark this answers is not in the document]
blagoev: we have discussed and since Dart does not have visibility modifiers we decided to go with this approach

CHANGELOG.md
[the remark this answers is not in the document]
nirinchev: Yeah, not sure why this was changed here - let's revert and then have a separate PR that unifies the wording and updates the action to generate the correct one.

lib/src/configuration.dart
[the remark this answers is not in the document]
desistefanova: onBeforeReset, onAfterDiscard, onAfterRecoveryCallback and onManualResetCallback
or
beforeResetCallback, afterDiscardCallback, afterRecoveryCallback and manualResetCallback
?

CHANGELOG.md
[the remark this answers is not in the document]
desistefanova: I'm not sure I understand how to change it.

lib/src/configuration.dart
```diff
@@ -54,6 +55,28 @@ typedef InitialDataCallback = void Function(Realm realm);
...
+/// The `clientResetError` argument holds useful data to be used when trying to manually recover from a client reset.
+typedef ClientResetCallback = FutureOr<void> Function(ClientResetError clientResetError);
+
+/// Callback that indicates a Client Reset is about to happen.
+///
+/// The `beforeFrozen` argument holds the frozen Realm just before the client reset happens.
+///
+/// The lifetime of the Realm is tied to the callback, so don't store references to the Realm or objects
+/// obtained from it for use outside of the callback.
+typedef BeforeResetCallback = FutureOr<void> Function(Realm beforeFrozen);
```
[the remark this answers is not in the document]
desistefanova: Changed to `beforeResetRealm` and `afterResetRealm`.

lib/src/configuration.dart
```diff
@@ -395,21 +418,161 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+/// {@category Sync}
+class RecoverUnsyncedChangesHandler extends ClientResetHandler {
+  /// The callback that will be executed just before the client reset happens.
+  final BeforeResetCallback? beforeResetCallback;
+
+  /// The callback that will be executed just after the client reset happens.
+  final AfterResetCallback? afterResetCallback;
+
+  const RecoverUnsyncedChangesHandler({this.beforeResetCallback, this.afterResetCallback, ClientResetCallback? manualResetFallback})
+      : super(manualResetFallback);
```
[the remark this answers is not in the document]
desistefanova: onManualResetFallback to be consistant with .net SDK

lib/src/configuration.dart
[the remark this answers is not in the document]
nirinchev: If we use `onXYZ`, we don't need the `Callback` suffix IMO.

common/lib/src/realm_types.dart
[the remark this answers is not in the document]
desistefanova: I had to move all these classes:
        SyncError,
        SyncClientError,
        SyncConnectionError,
        SyncSessionError,
        GeneralSyncError,
        ClientResetError,
        SyncErrorHandler,
 Only enums had left to realm_types.dart

CHANGELOG.md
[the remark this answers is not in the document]
desistefanova: I removed most of the explanations, because I write them into API doc

lib/src/configuration.dart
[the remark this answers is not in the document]
desistefanova: I made them all as follow: `onBeforeReset`, `onAfterDiscard`, `onAfterRecoveryCallback` and `onManualResetCallback`

lib/src/configuration.dart
```diff
@@ -605,6 +605,10 @@ class ClientResetError extends SyncError {
     return "SyncError message: $message category: $category code: $code isFatal: $isFatal";
   }
 
+  /// Initiates the client reset process.
+  ///
+  /// On Windows, all Realm instances for that path must be disposed before this method is called or an
```
nirinchev, line 610:
```suggestion
  /// All Realm instances for that path must be closed before this method is called or an
```

We should not be explicit about Windows here - generally, you should not expect things to work normally if you have an open Realm instance while you're deleting the file.

blagoev (changes requested):

lib/src/native/realm_core.dart
```diff
@@ -219,13 +219,29 @@ class _RealmCore {
...
           _realmLib.realm_sync_config_set_session_stop_policy(syncConfigPtr, config.sessionStopPolicy.index);
-
+          _realmLib.realm_sync_config_set_resync_mode(syncConfigPtr, config.clientResetHandler.clientResyncMode.index);
           final errorHandlerCallback =
               Pointer.fromFunction<Void Function(Handle, Pointer<realm_sync_session_t>, realm_sync_error_t)>(_syncErrorHandlerCallback);
           final errorHandlerUserdata = _realmLib.realm_dart_userdata_async_new(config, errorHandlerCallback.cast(), scheduler.handle._pointer);
           _realmLib.realm_sync_config_set_error_handler(syncConfigPtr, _realmLib.addresses.realm_dart_sync_error_handler_callback, errorHandlerUserdata.cast(),
               _realmLib.addresses.realm_dart_userdata_async_free);
 
+          if (config.clientResetHandler is! ManualRecoveryHandler) {
```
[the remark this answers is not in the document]
nirinchev: Yeah, I think this may have been written before establishing the type hierarchy. I think this should change to something like:

```dart
if (config.clientResetHandler.onBeforeReset != null) {
  // hook up the before reset callback in Core
}

if (config.clientResetHandler.onAfterDiscard != null || ...onAfterRecovery != null) {
  // hook up after reset callback
}
```

lib/src/native/realm_core.dart
```diff
@@ -480,15 +496,56 @@ class _RealmCore {
...
+    final afterRealm = RealmInternal.getUnowned(
+      syncConfig,
+      RealmHandle._unowned(_realmLib.realm_from_thread_safe_reference(afterReference, scheduler.handle._pointer)),
+    );
+    _continueWhenComplete(() => afterResetCallback(beforeRealm, afterRealm), unlockCallbackFunc);
+  }
+
+  static void _invokeNativeFunction(Pointer<Void> nativeFunction, [Object? error]) {
+    if (error != null) {
+      _realmLib.realm_register_user_code_callback_error(error.toPersistentHandle());
```
blagoev, line 537: when are we calling realm_get_last_error to retrieve this error. Aren't we leaking this handle or aren't we setting last error incorrectly for the next Core call to fail

lib/src/configuration.dart
```diff
@@ -101,20 +124,20 @@ abstract class Configuration implements Finalizable {
 
   /// Specifies the FIFO special files fallback location.
   ///
-  /// Opening a [Realm] creates a number of FIFO special files in order to
-  /// coordinate access to the [Realm] across threads and processes. If the [Realm] file is stored in a location
-  /// that does not allow the creation of FIFO special files (e.g. the FAT32 filesystem), then the [Realm] cannot be opened.
-  /// In that case [Realm] needs a different location to store these files and this property defines that location.
-  /// The FIFO special files are very lightweight and the main [Realm] file will still be stored in the location defined
+  /// Opening a `Realm` creates a number of FIFO special files in order to
```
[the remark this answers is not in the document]
desistefanova: How do you think it is correct to be? Why do we need to to have all these refferences to Realm? I tried to make the API doc consistent.

lib/src/native/realm_core.dart
```diff
@@ -480,15 +496,56 @@ class _RealmCore {
...
+    _continueWhenComplete(() => beforeResetCallback(realm), unlockCallbackFunc);
+  }
 
-    syncConfig.syncErrorHandler(syncError);
+  static void _syncAfterResetCallback(Object userdata, Pointer<shared_realm> beforeHandle, Pointer<realm_thread_safe_reference> afterReference, bool didRecover,
+      Pointer<Void> unlockCallbackFunc) {
+    final syncConfig = userdata as FlexibleSyncConfiguration;
+    final afterResetCallback = didRecover ? syncConfig.clientResetHandler.onAfterRecovery : syncConfig.clientResetHandler.onAfterDiscard;
+    if (afterResetCallback == null) {
+      _invokeNativeFunction(unlockCallbackFunc);
```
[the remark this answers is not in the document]
desistefanova: Discussion is moved to https://github.com/realm/realm-dart/pull/1003

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
desistefanova: Moved to https://github.com/realm/realm-dart/pull/1003

lib/src/native/realm_core.dart
desistefanova: This error is not provided when we are getting the last error. It is only logged that 'User-provided callback failed'.

lib/src/configuration.dart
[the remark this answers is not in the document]
desistefanova: Fixed. Please review them.

test/client_reset_test.dart
```diff
@@ -0,0 +1,609 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////////
+
+import 'dart:async';
+import 'dart:io';
+
+import 'package:test/test.dart' hide test, throws;
+import '../lib/realm.dart';
+import '../lib/src/configuration.dart' show ClientResetHandlerInternal, ClientResyncModeInternal, BeforeResetCallback, AfterResetCallback, ClientResetCallback;
```
blagoev, line 24: Are these all needed. Shouldn't `BeforeResetCallback` and the rest be already avaialble

lib/src/configuration.dart
```diff
@@ -395,21 +418,315 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+    return "SyncError message: $message category: $category code: $code isFatal: $isFatal";
+  }
+
+  /// Initiates the client reset process.
+  ///
+  /// All Realm instances for that path must be closed before this method is called or an
+  /// Exception will be thrown.
+  void resetRealm() {
+    if (config == null || config is! FlexibleSyncConfiguration) {
+      throw Exception("FlexibleSyncConfiguration is not set into ClientResetError");
```
blagoev, line 588: All exception that the SDK throws should be RealmExceptions or inheritors. This is also true for errors where it should be RealmError

lib/src/configuration.dart
```diff
@@ -395,21 +418,315 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+
+  @override
+  String toString() {
+    return "SyncError message: $message category: $category code: $code isFatal: $isFatal";
+  }
+
+  /// Initiates the client reset process.
+  ///
+  /// All Realm instances for that path must be closed before this method is called or an
+  /// Exception will be thrown.
```
blagoev, line 585:
```suggestion
  /// RealmException will be thrown.
```

lib/src/configuration.dart
```diff
@@ -293,15 +316,15 @@ class FlexibleSyncConfiguration extends Configuration {
 
   SessionStopPolicy _sessionStopPolicy = SessionStopPolicy.afterChangesUploaded;
 
-  /// Called when a [SyncError] occurs for this synchronized [Realm].
+  /// Called when a [SyncError] occurs for this synchronized realm.
```
blagoev, line 319:
```suggestion
  /// Called when a [SyncError] occurs for this synchronized `Realm`.
```

lib/src/configuration.dart
```diff
@@ -293,15 +316,15 @@ class FlexibleSyncConfiguration extends Configuration {
...
   ///
-  /// The default [SyncErrorHandler] prints to the console
+  /// The default [SyncErrorHandler] prints to the console.
   final SyncErrorHandler syncErrorHandler;
 
-  /// Called when a [SyncClientResetError] occurs for this synchronized [Realm]
+  /// Called when a [ClientResetError] occurs for this synchronized realm.
```
blagoev, line 324:
```suggestion
  /// Called when a [ClientResetError] occurs for this synchronized `Realm`.
```

lib/src/configuration.dart
```diff
@@ -293,15 +316,15 @@ class FlexibleSyncConfiguration extends Configuration {
...
   ///
-  /// The default [SyncClientResetErrorHandler] logs a message using the current Realm.logger
-  final SyncClientResetErrorHandler syncClientResetErrorHandler;
+  /// The default [ClientResetHandler] logs a message using the current `Realm.logger`.
```
blagoev, line 326:
```suggestion
  /// The default [ClientResetHandler] logs a message using the current [Realm.logger].
```

lib/src/configuration.dart
```diff
@@ -395,21 +418,315 @@ class RealmSchema extends Iterable<SchemaObject> {
...
 
-  /// Initializes a new instance of of [SyncClientResetErrorHandler].
-  const SyncClientResetErrorHandler(this.callback);
+/// [ClientResetHandler] is triggered if the device and server cannot agree
+/// on a common shared history for the realm file
+/// or when it is impossible for the device to upload or receive any changes.
+/// This can happen if the server is rolled back or restored from backup.
+/// {@category Sync}
+abstract class ClientResetHandler {
+  // Defines what should happen in case of a Client reset
```
blagoev, line 432:
```suggestion
  // Defines what should happen in case of a client reset
```

lib/src/configuration.dart
```diff
@@ -395,21 +418,315 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+  final ClientResyncModeInternal _mode;
+
+  final BeforeResetCallback? _onBeforeReset;
+  final AfterResetCallback? _onAfterDiscard;
+  final AfterResetCallback? _onAfterRecovery;
+
+  /// The callback that handles the [ClientResetError].
+  final ClientResetCallback? onManualReset;
+
+  /// Initializes a new instance of [ClientResetHandler].
```
blagoev, line 442:
```suggestion
```

lib/src/configuration.dart
```diff
@@ -395,21 +418,315 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+  final Configuration? config;
+
+  /// The [ClientResetError] has error code of [SyncClientErrorCode.autoClientResetFailure]
+  SyncClientErrorCode get code => SyncClientErrorCode.autoClientResetFailure;
+
+  ClientResetError(String message, [this.config]) : super(message, SyncErrorCategory.client, SyncClientErrorCode.autoClientResetFailure.code);
+
+  @override
+  String toString() {
+    return "SyncError message: $message category: $category code: $code isFatal: $isFatal";
```
blagoev, line 579:
```suggestion
    return "ClientResetError message: $message category: $category code: $code isFatal: $isFatal";
```

lib/src/configuration.dart
blagoev: also the exception text is not clear. What are we trying to say here?

Probably `The current configuration is not FlexibleSyncConfiguration`

lib/src/configuration.dart
```diff
@@ -395,21 +418,315 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+  SyncClientError(
+    String message,
+    SyncErrorCategory category,
+    SyncClientErrorCode errorCode, {
+    this.isFatal = false,
+  }) : super(message, category, errorCode.code);
+
+  @override
+  String toString() {
+    return "SyncError message: $message category: $category code: $code isFatal: $isFatal";
```
blagoev, line 653:
```suggestion
    return "SyncClientError message: $message category: $category code: $code isFatal: $isFatal";
```

lib/src/configuration.dart
```diff
@@ -395,21 +418,315 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+  SyncConnectionError(
+    String message,
+    SyncErrorCategory category,
+    SyncConnectionErrorCode errorCode, {
+    this.isFatal = false,
+  }) : super(message, category, errorCode.code);
+
+  @override
+  String toString() {
+    return "SyncError message: $message category: $category code: $code isFatal: $isFatal";
```
blagoev, line 675:
```suggestion
    return "SyncConnectionError message: $message category: $category code: $code isFatal: $isFatal";
```

lib/src/configuration.dart
```diff
@@ -395,21 +418,315 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+  SyncSessionError(
+    String message,
+    SyncErrorCategory category,
+    SyncSessionErrorCode errorCode, {
+    this.isFatal = false,
+  }) : super(message, category, errorCode.code);
+
+  @override
+  String toString() {
+    return "SyncError message: $message category: $category code: $code isFatal: $isFatal";
```
blagoev, line 697:
```suggestion
    return "SyncSessionError message: $message category: $category code: $code isFatal: $isFatal";
```

lib/src/configuration.dart
```diff
@@ -395,21 +418,315 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+/// A general or unknown sync error
+class GeneralSyncError extends SyncError {
+  /// The numeric value indicating the type of the general sync error.
+  int get code => codeValue;
+
+  GeneralSyncError(String message, SyncErrorCategory category, int code) : super(message, category, code);
+
+  @override
+  String toString() {
+    return "SyncError message: $message category: $category code: $code";
```
blagoev, line 710:
```suggestion
    return "GeneralSyncError message: $message category: $category code: $code";
```

lib/src/native/realm_core.dart
```diff
@@ -2179,6 +2245,13 @@ class _RealmCore {
       return completer.future;
     });
   }
+
+  void immediatelyRunFileActions(App app, String realmPath) {
+    using((arena) {
+      _realmLib.invokeGetBool(() => _realmLib.realm_sync_immediately_run_file_actions(app.handle._pointer, realmPath.toCharPtr(arena)),
+          "An error occurred while deleting Realm file. Check if the file is in use: '$realmPath'");
```
blagoev, line 2252:
```suggestion
          "An error occurred while deleting the Realm file. Check if the file is in use: '$realmPath'");
```
should we replace `deleting the` with `working with the`

blagoev: the `immediatelyRunFileActions` does not imply deletions but that's probably what is going on. still we could be vague about it.

lib/src/configuration.dart
```diff
@@ -395,21 +418,315 @@ class RealmSchema extends Iterable<SchemaObject> {
...
+  recover,
+  recoverOrDiscard,
+}
+
+/// An error type that describes a client reset error condition.
+/// {@category Sync}
+class ClientResetError extends SyncError {
+  /// If true the received error is fatal.
+  final bool isFatal = true;
+  final Configuration? config;
```
blagoev, line 570: no api doc

blagoev, line 570: why do we have the config here or why is it public?

blagoev (changes requested):

lib/src/configuration.dart
desistefanova: We use it for `resetRealm` method. I made it private.

lib/src/native/realm_core.dart
desistefanova: "An error occurred while resetting the Realm. Check if the file is in use:

lib/src/configuration.dart
```diff
@@ -567,27 +567,28 @@ enum ClientResyncModeInternal {
...
 
   /// Initiates the client reset process.
   ///
   /// All Realm instances for that path must be closed before this method is called or an
-  /// Exception will be thrown.
+  /// [RealmException] will be thrown.
   void resetRealm() {
-    if (config == null || config is! FlexibleSyncConfiguration) {
-      throw Exception("FlexibleSyncConfiguration is not set into ClientResetError");
+    if (_config == null || _config is! FlexibleSyncConfiguration) {
```
blagoev, line 588: if (_config is! FlexibleSyncConfiguration) {

CHANGELOG.md
```diff
@@ -10,6 +14,30 @@
 * Support named backlinks on realm models. You can now add and annotate a realm object iterator field with `@Backlink(#fieldName)`. ([#996](https://github.com/realm/realm-dart/pull/996))
 * Allow `@Indexed` attribute on all indexable type, and ensure appropriate indexes are created in the realm. ([#797](https://github.com/realm/realm-dart/issues/797))
 * Add `parent` getter on embedded objects. ([#979](https://github.com/realm/realm-dart/pull/979))
+* Support [Client Resets](https://www.mongodb.com/docs/atlas/app-services/sync/error-handling/client-resets/). Flutter SDK automatically detects the need for client resets and automatically performs it according to the configured callbacks for the type of client reset handlers set to `FlexibleSyncConfiguration`. 
+  Added a parameter `clientResetHandler` to `Configuration.flexibleSync`.
```
nielsenko, line 18:
```suggestion
* Added a parameter `clientResetHandler` to `Configuration.flexibleSync`. 
```

CHANGELOG.md
```diff
@@ -10,6 +14,30 @@
...
+          },
+          onAfterDiscard: (beforeResetRealm, afterResetRealm) {
+            // Executed after an automatic recovery from a client reset has failed but the Discard has completed.
+          },
+          onManualResetFallback: (clientResetError) {
+            // Handle the reset manually in case some of the callbacks above throws an exception
+          },
+        )
+    );
+```
```
nielsenko, line 40: I miss a ref to an issue/pr here

lib/src/realm_class.dart
```diff
@@ -77,8 +70,22 @@ export "configuration.dart"
...
         SyncErrorHandler,
-        SyncClientResetErrorHandler;
+        ClientResetHandler,
+        ManualRecoveryHandler,
+        DiscardUnsyncedChangesHandler,
+        RecoverUnsyncedChangesHandler,
+        RecoverOrDiscardUnsyncedChangesHandler,
+        BeforeResetCallback,
+        AfterResetCallback,
+        ClientResetCallback;
```
nielsenko, line 88: Should we perhaps just sort all of these alphabetically?

nielsenko (approved):
LGTM. Have a few suggestions. Mainly to the CHANGELOG.md

blagoev (approved):
one small suggestion

CHANGELOG.md
```diff
@@ -10,6 +14,30 @@
 * Support named backlinks on realm models. You can now add and annotate a realm object iterator field with `@Backlink(#fieldName)`. ([#996](https://github.com/realm/realm-dart/pull/996))
 * Allow `@Indexed` attribute on all indexable type, and ensure appropriate indexes are created in the realm. ([#797](https://github.com/realm/realm-dart/issues/797))
 * Add `parent` getter on embedded objects. ([#979](https://github.com/realm/realm-dart/pull/979))
+* Support [Client Resets](https://www.mongodb.com/docs/atlas/app-services/sync/error-handling/client-resets/). Flutter SDK automatically detects the need for client resets and automatically performs it according to the configured callbacks for the type of client reset handlers set to `FlexibleSyncConfiguration`.
```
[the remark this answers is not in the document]
desistefanova: This text is based on https://www.mongodb.com/docs/atlas/app-services/sync/error-handling/client-resets/#handle-a-client-reset, where it is written that the SDKs detect the client reset.

CHANGELOG.md
desistefanova: This isn't a new functionality. The text just continue explaining what is changed in relation to Client reset.

lib/src/realm_class.dart
desistefanova: I don't think they are grouped by types: erros, callbacks, handlers...

CHANGELOG.md
blagoev: yeah that is part of the same `Support Client Reset` bullet above.

CHANGELOG.md
blagoev: its the `Support Client Reset` bullet above

lib/src/realm_class.dart
blagoev: if we decide to do that it will be easy since vs code has that command

desistefanova: I like the classes for similar things to be side by side. Is it a common practice to be alphabetically?

CHANGELOG.md
desistefanova: Fixed.

lib/src/realm_class.dart
desistefanova: Done

PR #930, opened 2022-10-03
Add support for writeAsync

nirinchev:


lib/src/native/realm_core.dart
```diff
@@ -2645,6 +2712,64 @@ extension LevelExt on Level {
...
+  final Realm _realm;
+
+  WriteCompleter(this._realm, this._cancellationToken) : _internalCompleter = Completer<void>() {
+    if (_cancellationToken != null) {
+      if (_cancellationToken!.isCancelled) {
+        _internalCompleter.completeError(CancelledException());
+      } else {
+        _cancellationToken?.attach(this);
+      }
+    }
```
nielsenko, line 2728:
```suggestion
    final ct = _cancellationToken; // <-- allow compiler to correctly infer nullability after null check
    if (ct != null) {
      if (ct.isCancelled) {
        _internalCompleter.completeError(CancelledException());
      } else {
        ct.attach(this);
      }
    }
```

lib/src/realm_class.dart
```diff
@@ -60,6 +61,8 @@ export 'package:realm_common/realm_common.dart'
         ObjectId,
         Uuid;
 
+export 'package:cancellation_token/cancellation_token.dart' show CancellationToken, CancelledException;
```
nielsenko, line 64: Why this export?

lib/src/realm_class.dart
```diff
@@ -376,32 +415,62 @@ class Realm implements Finalizable {
...
+
+  /// Commits the changes to the Realm asynchronously.
+  /// Canceling the commit using the [cancellationToken] will not abort the transaction, but
+  /// rather resolve the future immediately with a [CancelledException].
+  Future<void> commitAsync({CancellationToken? cancellationToken}) async {
+    _ensureOpen('commit');
+
+    await realmCore.commitWriteAsync(_realm!, cancellationToken);
+
+    _closeTransaction();
```
nielsenko, line 453: Should we ensure this is called even if the above throws?

lib/src/realm_class.dart
```diff
@@ -376,32 +415,62 @@ class Realm implements Finalizable {
...
   }
 
+  /// Undoes all changes made in the transaction.
   void rollback() {
-    if (_realm == null) {
-      throw RealmException('Transaction was already closed. Cannot rollback');
+    _ensureOpen('rollback');
+
+    if (!_realm!.isClosed) {
+      realmCore.rollbackWrite(_realm!);
```
nielsenko, line 461: Can this throw?

test/writeasync_test.dart
```diff
@@ -0,0 +1,244 @@
...
+    expect(acquisitionOrder, [0, 1, 2, 3, 4]);
+  });
+
+  test('Realm.beginWriteAsync with cancellation token', () async {
+    final realm1 = getRealm(Configuration.local([Person.schema]));
+    final realm2 = getRealm(Configuration.local([Person.schema]));
+    final t1 = realm1.beginWrite();
+
+    final token = CancellationToken();
+    Future<void>.delayed(Duration(milliseconds: 1)).then((value) => token.cancel());
```
nielsenko, line 164:
```suggestion
    final token = TimeoutCancellationToken(const Duration(milliseconds: 1), timeoutException: CancelledException());
```
Slightly shorter.. In case you didn't know about this?

test/writeasync_test.dart
```diff
@@ -0,0 +1,244 @@
...
+
+  test('Realm.writeAsync with cancellation token', () async {
+    final realm1 = getRealm(Configuration.local([Person.schema]));
+    final realm2 = getRealm(Configuration.local([Person.schema]));
+    final t1 = realm1.beginWrite();
+
+    final token = CancellationToken();
+    Future<void>.delayed(Duration(milliseconds: 1)).then((value) => token.cancel());
+
+    await expectLater(() async => await realm2.writeAsync(() {}, cancellationToken: token), throwsA(isA<CancelledException>()));
```
nielsenko, line 179:
```suggestion
    await expectLater(realm2.writeAsync(() {}, cancellationToken: token), throwsA(isA<CancelledException>()));
```

test/writeasync_test.dart
```diff
@@ -0,0 +1,244 @@
...
+    final realm = getRealm(Configuration.local([Person.schema]));
+
+    final token = CancellationToken();
+    await expectLater(
+        () => realm.writeAsync(() {
+              realm.add(Person('A'));
+              token.cancel();
+              realm.add(Person('B'));
+            }, cancellationToken: token),
+        throwsA(isA<CancelledException>()));
```
nielsenko, line 207:
```suggestion
    await expectLater(
        realm.writeAsync(() {
          realm.add(Person('A'));
          token.cancel();
          realm.add(Person('B'));
        }, cancellationToken: token),
        throwsA(isA<CancelledException>()));
```

test/writeasync_test.dart
```diff
@@ -0,0 +1,244 @@
...
+    expect(realm.isInTransaction, false);
+  });
+
+  test('Realm.writeAsync with a canceled token throws', () async {
+    final realm = getRealm(Configuration.local([Person.schema]));
+
+    final token = CancellationToken();
+    token.cancel();
+
+    await expectLater(() async => await realm.writeAsync(() {}, cancellationToken: token), throwsA(isA<CancelledException>()));
```
nielsenko, line 219:
```suggestion
    await expectLater(realm.writeAsync(() {}, cancellationToken: token), throwsA(isA<CancelledException>()));
```

test/writeasync_test.dart
```diff
@@ -0,0 +1,244 @@
...
+    expect(realm.isInTransaction, false);
+  });
+
+  test('Realm.beginWriteAsync with a canceled token throws', () async {
+    final realm = getRealm(Configuration.local([Person.schema]));
+
+    final token = CancellationToken();
+    token.cancel();
+
+    await expectLater(() async => await realm.beginWriteAsync(cancellationToken: token), throwsA(isA<CancelledException>()));
```
nielsenko, line 229:
```suggestion
    await expectLater(realm.beginWriteAsync(cancellationToken: token), throwsA(isA<CancelledException>()));
```

test/writeasync_test.dart
```diff
@@ -0,0 +1,244 @@
...
+
+  test('Transaction.commitAsync with a canceled token throws', () async {
+    final realm = getRealm(Configuration.local([Person.schema]));
+
+    final transaction = await realm.beginWriteAsync();
+
+    final token = CancellationToken();
+    token.cancel();
+
+    await expectLater(() async => await transaction.commitAsync(cancellationToken: token), throwsA(isA<CancelledException>()));
```
nielsenko, line 241:
```suggestion
    await expectLater(transaction.commitAsync(cancellationToken: token), throwsA(isA<CancelledException>()));
```

nielsenko (changes requested):

nielsenko:
Have combined my suggestions in #961

lib/src/realm_class.dart
nirinchev: So that people don't need to import an additional package to use our API. I believe it's similar to what we do for the datatypes.

nielsenko: You are thinking about `Uuid` and `ObjectId` from `realm_common`?

nirinchev: Yes

test/writeasync_test.dart
nirinchev: No, but this one does:

```dart
  test('example', () async {
    Future<void> test() {
      throw Exception('before first await!');
    }

    await expectLater(test(), throwsA(isA<Exception>()));
  });
```

nielsenko: Yes, but that is just [bad style](https://dart.dev/guides/language/effective-dart/usage#prefer-asyncawait-over-using-raw-futures), and that is not how `writeAsync` is written?

nielsenko: Anyway, I'm fine with both 👍

lib/src/realm_class.dart
nielsenko: I'm not a fan of those exports either 😃 .. 

Anyway, I have made my view-point known, and we will let the community judge 😄

nielsenko:
I noticed you rolled back the guards to call `_closeTransaction`. Can you elaborate a little?

test/writeasync_test.dart
nirinchev: > How would you return a future from test

Not sure I understand - in the example above, we're refactoring the implementation of `writeAsync`, nothing to do with tests. Yes, it's returning a raw future, but it's not using `.then/.catch` which is "bad style" the way I read the guide. In any case, I still prefer the lambda style, because it's consistent with non-promise asserts:

```dart
expect(() => doSomething(), throws<Exception>())
await expectLater(() => doSomethingAsync(), throws<Exception>());
```

nielsenko: As I read the guide you should use `async` unless dropping it doesn't change behaviour.

You can write:
```dart
Future<T> writeAsync(...) {
  return _writeAsyncInternal(...);
}
```
but replacing:
```dart
Future<T> writeAsync(...) async {
  if (something) {
    throw Exception();
  }

  return awaIt _writeAsyncInternal(...);
}
```
with
```dart
Future<T> writeAsync(...) {
  if (something) {
    throw Exception();
  }

  return _writeAsyncInternal(...);
}
```
does change behaviour and should be avoided.

Anyway, as I said, I can live with the defensive code in the tests, as long as we don't actually ever push such behaviour onto our users.

nirinchev:
Yes, I thought about it and I don't think it's the right thing to do. Imagine you have the following code:

```dart
final t = realm.beginWrite();
// Do stuff

try {
  t.commit();
} catch (e) {
  // Commit failed - what is the state of the transaction here?
}
```

My opinion is that in the catch, the transaction should still be open. If the operation that terminates the transaction fails, wouldn't you expect the transaction to still be open and for you to be able to retry?

nielsenko:
Do you mean retry the `commit()`? If it failed before the `commit()`, then yes I expect the transaction to still be open, but if either `commit()` or `rollback()` fails? In what cases would it make sense for the transaction to still be open?

nielsenko (approved):
LGTM

test/writeasync_test.dart
```diff
@@ -0,0 +1,244 @@
...
+
+import 'dart:io';
+
+import 'package:cancellation_token/cancellation_token.dart';
+import 'package:test/expect.dart';
+
+import '../lib/realm.dart';
+import 'test.dart';
+
+Future<void> main([List<String>? args]) async {
```
blagoev, line 27: We generally group tests based on the entity. Let's include these in `realm_test.dart` instead

blagoev (changes requested):

lib/src/native/realm_core.dart
```diff
@@ -2645,6 +2712,65 @@ extension LevelExt on Level {
...
 }
 
+class WriteCompleter with Cancellable implements Completer<void> {
+  final Completer<void> _internalCompleter;
+  final CancellationToken? _cancellationToken;
+  int? _id;
+  final Realm _realm;
+
+  WriteCompleter(this._realm, this._cancellationToken) : _internalCompleter = Completer<void>() {
+    final ct = _cancellationToken;
```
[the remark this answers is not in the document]
nirinchev: To allow dart null analysis to kick in after the `if (ct != null)` check. Otherwise, we'll have to always add `!` when calling API on `_cancellationToken`.

lib/src/native/realm_core.dart
```diff
@@ -2645,6 +2712,65 @@ extension LevelExt on Level {
...
+  @override
+  void completeError(Object error, [StackTrace? stackTrace]) {
+    if (isCancelled) return;
+    _cancellationToken?.detach(this);
+    _internalCompleter.completeError(error, stackTrace);
+  }
+
+  @override
+  void onCancel(Exception cancelException, [StackTrace? stackTrace]) {
+    if (_id == null || realmCore._cancelAsync(_realm, _id!)) {
```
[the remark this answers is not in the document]
nirinchev: No - if `cancelAsync` returns false, then we should not cancel the future because the cancellation did not succeed. This is typically the case when there's a race between the user calling cancel and Core actually acquiring the write lock. If we complete with an error here, we'd end up in a situation where the dart `beginWrite` is canceled, but the transaction is still open.

lib/src/native/realm_core.dart
```diff
@@ -582,6 +583,72 @@ class _RealmCore {
...
+  Future<void> beginWriteAsync(Realm realm, CancellationToken? ct) {
+    final completer = WriteCompleter(realm, ct);
+    if (completer.isCancelled) {
+      return completer.future;
+    }
+
+    final cancelationId = _realmLib.realm_async_begin_write(realm.handle._pointer, Pointer.fromFunction(_completeAsyncBeginWrite),
+        completer.toPersistentHandle(), _realmLib.addresses.realm_dart_delete_persistent_handle, true);
+
+    if (completer.isCancelled) {
```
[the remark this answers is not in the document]
nirinchev: Still have my C# hat on 😅

lib/src/realm_class.dart
nirinchev: No, we discussed it in the PR comments - https://github.com/realm/realm-dart/pull/930#issuecomment-1276324855 and the next couple of comments. Ultimately, we don't want to close the transaction if committing fails because that would be the wrong behavior. The user should still be able to retry the commit, even if it's unlikely that it'll succeed.

test/writeasync_test.dart
nirinchev: I changed the code to assert on the future directly rather than a lambda.

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
blagoev: got it

lib/src/realm_class.dart
blagoev: ok, I noticed the discussion but this comment was still open with a thumbs up. Hence asking.

lib/src/realm_class.dart
```diff
@@ -376,32 +415,65 @@ class Realm implements Finalizable {
   /// Disclaimer: This method is mostly needed on Dart standalone and if not called the Dart program will hang and not exit.
   /// This is a workaround of a Dart VM bug and will be removed in a future version of the SDK.
   static void shutdown() => scheduler.stop();
+
+  void _throwIfFrozen() {
```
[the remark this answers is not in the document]
blagoev: `_isWritable` is even better

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
nirinchev: That is fine though - the completer will resolve normally as if `cancel` was never called. If `_cancelAsync` returns `false`, it means Core didn't register your cancellation attempt and will eventually invoke the callback for the async operation.

test/realm_test.dart
```diff
@@ -829,6 +854,219 @@ Future<void> main([List<String>? args]) async {
...
+    t1.rollback();
+  });
+
+  test('Realm.writeAsync with cancellation token', () async {
+    final realm1 = getRealm(Configuration.local([Person.schema]));
+    final realm2 = getRealm(Configuration.local([Person.schema]));
+    final t1 = realm1.beginWrite();
+
+    final token = CancellationToken();
+    Future<void>.delayed(Duration(milliseconds: 1)).then((value) => token.cancel());
```
blagoev, line 1003: Is this testing it right. I suspect this is a race which is won by the Future so the token is canceled before the `writeAsync` has even started each time. 
Shouldn't we somehow trick it so we are sure the writeAsync had a chance to open a transaction and then the cancel happens.

blagoev: ie we are never calling `realm_async_begin_write` in this test

blagoev: In other words the delayed future here does not have any effect. 
In fact if I disable the `cancel` call here the test never completes with an failure.

blagoev: In fact if I comment out the `cancel` operation, this test never completes with a failure.

blagoev: Actually I think our use of `expecLater` is incorrect. This function should be used when we have async matchers,  and we don't have such. So using it here and awaiting on it is causing the test to hang. I pretty sure we use it incorrectly in most of the other places as well. But we can discuss that on the round table.

test/realm_test.dart
```diff
@@ -829,6 +854,219 @@ Future<void> main([List<String>? args]) async {
...
+  });
+
+  test('Realm.beginWriteAsync with cancellation token', () async {
+    final realm1 = getRealm(Configuration.local([Person.schema]));
+    final realm2 = getRealm(Configuration.local([Person.schema]));
+    final t1 = realm1.beginWrite();
+
+    final token = TimeoutCancellationToken(const Duration(milliseconds: 1), timeoutException: CancelledException());
+
+    await expectLater(realm2.beginWriteAsync(token), throwsA(isA<CancelledException>()));
```
blagoev, line 992: We should discuss the exception type. `CancelledException` is not an `RealmException` now.

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
nirinchev: You can try, it won't. It's a class level variable, so it can't ensure that it's not set to null after the if check.

test/realm_test.dart
nirinchev: `writeAsync` can't open a transaction because `t1` is open and write transactions are exclusive. I don't see a reason to believe there's a race here - `Future.delayed` continuation will execute when the thread is suspended, so it doesn't matter if it's 1 millisecond or 1 microseconds - it will always execute after whatever synchronous work is going on. The call to `writeAsync` is synchronous up until we return the `completer.future` and start waiting on Core.

I'm not sure what's wrong with our use of `expectLater` - the way I read the docs, it will execute the matchers when the future resolves. If you comment out the cancel you end up in a deadlock so it's not surprising the test hangs. The issue is that you're holding the write lock with `t1`, then awaiting the result of acquiring a write lock, which can not succeed until `t1` is relinquished. The fact that the test doesn't deadlock is evidence that the cancelation logic works.

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
blagoev: got it.

blagoev (approved):

lib/src/native/realm_core.dart
```diff
@@ -2645,6 +2696,65 @@ extension LevelExt on Level {
   }
 }
 
+class WriteCompleter with Cancellable implements Completer<void> {
+  final Completer<void> _internalCompleter;
+  final CancellationToken? _cancellationToken;
+  int? _id;
+  final Realm _realm;
+
+  set id(int value) {
```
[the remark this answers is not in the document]
blagoev: since it is not used externally we decided to leave it as is

lib/src/native/realm_core.dart
```diff
@@ -582,6 +583,72 @@ class _RealmCore {
     _realmLib.invokeGetBool(() => _realmLib.realm_commit(realm.handle._pointer), "Could not commit write");
   }
 
+  Future<void> beginWriteAsync(Realm realm, CancellationToken? ct) {
+    final completer = WriteCompleter(realm, ct);
+    if (completer.isCancelled) {
```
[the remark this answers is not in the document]
blagoev: my bad. we need the WriteCompleter for the Core call.

test/realm_test.dart
nirinchev: Resolved offline - for context - we added print statements to verify that the Core cancellation method is indeed called and the cancellation token is not already canceled by the time we reach `writeAsync`.

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
blagoev: After a discussion it seems my assumption was incorrect that we want the developer to be able to cancel twice. The thing is if Core returns false on the cancel operation then it will invoke the callback since the operation already completed. By that time it is too late to cancel the operation. So this is a race condition that which was won by Core succeeding the operation and not the user cancelling the op.
So the code is correct here.

test/realm_test.dart
blagoev: test is working fine after an investigation

nirinchev:
The macOS arm tests are failing due to a 500 error on cloud and the failures are unrelated to this change.

test/realm_test.dart
blagoev: Since the user can pass a custom exception and the exception is not thrown cause of SDK issue but rather use action we have decided to leave this Exception as non RealmException

PR #931, opened 2022-10-03
update Core to 12.9

blagoev:


nielsenko (commented):
Core is updated to f6bbd45cbb6cfb3a2d4212eeba59b4a46bca22d6 which does not match with the v12.9.0 tag which is at b86d2107fec4246b55af24a3b11e00275575f808

blagoev:
@nielsenko updated to the tagged release commit

nielsenko (approved):
LGTM

test/credentials_test.dart
```diff
@@ -284,12 +284,12 @@ Future<void> main([List<String>? args]) async {
     const String newPassword = "!@#!DQXQWD!223eda";
     final authProvider = EmailPasswordAuthProvider(app);
     await authProvider.registerUser(username, strongPassword);
-    await expectLater(() {
+    await expectLater(() async {
```
desistefanova, line 287: @blagoev Why do we need `async` and `await` inside `expectLater`? It is not needed. I have also added them by mistake in some of my old tests, but I think I have to remove them.

blagoev: its not problematic per se. you can do it or not do it.

Issue #656, opened 2022-06-07
Add input to prepare release workflow for version


nirinchev:
Should default to inferring the version from the changelog, but allow manually specifying a specific one.

PR #934, opened 2022-10-04
Add version input to the prepare release workflow

nirinchev:
Fixes https://github.com/realm/realm-dart/issues/656.

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -2,6 +2,11 @@ name: Prepare Release
 
 on:
   workflow_dispatch:
+    inputs:
+      version:
+        description: Use this to provide a version, instead of deriving it from the changelog.
+        required: false
+        type: string
 env:
   REALM_CI: true
 
@@ -16,9 +21,10 @@ jobs:
 
       - name: Update Changelog
         id: update-changelog
-        uses: realm/ci-actions/update-changelog@92b1c91f266809c4374947b8846c09c9049b6f3d
+        uses: realm/ci-actions/update-changelog@21e5e3c4acab316ccdeab748334c9c194d327fb7
         with:
           changelog: ${{ github.workspace }}/CHANGELOG.md
+          version: ${{ inputs.version }}
           version-suffix: '+beta'
 
       - name: Update pubspec.yaml


nielsenko (approved):
LGTM

.github/workflows/prepare-release.yml
```diff
@@ -2,6 +2,11 @@ name: Prepare Release
 
 on:
   workflow_dispatch:
+    inputs:
+      version:
```
desistefanova, line 6: How do we pass this argument `version`. Is it required when we manually trigger the workflow?

blagoev (approved):

.github/workflows/prepare-release.yml
```diff
@@ -2,6 +2,11 @@ name: Prepare Release
...
+        description: Use this to provide a version, instead of deriving it from the changelog.
+        required: false
```
blagoev, line 8: If it is not required what will be the version if we don't provide a value here

nirinchev: It'll be determined based on the changelog like it used to work before.

.github/workflows/prepare-release.yml
```diff
@@ -16,9 +21,10 @@ jobs:
 
       - name: Update Changelog
         id: update-changelog
-        uses: realm/ci-actions/update-changelog@92b1c91f266809c4374947b8846c09c9049b6f3d
+        uses: realm/ci-actions/update-changelog@21e5e3c4acab316ccdeab748334c9c194d327fb7
```
desistefanova, line 24: What is the reason of the change in [realm/ci-actions/update-changelog@21e5e3c4acab316ccdeab748334c9c194d327fb7](https://github.com/realm/ci-actions/blob/21e5e3c4acab316ccdeab748334c9c194d327fb7/update-changelog/src/helpers.ts#L118) ?

desistefanova (commented):
Added some questions that are not clear to me.

.github/workflows/prepare-release.yml
nirinchev: We pass it in the github UI when we manually trigger the workflow - it's not required and if you leave it blank, it will use the version from the changelog.

.github/workflows/prepare-release.yml
nirinchev: It's just picking up the latest version from main. The reason is that the ability to specify the version from the outside was added in a recent commit.

desistefanova (approved):

PR #938, opened 2022-10-05
Async Realm open with external package

### LINKED ISSUE #71: Open: Support async Realm open

desistefanova:
Support async Realm open using cancellation_tocken package. Fixes #71


lib/src/realm_class.dart
```diff
@@ -110,12 +112,58 @@ class Realm implements Finalizable {
...
+  /// * `cancellationToken` - an optional [CancellationToken] used to cancel the operation.
+  /// * `onProgressCallback` - a callback for receiving download progress notifications for synced [Realm]s.
+  ///
+  /// Returns [Future<Realm>] that completes with the `realm` once the remote realm is fully synchronized or with a [CancelledException] if operation is canceled.
+  /// When the configuration is [LocalConfiguration] this completes right after the local realm is opened or operation is canceled.
+  static Future<Realm> open(Configuration config, {CancellationToken? cancellationToken, ProgressCallback? onProgressCallback}) async {
+    Realm realm = (() => Realm(config)).asCancellable(
+      cancellationToken,
+      onCancel: (result) => result?.close(),
+    );
```
desistefanova, line 136: I created `asCancellable` not only for futures (which come from the `cancellation_token` package), but also for synchronous functions. That way, whenever the token is canceled, we can release the resource that was returned by the function. Even if `cancel` is called before or after the main function.
```Dart
    Realm realm = (() => Realm(config)).asCancellable(
      cancellationToken,
      onCancel: (result) => result?.close(),
    );
```

test/realm_test.dart
```diff
@@ -788,6 +788,220 @@ Future<void> main([List<String>? args]) async {
...
+    final configuration = Configuration.flexibleSync(user, [Task.schema]);
+
+    var cancellationToken = CancellationToken();
+    final realm = await RealmA.open(configuration, cancellationToken: cancellationToken);
+
+    expect(realm, isNotNull);
+    expect(realm.isClosed, false);
+
+    cancellationToken.cancel();
+    expect(realm.isClosed, false);
```
desistefanova, line 938: What should be the correct behavior if we cancel after the realm is returned?

PR #952, opened 2022-10-10
fix release commit message

blagoev:


desistefanova (commented):
Why is that change? It sounds to me OK the commit message to be 'Prepare for'.

nielsenko (commented):
This will make us deviate somewhat from what realm-core, realm-dotnet, realm-java and realm-js does, but it is in line with realm-kotlin and realm-swift. Not usre if this is anything we want to align on, and in what direction we then would want to go.

nirinchev:
I think the commit message is completely irrelevant and we should definitely not be spending time on aligning across SDKs. I don't care what it is - if you prefer square braces over no braces, go ahead and change it.

nielsenko:
I'm fine with either too

PR #960, opened 2022-10-12
Kn/open async test suggestions

nielsenko:
PR review suggestions for test of asynchronous `Realm.open`

### DIFF
--- a/test/realm_test.dart
+++ b/test/realm_test.dart
@@ -18,7 +18,9 @@
 
 // ignore_for_file: unused_local_variable, avoid_relative_lib_imports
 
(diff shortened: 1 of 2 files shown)


desistefanova (commented):
Yes, this is also an option. I changed this check method to return `true` or `false` and renamed it to `thenIsCancelled` based on your request to distinguish both cases. If you agree with the change in [Open realm async](https://github.com/realm/realm-dart/pull/731#top), please close this PR.

nielsenko:
Closing this as the spirit of the change has already been applied

PR #962, opened 2022-10-13
add a test for working encryption

blagoev:
use more clear test names

### DIFF
--- a/test/configuration_test.dart
+++ b/test/configuration_test.dart
@@ -555,7 +555,7 @@ Future<void> main([List<String>? args]) async {
     expect(disconnectedRealm.find<Task>(oid), isNotNull);
   });
 
-  test('Configuration.local set invalid encryption key size', () {
+  test('Configuration.local set too short encryption key size', () {
     List<int> key = [1, 2, 3];
     expect(
       () => Configuration.local([Car.schema], encryptionKey: key),
@@ -577,7 +577,7 @@ Future<void> main([List<String>? args]) async {
     expect(config.encryptionKey, key);
   });
 
-  baasTest('FlexibleSyncConfiguration set invalid encryption key size', (appConfiguration) async {
+  baasTest('FlexibleSyncConfiguration set too long encryption key size', (appConfiguration) async {
     final app = App(appConfiguration);
     final credentials = Credentials.anonymous();
     final user = await app.logIn(credentials);
--- a/test/configuration_test.dart
+++ b/test/configuration_test.dart
@@ -555,7 +555,7 @@ Future<void> main([List<String>? args]) async {
     expect(disconnectedRealm.find<Task>(oid), isNotNull);
   });
 
-  test('Configuration set short encryption key', () {
+  test('Configuration.local set invalid encryption key size', () {
     List<int> key = [1, 2, 3];
     expect(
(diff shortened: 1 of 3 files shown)


test/realm_test.dart
```diff
@@ -789,33 +791,49 @@ Future<void> main([List<String>? args]) async {
     expect(stored.location.name, 'Europe/Copenhagen');
   });
 
-  test('Realm - open local not encrypted realm with encryption key', () {
-    openEncryptedRealm(null, generateValidKey());
+  test('Realm - encrytpion works', () {
+    var config = Configuration.local([Friend.schema], path: p.join(Configuration.defaultStoragePath, "${generateRandomString(8)}.realm"));
```
nirinchev, line 795: Why do we need this path? I guess it's because you want different paths for the encrypted/unencrypted realm, but a simpler solution imo would be something like:

```dart
final unencryptedConfig = Configuration.local([Friend.schema]);
final encryptedConfig = Configuration.local([Friend.schema], path: '${unencryptedConfig.path}_encrypted');
```

test/realm_test.dart
```diff
@@ -789,33 +791,49 @@ Future<void> main([List<String>? args]) async {
     expect(stored.location.name, 'Europe/Copenhagen');
   });
 
-  test('Realm - open local not encrypted realm with encryption key', () {
-    openEncryptedRealm(null, generateValidKey());
+  test('Realm - encrytpion works', () {
```
nirinchev, line 794:
```suggestion
  test('Realm - encryption works', () {
```

test/configuration_test.dart
```diff
@@ -555,28 +555,29 @@ Future<void> main([List<String>? args]) async {
     expect(disconnectedRealm.find<Task>(oid), isNotNull);
   });
 
-  test('Configuration set short encryption key', () {
+  test('Configuration.local set invalid encryption key size', () {
```
nirinchev, line 558: I like the previous name better here - maybe we can still reword it, but I think it was important that we could easily see that we're testing both short and long encryption keys. Now both tests just state "invalid", so you need to go into the test implementation to confirm that both cases are covered.

desistefanova (commented):
Looks OK. I agree with @nirinchev comments

test/configuration_test.dart
blagoev: will revert

test/realm_test.dart
blagoev: I need a completely new realm that is not the default one. Since I am reading it as a string

desistefanova (approved):

test/realm_test.dart
nirinchev: The default realm is randomly generated for each test, so it should be as good as new.

blagoev: yeah I prefer to be more explicit here. So two new realms with random names created in the test. Names are not relevant much, but the fact, that we are creating them brand new, is relevant.

blagoev: I am not sure what the problem is in the way test is written. As I said I prefer to have this test explicitly set the name of the two files highlighting the fact we are creating a brand new realm and reading it as a Dart string. Also this test was send as an example to the docs team so they can write their own. One more reason I have chosen to set explicitly the path there.

PR #974, opened 2022-10-19
Throws last core error when beginWriteAsync and commitWriteAsync

### LINKED ISSUE #1047: Check if beginWriteAsync and commitWriteAsync rethrow core errors

desistefanova:
Add tests to check that write is not possible when:

-  Realm is closed,
-  Realm is read-only, 
-  write and beginWrite are nested,
-  Realm is frozen.

Throw core errors as RealmExceptions.
Fixes #1047


nirinchev (changes requested):

lib/src/native/realm_core.dart
```diff
@@ -2378,6 +2382,12 @@ extension _RealmLibraryEx on RealmLibrary {
     }
     return result;
   }
+
+  void invokeGetInt(int Function() callback, [String? errorMessage]) {
```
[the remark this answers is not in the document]
desistefanova: What is the bug? Is that the previous error is not cleared?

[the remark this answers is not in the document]
nirinchev: The bug is the API doesn't indicate whether an error has occurred. The guiding principle of the C API is such that if an API can throw an exception, it should return a boolean and the actual value should be passed as an out parameter. So we should not have a `invokeGetInt`.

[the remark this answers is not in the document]
desistefanova: An issue is created https://github.com/realm/realm-core/issues/5954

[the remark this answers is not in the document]
desistefanova: @nirinchev There is a fix already in core 12.13 https://github.com/realm/realm-core/pull/6039
I will open this PR soon to complete it.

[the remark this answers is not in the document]
nielsenko: @desistefanova / @nirinchev The fix is folded into #1042

CHANGELOG.md
```diff
@@ -20,6 +21,7 @@
 * Removed the ".tmp_compaction_space" file being left over after compacting a Realm on Windows. (Core upgrade).
 * Restore fallback to full barrier when F_BARRIERSYNC is not available on Apple platforms. (Core upgrade, since v0.8.0+rc)
 * Fixed wrong assertion on query error that could result in a crash. (Core upgrade)
+* Writing to a read-only realm throws `RealmException` insted to block the isolate. ([#974](https://github.com/realm/realm-dart/pull/974))
```
nirinchev, line 24: This is worded weirdly - do you mean "instead of blocking the isolate"?

test/realm_test.dart
```diff
@@ -576,12 +577,12 @@ Future<void> main([List<String>? args]) async {
     expect(returnedCar, car);
   });
 
-  test('Realm write inside another write throws', () {
+  test('Realm write inside another write throws', () async {
     final config = Configuration.local([Car.schema]);
     final realm = getRealm(config);
-    realm.write(() {
+    realm.write(() async {
```
nirinchev, line 583: Why is this async?

test/realm_test.dart
```diff
@@ -576,12 +577,12 @@ Future<void> main([List<String>? args]) async {
     expect(returnedCar, car);
   });
 
-  test('Realm write inside another write throws', () {
+  test('Realm write inside another write throws', () async {
```
nirinchev, line 580: This shouldn't be async

nirinchev (approved):
Looks good - just a few small questions from me.

lib/src/realm_class.dart
```diff
@@ -496,12 +490,6 @@ class Realm implements Finalizable {
     }
   }
 
-  void _ensureWritable() {
```
blagoev, line 499: I am not sure I am following. Has this moved to core or why we are removing this from here?

test/test.dart
```diff
@@ -710,3 +710,16 @@ Future<void> enableAllAutomaticRecovery() async {
     await client.setAutomaticRecoveryEnabled(appName, true);
   }
 }
+
+
+extension $WriteExtension on Realm {
```
blagoev, line 715: Why do we need this for all tests. I think this is very specific to the tests we are changing in this PR so I would suggest moving this close to where it is used. I don't think we are going to use these in any other place.

blagoev, line 715: why the `$` here

test/test.dart
```diff
@@ -710,3 +710,16 @@ Future<void> enableAllAutomaticRecovery() async {
...
+  void writesExpectException<TException>(String exceptionMessage) {
```
blagoev, line 716: I would rename these methods to `writeAndExpectException`

blagoev, line 716: Unlike .NET we are not really naming type arguments like that `TException`. I would suggest doing something like 
```suggestion
  void writesExpectException<T extends Exception>(String message) {
```

test/test.dart
blagoev: also making this extension private after the move

blagoev: I would also consider not using an extension since it is visible in the whole file and this is not useful in that file as well except the test that are using it. I would move this to a regular method

blagoev (changes requested):

lib/src/realm_class.dart
desistefanova: Yes, this check is done in core and the exception is thrown by core. We don't need to it here. It doesn't cover all the cases that core checks.

test/test.dart
desistefanova: We have one case that throws `RealmClosedError`. It is not an Exception. This is in test  'Realm write after realm is closed'. I will just rename `TException` to `T`. Or maybe `RealmClosedError` should have been Exception?

test/test.dart
desistefanova: I wanted to hide this extension. But, yes, it is better they to be methods.

test/test.dart
blagoev: This test should not use `writesExpectException`. Lets this method only in the tests that require the double writes.

test/configuration_test.dart
```diff
@@ -174,16 +174,6 @@ Future<void> main([List<String>? args]) async {
     expect(cars.length, 1);
   });
 
-  test('Configuration readOnly - writing on read-only Realms throws', () {
```
blagoev, line 177: Is this test not needed anymore and why is that?

test/realm_test.dart
```diff
@@ -268,14 +268,26 @@ Future<void> main([List<String>? args]) async {
...
-  test('Realm write after realm is closed', () {
+  test('Realm write after realm is closed', () async {
     var config = Configuration.local([Car.schema]);
     var realm = getRealm(config);
 
     final car = Car('Tesla');
 
     realm.close();
-    expect(() => realm.write(() {}), throws<RealmClosedError>("Cannot access realm that has been closed"));
+    _writeAndExpectException<RealmClosedError>(realm, "Cannot access realm that has been closed");
```
blagoev, line 278: We should not be changing all the test to use this API. Lets use this method only where it was needed for the double write.

blagoev (changes requested):

test/realm_test.dart
```diff
@@ -1647,3 +1631,13 @@ When newWhen([tz.TZDateTime? time]) {
 extension _IterableEx<T> on Iterable<T> {
   Iterable<T> except(T exclude) => where((o) => o != exclude);
 }
+
+void _writeAndExpectException<T>(Realm realm, String exceptionMessage) {
```
blagoev, line 1635: I think this should only be used where needed. And I suggest we make it as 
```suggestion
void _writeAndExpectException<T extends Execption>(Realm realm, String exceptionMessage) {
```

blagoev (changes requested):

test/realm_test.dart
desistefanova: The reason to use it was because I had to test all the cases where async write throws exceptions. And async write when realm is closed is one of them.

test/configuration_test.dart
desistefanova: It is moved to realm_test.dart in order to be able to use `_writeAndExpectException`, since you wanted `_writeAndExpectException` to be private.

test/realm_test.dart
desistefanova: I used it where it is needed. If you don't like these methods I can copy paste their implementation to all the places where it is needed.

test/realm_test.dart
blagoev: I see you point. In this case could we rename this to `_expectAllWritesToThrow<T extends Exception>`
For the `RealmClosedError` I think we should leave it in the handle getter but also check in each write method if the realm is closed and throw a `RealmException("Write failed. The realm is already closed.")`

test/realm_test.dart
blagoev: Yes, sorry. I got your meaning now. added a suggestion in the comment above.

test/realm_test.dart
desistefanova: If you search by `RealmClosedError` you can see that it is used in the the handle getters for list, results, realm, realm object, session and so many places and so many tests. I suggest if you want we to do this it to be in a separate PR. It will introduce bricking changes.

desistefanova: I can rename these methods to `_expectWriteToThrow `and `_expectWriteAsyncToThrow` if you prefer. I need to have them separately because for the one of the tests 'Realm write inside another write throws' I expect only `write` and `beginWrite` to throw inside another write. The async writes are allowed inside another write block.

blagoev: `_expectAllWritesToThrow` and `_expectAllAsyncWritesToThrow`?

blagoev: Lets leave `RealmCloseError` as it is

desistefanova: Done

test/test.dart
desistefanova: I'm leaving this with <T> because of `RealmClosedError`

test/realm_test.dart
```diff
@@ -1647,3 +1631,13 @@ When newWhen([tz.TZDateTime? time]) {
 extension _IterableEx<T> on Iterable<T> {
   Iterable<T> except(T exclude) => where((o) => o != exclude);
 }
+
+void _expectAllWritesToThrow<T>(Realm realm, String exceptionMessage) {
```
blagoev, line 1635:
```suggestion
void _expectAllWritesToThrow<T extends Exception>(Realm realm, String exceptionMessage) {
```

Whe should be throwing a RealmException on write. I had suggested previously to check that in all writes and throw it instead of RealmClosedError.

blagoev (changes requested):

nielsenko (approved):
LGTM - might have some changes regarding exceptions/errors later

test/realm_test.dart
desistefanova: We decided in a call that no additional check for r`ealm.isClosed` is needed in write methods. We already check if the handle is released.

blagoev (approved):

PR #975, opened 2022-10-19
Port embedded object support to master

nirinchev:
Alternative to https://github.com/realm/realm-dart/pull/892

TODO:
- [x] Tests for [MapTo]
- [x] Tests for freezing
- [x] Tests for changes
- [x] Changelog
- [x] Relax the generic constraint of `Realm.delete` to allow `RealmObjectBase`

PR #978, opened 2022-10-20
Improve robustness of API key tests

nirinchev:
I don't think we have an issue for this, but took a stab at adding some retry logic to our API key tests.

### DIFF
--- a/test/test.dart
+++ b/test/test.dart
@@ -551,21 +551,29 @@ Future<User> loginWithRetry(App app, Credentials credentials, {int retryCount = 
 }
 
 Future<void> waitForCondition(
-  bool Function() condition, {
+  FutureOr<bool> Function() condition, {
   Duration timeout = const Duration(seconds: 1),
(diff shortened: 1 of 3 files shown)


blagoev (approved):

test/test.dart
```diff
@@ -551,21 +551,29 @@ Future<User> loginWithRetry(App app, Credentials credentials, {int retryCount =
...
-    Future.doWhile(() async {
-      if (condition()) {
-        return false;
-      }
-      await Future<void>.delayed(retryDelay);
-      return true;
-    })
-  ]);
+}) {
+  return waitForConditionWithResult<bool>(() => condition(), (value) => value == true);
```
desistefanova, line 559:
```suggestion
return waitForConditionWithResult<bool>(() => condition(), (value) => value);
```

nirinchev: Fixed in https://github.com/realm/realm-dart/pull/975.

PR #979, opened 2022-10-20
Expose embeddedObject.getParent

nirinchev:
This is a follow up on the embedded objects PR that I pulled to a separate branch because it depends on https://github.com/realm/realm-core/pull/5960.

### DIFF
--- a/lib/src/native/realm_core.dart
+++ b/lib/src/native/realm_core.dart
@@ -44,6 +44,7 @@ import '../scheduler.dart';
 import '../subscription.dart';
 import '../user.dart';
 import '../session.dart';
+import '../util.dart';
 import 'realm_bindings.dart';
 import '../migration.dart';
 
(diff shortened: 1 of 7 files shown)


desistefanova (commented):
We have to update the core version in dart Changelog.

desistefanova (commented):
Do you think having these new functions can help us to implement `DynamicRealm` `find` and `all` to support embedded objects?
Or it must be similar to Realm class and to support only RealmObjects?

nirinchev:
They should be equivalent - i.e. we should not allow users to lookup embedded objects directly.

lib/src/realm_object.dart
```diff
@@ -411,7 +411,17 @@ mixin RealmObjectBase on RealmEntity implements Finalizable {
 mixin RealmObject on RealmObjectBase {}
 
 /// @nodoc
-mixin EmbeddedObject on RealmObjectBase {}
+mixin EmbeddedObject on RealmObjectBase {
+  RealmObjectBase? getParent() {
```
nielsenko, line 415: Should we make this a `parent` getter instead?

blagoev: +1

blagoev, line 415: no api doc

lib/src/util.dart
```diff
@@ -0,0 +1,6 @@
+class Tuple<T1, T2> {
```
blagoev, line 1: we could move this to realm_common_base.dart and delete the utils.dart

lib/src/native/realm_core.dart
```diff
@@ -1054,6 +1067,10 @@ class _RealmCore {
     return _realmLib.realm_equals(first._pointer.cast(), second._pointer.cast());
   }
 
+  int getObjectKey(RealmObjectBase obj) {
```
blagoev, line 1070: is this used somewhere ?

blagoev (commented):
It's ok but I have the feeling we missed something while rebasing. Since this `getObjectKey` does not seem to be used anywhere

blagoev:
should have a changelog entry

blagoev (changes requested):
can we check the rebase and also add a changelog entry. Just making it more visible here

lib/src/realm_object.dart
nirinchev: I don't think it's a good idea, because users may want to use `parent` as a property on their models.

nielsenko: Had a long discussion about this with @nirinchev and we settled on trying with a getter

nielsenko: Added api doc on `getParent` now `get parent`

lib/src/util.dart
nielsenko: Done

blagoev (approved):

desistefanova (approved):

nielsenko (approved):

PR #982, opened 2022-10-21
Generator fixes

blagoev:
print the exception stacktrace if generator encounters an exception
throw on no expected file found. - previously if an expected file was not found the test did succeed. This is not ideal since the file might be there with the wrong name giving a false impression that the test succeeds or not there at all.
generator tests are named after the exact file name. - previously the underscore was removed. This mostly helps when filtering the test runs in the console

### DIFF
--- a/generator/test/test_util.dart
+++ b/generator/test/test_util.dart
@@ -15,16 +15,18 @@ Map<String, String> getListOfTestFiles(String directory) {
   for (var file in files) {
     if (_path.extension(file.path) == '.dart' && !file.path.endsWith('g.dart')) {
       var expectedFileName = _path.setExtension(file.path, '.expected');
-      if (!files.any((f) => f.path == expectedFileName)) expectedFileName = '';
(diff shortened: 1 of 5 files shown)


generator/test/test_util.dart
```diff
@@ -15,16 +15,18 @@ Map<String, String> getListOfTestFiles(String directory) {
...
+        throw "Expected file not found. $expectedFileName";
+      }
       result.addAll({_path.basename(file.path): _path.basename(expectedFileName)});
     }
   }
   return result;
 }
 
-Future<dynamic> generatorTestBuilder(String directoryName, String inputFileName, [String expectedFileName = ""]) async {
+Future<dynamic> generatorTestBuilder(String directoryName, String inputFileName, [String? expectedFileName]) async {
```
nirinchev, line 27: Why is this optional now? If we throw an exception when there's no expected filename, I would imagine this will always be called with some value.

nirinchev (approved):
Minor question from me, otherwise looks good.

PR #983, opened 2022-10-21
Reorder AuthProviderType values

### LINKED ISSUE #992: [Bug]: Wrong user.provider type for facebook, google and apple
### What happened?

When login with google credentials it returns user.provider = AuthProviderType.facebook.

### Repro steps

 ```dart
  final tokenId ="some valid google token id";
  final credentials = Credentials.googleIdToken(tokenId);
  final user = await app.logIn(credentials);
  expect(user.provider, AuthProviderType.google);// actually it was facebook
```

### Version

0.5.0+beta and 0.6.0+beta

### What Realm SDK flavor are you using?

MongoDB Atlas (i.e. Sync, auth, functions)

Dart standalone application

### Client OS and version

Windows 11

### Code snippets

 ```dart
  final tokenId ="some valid google token id";
  final credentials = Credentials.googleIdToken(tokenId);
  final user = await app.logIn(credentials);
  expect(user.provider, AuthProviderType.google);// actually it was facebook
```

_No response_

### Relevant log output

_No response_

desistefanova:
- Returned `user.provider` was wrong for facebook, google and apple. All the other providers were correct, because their numbers remains the same.

- When we login with `Credentials.anonymous(reuseCredentials: false)` the returned user.provider is also `AuthProviderType.anonymous` instead of `anonymousNoReuse`. To be discussed.

Fixes  #992

### DIFF
--- a/lib/src/credentials.dart
+++ b/lib/src/credentials.dart
@@ -139,6 +139,8 @@ extension CredentialsInternal on Credentials {
   }
 
   RealmAppCredentialsHandle get handle => _handle;
+
+  AuthProviderType readProviderFromCore() => realmCore.userGetCredentialsProviderType(this);
 }
 
 /// A class, encapsulating functionality for users, logged in with [Credentials.emailPassword()].
--- a/lib/src/credentials.dart
+++ b/lib/src/credentials.dart
@@ -140,7 +140,8 @@ extension CredentialsInternal on Credentials {
 
   RealmAppCredentialsHandle get handle => _handle;
 
-  AuthProviderType readProviderFromCore() => realmCore.userGetCredentialsProviderType(this);
+  /// This should be only used for testing
+  AuthProviderType get credentialsProviderType => realmCore.userGetCredentialsProviderType(this);
 }
 
 /// A class, encapsulating functionality for users, logged in with [Credentials.emailPassword()].
--- a/lib/src/credentials.dart
+++ b/lib/src/credentials.dart
@@ -64,15 +64,8 @@ extension AuthProviderTypeInternal on AuthProviderType {
   static AuthProviderType getByValue(int value) {
     if (value == AuthProviderType._anonymousNoReuse._value) {
       return AuthProviderType.anonymous;
-    } else {
-      for (final type in AuthProviderType.values) {
-        if (type._value == value) {
-          return type;
-        }
-      }
     }
-
-    throw ArgumentError('Invalid AuthProviderType value: $value');
+    return AuthProviderType.values.firstWhere((v) => v._value == value);
   }
 }
 
--- a/lib/src/credentials.dart
+++ b/lib/src/credentials.dart
@@ -80,60 +80,43 @@ extension AuthProviderTypeInternal on AuthProviderType {
 /// {@category Application}
 class Credentials implements Finalizable {
   final RealmAppCredentialsHandle _handle;
-  final AuthProviderType provider;
 
   /// Returns a [Credentials] object that can be used to authenticate an anonymous user.
   /// Setting [reuseCredentials] to [false] will create a new anonymous user, upon [App.logIn].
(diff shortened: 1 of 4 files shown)


nirinchev (changes requested):
This looks fine, but let's add tests to make sure it doesn't regress. You can use `realm_auth_credentials_get_provider` to make sure the provider is the same. I'm thinking of tests like:

```dart
final credentials = Credentials.anonymous();
expect(credentials.readProviderFromCore(), AuthProviderType.anonymous);
expect(credentials.provider, AuthProviderType.anonymous);
```

where `readProviderFromCore` is an internal function that calls `realm_auth_credentials_get_provider`.

test/credentials_test.dart
```diff
@@ -38,6 +39,8 @@ Future<void> main([List<String>? args]) async {
 
     expect(user1, user2);
     expect(user1, isNot(user3));
+    expect(user1.provider, AuthProviderType.anonymous);
+    expect(user3.provider, AuthProviderType.anonymous);
```
desistefanova, line 43: After login with with anonymous `reuseCredentials: false`  we receive `AuthProviderType.anonymous` instead of `AuthProviderType.anonymousNoReuse`. It is because for the server both credentials are "anon-user" and this is the way they are mapped in `provider_type_from_enum`. @nirinchev do you think we have to create an issue about that?

nirinchev: No, apparently that's intentional. We should probably remove the noreuse provider type. As that's not something users should ever deal with.

lib/src/credentials.dart
```diff
@@ -139,6 +139,8 @@ extension CredentialsInternal on Credentials {
   }
 
   RealmAppCredentialsHandle get handle => _handle;
+
+  AuthProviderType readProviderFromCore() => realmCore.userGetCredentialsProviderType(this);
```
blagoev, line 143:
```suggestion
  /// This should be only used for testing
  AuthProviderType get credentialsProviderType => realmCore.userGetCredentialsProviderType(this);
```
we have similar comment elsewhere and also suggested a change to a getter and the rename

blagoev (approved):
Its ok, I added small suggestion

test/credentials_test.dart
desistefanova: Maybe we have to think about it, because when we pass the credentials to core they expect `anonymousNoReuse` to know whether to reuse the users or not.

nirinchev: Right, but we construct the credentials by calling `realmCore.createAppCredentialsAnonymous` - we don't provide the actual enum value to anyone. We store it as a field on the dart credentials, but I don't believe that's actually ever used anywhere and we might as well turn that into a property that reads the value from Core.

blagoev:
@desistefanova why not changelog. Isn't this a fix or this was a fix on a version we have not released yet?

test/credentials_test.dart
desistefanova: OK, I made anonymousNoReuse hidden.

desistefanova:
> @desistefanova why not changelog. Isn't this a fix or this was a fix on a version we have not released yet?

Yes I will add a row. It is a fix for a released version.

lib/src/credentials.dart
```diff
@@ -61,9 +62,13 @@ enum AuthProviderType {
...
-        return type;
+    if (value == AuthProviderType._anonymousNoReuse._value) {
+      return AuthProviderType.anonymous;
+    } else {
+      for (final type in AuthProviderType.values) {
+        if (type._value == value) {
+          return type;
+        }
       }
     }
```
nirinchev, line 73:
```suggestion
    if (value == AuthProviderType._anonymousNoReuse._value) {
      return AuthProviderType.anonymous;
    }
    
    return AuthProviderType.values.firstWhere((v) => v._value == value);
```

And we can remove the argument exception at the end.

nirinchev (approved):

PR #984, opened 2022-10-24
Don't assume api key order in fetchAll

nirinchev:
This should fix our newly failing tests.

### DIFF
--- a/test/user_test.dart
+++ b/test/user_test.dart
@@ -230,7 +230,8 @@ Future<void> main([List<String>? args]) async {
     final fetched = await user.apiKeys.fetchAll();
 
     for (var i = 0; i < 5; i++) {
-      expectApiKey(fetched[i], original[i]);
+      final fetchedKey = fetched.firstWhere((key) => key.id == original[i].id);
+      expectApiKey(fetchedKey, original[i]);
     }
   });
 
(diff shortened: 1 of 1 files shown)


nielsenko (approved):

blagoev (approved):

test/user_test.dart
```diff
@@ -356,20 +359,26 @@ Future<void> main([List<String>? args]) async {
...
+    final refetchedFirst = refetched.firstWhere((k) => k.id == first.id);
 
-    expect(refetched[1].id, second.id);
-    expect(refetched[1].isEnabled, true);
+    expect(refetchedFirst.id, first.id);
+    expect(refetchedFirst.isEnabled, true);
+
+    final refetchedSecond = refetched.firstWhere((k) => k.id == second.id);
+
+    expect(refetchedSecond.id, second.id);
```
desistefanova, line 380:
```suggestion
    expect( refetched.firstWhere((k) => k.id == second.id), isNotNull);
```
We can use even singleWhere. This is relevant for all the tests.

nirinchev: `firstWhere` will not return `null`. But I changed it to use `singleWhere` to make sure we don't get duplicate entries.

test/user_test.dart
```diff
@@ -356,20 +359,26 @@ Future<void> main([List<String>? args]) async {
...
-    expect(refetched[0].isEnabled, true);
 
-    expect(refetched[1].id, second.id);
-    expect(refetched[1].isEnabled, true);
+    final refetchedFirst = refetched.singleWhere((k) => k.id == first.id);
+    expect(refetchedFirst.id, first.id);
+    expect(refetchedFirst.isEnabled, true);
+
+    final refetchedSecond = refetched.singleWhere((k) => k.id == second.id);
+    expect(refetchedSecond.id, second.id);
```
desistefanova, line 380: I think we don't need these checks anymore. We already do it using singleWhere.

desistefanova (approved):

PR #987, opened 2022-10-24
Don't skip any tests on flutter

nielsenko:
- Add flavor specific copyOldRealmTo method, and bundle old_realm as asset in driver tests

Fix: #805


nielsenko:
This is a bit of a hack..

desistefanova (approved):

blagoev (approved):

flavor_helpers.dart
```diff
@@ -0,0 +1,6 @@
+import 'dart:io';
+
+Future<void> copyOldFormatRealmTo(String path) async {
+  const realmBundleFile = 'flutter/realm_flutter/tests/data/realm_files/old-format.realm';
```
blagoev, line 4: I have missed this one. Lets keep packages separate. Isn't this file already available in tets/data/realm_files for the dart tests?

blagoev (changes requested):

flavor_helpers.dart
nielsenko: I need to move to a place where the tests app can pick it up. It seemed redundant to have the it two places.

desistefanova (approved):

flavor_helpers.dart
```diff
@@ -0,0 +1,6 @@
+import 'dart:io';
+
+Future<void> copyOldFormatRealmTo(String path) async {
```
nirinchev, line 3: Can we generalize this a little? Like `copyBundledFile(String name, String to)`? That way we can later add more files to the `data/realm_files` folder and they can use the same function?

nielsenko: Yes, but perhaps we can just wait until the need arise?

flavor_helpers.dart
nielsenko: As discussed, I will use a symlink instead

flavor_helpers.dart
nirinchev: We can, my only concern is that with such a specific name, people may not realize that this is the method they need to extend in case they need it. Not a dealbreaker for me though.

blagoev (approved):

nirinchev (approved):

PR #988, opened 2022-10-24
Update realm-core to v12.11.0

nielsenko:


desistefanova (approved):

blagoev (commented):
Should we add some changelog entries here for the core fixes
>Fix a use-after-free if the last external reference to an encrypted Realm was closed between when a client reset error was received and when the download of the new Realm began.
this one seems deserving an entry

CHANGELOG.md
```diff
@@ -6,13 +6,15 @@
 * Added `MutableSubscriptionSet.removeByType` for removing subscriptions by their realm object type. (Issue [#317](https://github.com/realm/realm-dart/issues/317))
 
 ### Fixed
-* None
+* CompensatingWriteErrorInfo reported string primary keys as boolean values instead. (Core upgrade)
```
nirinchev, line 9: This is not exposed in the dart SDK.

CHANGELOG.md
```diff
@@ -6,13 +6,15 @@
...
+* Fix a use-after-free if the last external reference to an encrypted Realm was closed between when a client reset error was received and when the download of the new Realm began. (Core upgrade)
+* Fixed an assertion failure during client reset with recovery when recovering a list operation on an embedded object that has a link column in the path prefix to the list from the top level object. (Core upgrade)
```
nirinchev, line 11: This is not exposed either, so users wouldn't have experienced it.

nielsenko: It now says:
> ### Fixed
>* Fix a use-after-free if the last external reference to an encrypted Realm was closed between when a client reset error was received and when the download of the new Realm began. (Core upgrade)

nirinchev: The thing is, we don't have client reset with recovery, so there's no way for users to have gotten in this situation. This is only for client reset with recovery and doesn't affect the manual client reset mechanism.

nirinchev (approved):

desistefanova (approved):

blagoev (approved):

PR #989, opened 2022-10-24
lazy init apikeyclient

blagoev:


desistefanova (approved):

nirinchev (changes requested):
I thought in dart, late fields are lazy initialized, so the client should not be constructed until the `_apiKeys` field is accessed.

blagoev:
@nirinchev yes, I don't know how I was reading that code last night. Lets close this PR.

PR #990, opened 2022-10-25
sync readme

blagoev:


README.md
```diff
@@ -9,6 +9,15 @@ This repository holds the source code for the Realm SDK for Flutter™ and Dart
 
 **This project is in the Beta stage. The API should be quite stable, but occasional breaking changes may be made.**
 
+## Features
+
+* **Mobile-first:** Realm is the first database built from the ground up to run directly inside phones, tablets, and wearables.
+* **Simple:** Data is directly [exposed as objects](https://docs.mongodb.com/realm/dotnet/objects/) and [queryable by code](https://docs.mongodb.com/realm/dotnet/query-engine/), removing the need for ORM's riddled with performance & maintenance issues. Plus, we've worked hard to [keep our API down to just a few common classes](https://pub.dev/documentation/realm/latest/): most of our users pick it up intuitively, getting simple apps up & running in minutes.
```
desistefanova, line 15:
```suggestion
* **Simple:** Data is directly [exposed as objects](https://www.mongodb.com/docs/realm/sdk/flutter/realm-database/define-realm-object-schema/) and [queryable by code](https://www.mongodb.com/docs/realm/sdk/flutter/realm-database/read-and-write-data/), removing the need for ORM's riddled with performance & maintenance issues. Plus, we've worked hard to [keep our API down to just a few common classes](https://pub.dev/documentation/realm/latest/): most of our users pick it up intuitively, getting simple apps up & running in minutes.
```

README.md
```diff
@@ -9,6 +9,15 @@ This repository holds the source code for the Realm SDK for Flutter™ and Dart
...
+* **Modern:** Realm supports latest Dart and Flutter versions and is build with sound null-safety.
+* **Fast:** Realm is faster than even raw SQLite on common operations while maintaining an extremely rich feature set.
+* **[Device Sync](https://www.mongodb.com/atlas/app-services/device-sync)**: Makes it simple to keep data in sync across users, devices, and your backend in real-time. [Get started](http://mongodb.com/realm/register?utm_medium=github_atlas_CTA&utm_source=realm_dart_github) for free with a template application that includes a cloud backend and Sync.
+
+
```
desistefanova, line 20: Maybe it is good to mention the new link to Get Started page for flexible Sync https://www.mongodb.com/docs/atlas/app-services/get-started/

desistefanova (changes requested):

README.md
blagoev: that url is not the same so I suggest we keep it to .NET until we have the correct section in the Flutter docs https://www.mongodb.com/docs/realm/sdk/flutter/realm-database/define-realm-object-schema/

README.md
blagoev: We should be in sync with the rest of the SDKS

README.md
desistefanova: I think these two links have the same meaning. https://www.mongodb.com/docs/realm/sdk/flutter/realm-database/define-realm-object-schema/ and https://www.mongodb.com/docs/realm/sdk/flutter/realm-database/read-and-write-data/
Do we expect to have the same pages as .NEt SDK?

blagoev: They do not have similar meaning. The contents is very different. And yes we should have these fundamental pages documented for Flutter as well.

desistefanova: OK

desistefanova (approved):

README.md
nirinchev: I think we can as well remove the links and leave it as text. I think linking to the .NET docs here will be more confusing than helpful.

PR #996, opened 2022-10-27
Backlink support

nielsenko:
We now support named back links
```dart
@RealmModel()
class _Source {
  _Target? target
}

@RealmModel()
class _Target {
  @Backlink(#target)
  late Iterable<_Source> backlinks;
}

...
final target = Target();
final source = realm.write(() => realm.add(Source(target: target)));

expect(source.target, target);
expect(target.backlinks, [source]);
```

Resolves: #693

### DIFF
--- a/common/lib/src/realm_common_base.dart
+++ b/common/lib/src/realm_common_base.dart
@@ -97,3 +97,10 @@ class Indexed {
 class Ignored {
   const Ignored();
 }
+
+/// Indicates a backlink property.
+/// {@category Annotations}
+class Backlink {
+  final Symbol symbol;
+  const Backlink(this.symbol);
+}
--- a/common/lib/src/realm_common_base.dart
+++ b/common/lib/src/realm_common_base.dart
@@ -98,9 +98,10 @@ class Ignored {
   const Ignored();
 }
 
-/// Indicates a backlink property.
+/// Indicates that the field it decorates is the inverse end of a relationship.
 /// {@category Annotations}
 class Backlink {
-  final Symbol symbol;
-  const Backlink(this.symbol);
+  /// The name of the field in the other class that links to this class.
+  final Symbol fieldName;
+  const Backlink(this.fieldName);
 }
--- a/generator/lib/src/class_element_ex.dart
+++ b/generator/lib/src/class_element_ex.dart
@@ -51,9 +51,7 @@ extension on Iterable<FieldElement> {
         element: field,
         primarySpan: field.span!,
         primaryLabel: 'second primary key',
-        secondarySpans: {
-          ...{for (final p in primaryKeys..removeAt(1)) p.fieldElement.span!: ''},
-        },
+        secondarySpans: {for (final p in primaryKeys..removeAt(1)) p.fieldElement.span!: ''},
       );
     }
   }
@@ -87,8 +85,13 @@ extension ClassElementEx on ClassElement {
       }
 
       if (!modelName.endsWith(suffix)) {
-        throw RealmInvalidGenerationSourceError('Missing suffix on realm model name',
-            element: this, primarySpan: span, primaryLabel: 'missing suffix', todo: 'Align class name to have suffix $suffix,');
+        throw RealmInvalidGenerationSourceError(
+          'Missing suffix on realm model name',
+          element: this,
+          primarySpan: span,
+          primaryLabel: 'missing suffix',
+          todo: 'Align class name to have suffix $suffix,',
+        );
       }
 
       // Remove suffix and prefix, if any.
@@ -170,7 +173,7 @@ extension ClassElementEx on ClassElement {
     } catch (e, s) {
       // Fallback. Not perfect, but better than just forwarding original error.
       throw RealmInvalidGenerationSourceError(
-        '$e \n $s',
+        '$e\n$s',
         todo: //
             'Unexpected error. Please open an issue on: '
             'https://github.com/realm/realm-dart',
--- a/generator/lib/src/class_element_ex.dart
+++ b/generator/lib/src/class_element_ex.dart
@@ -161,6 +161,9 @@ extension ClassElementEx on ClassElement {
             todo: 'Remove the @PrimaryKey annotation from the field or set the model type to a value different from ObjectType.embeddedObject.');
       }
 
+      // Computed fields go last. This is important for the schema generation.
+      mappedFields.sort((a, b) => a.isComputed ^ b.isComputed ? (a.isComputed ? 1 : -1) : -1);
+
       return RealmModelInfo(name, modelName, realmName, mappedFields, objectType);
     } on InvalidGenerationSourceError catch (_) {
       rethrow;
--- a/generator/lib/src/class_element_ex.dart
+++ b/generator/lib/src/class_element_ex.dart
@@ -152,7 +152,8 @@ extension ClassElementEx on ClassElement {
 
       final objectType = ObjectType.values[modelInfo.value.getField('type')!.getField('index')!.toIntValue()!];
 
-      final mappedFields = fields.realmInfo.toList();
+      // Computed fields go last. This is important for the schema generation.
+      final mappedFields = fields.realmInfo.toList()..sort((a, b) => a.isComputed ^ b.isComputed ? (a.isComputed ? 1 : -1) : -1);
 
       if (objectType == ObjectType.embeddedObject && mappedFields.any((field) => field.isPrimaryKey)) {
         final pkSpan = fields.firstWhere((field) => field.realmInfo?.isPrimaryKey == true).span;
@@ -163,9 +164,6 @@ extension ClassElementEx on ClassElement {
             primaryLabel: "$realmName is marked as embedded but has primary key defined",
             todo: 'Remove the @PrimaryKey annotation from the field or set the model type to a value different from ObjectType.embeddedObject.');
       }
-
-      // Computed fields go last. This is important for the schema generation.
-      mappedFields.sort((a, b) => a.isComputed ^ b.isComputed ? (a.isComputed ? 1 : -1) : -1);
 
       return RealmModelInfo(name, modelName, realmName, mappedFields, objectType);
     } on InvalidGenerationSourceError catch (_) {
--- a/generator/lib/src/class_element_ex.dart
+++ b/generator/lib/src/class_element_ex.dart
@@ -152,7 +152,7 @@ extension ClassElementEx on ClassElement {
 
       final objectType = ObjectType.values[modelInfo.value.getField('type')!.getField('index')!.toIntValue()!];
 
-      // Computed fields go last. This is important for the schema generation.
+      // Realm Core requires computed properties at the end so we sort them at generation time versus doing it at runtime every time.
       final mappedFields = fields.realmInfo.toList()..sort((a, b) => a.isComputed ^ b.isComputed ? (a.isComputed ? 1 : -1) : -1);
 
       if (objectType == ObjectType.embeddedObject && mappedFields.any((field) => field.isPrimaryKey)) {
--- a/generator/lib/src/dart_type_ex.dart
+++ b/generator/lib/src/dart_type_ex.dart
@@ -61,11 +61,11 @@ extension DartTypeEx on DartType {
 
   DartType get mappedType {
     final self = this;
-    final provider = session.typeProvider;
     if (isRealmCollection) {
       if (self is ParameterizedType) {
         final mapped = self.typeArguments.last.mappedType;
         if (self != mapped) {
+          final provider = session.typeProvider;
           if (self.isDartCoreList) {
             final mappedList = provider.listType(mapped);
             return PseudoType('Realm${mappedList.getDisplayString(withNullability: true)}', nullabilitySuffix: mappedList.nullabilitySuffix);
--- a/generator/lib/src/dart_type_ex.dart
+++ b/generator/lib/src/dart_type_ex.dart
@@ -31,12 +31,12 @@ extension DartTypeEx on DartType {
   bool isExactly<T>() => TypeChecker.fromRuntime(T).isExactlyType(this);
 
   bool get isRealmAny => const TypeChecker.fromRuntime(RealmAny).isAssignableFromType(this);
-  bool get isRealmBacklink => isDartCoreIterable; // && element2 != null ? backlinkChecker.annotationsOfExact(element2!).isNotEmpty : false;
   bool get isRealmCollection => realmCollectionType != RealmCollectionType.none;
-  bool get isRealmModel => element != null ? realmModelChecker.annotationsOfExact(element!).isNotEmpty : false;
+  bool get isRealmModel => element2 != null ? realmModelChecker.annotationsOfExact(element2!).isNotEmpty : false;
 
   bool get isNullable => session.typeSystem.isNullable(this);
   DartType get asNonNullable => session.typeSystem.promoteToNonNull(this);
+  DartType get asNullable => session.typeSystem.leastUpperBound(this, session.typeProvider.nullType);
 
   RealmCollectionType get realmCollectionType {
     if (isDartCoreSet) return RealmCollectionType.set;
@@ -51,7 +51,7 @@ extension DartTypeEx on DartType {
 
   DartType get basicType {
     final self = this;
-    if (self is ParameterizedType && (isRealmCollection || isRealmBacklink)) {
+    if (self is ParameterizedType && (isRealmCollection || isDartCoreIterable)) {
       return self.typeArguments.last;
     }
     return this;
@@ -80,7 +80,7 @@ extension DartTypeEx on DartType {
           }
         }
       }
-    } else if (isRealmBacklink) {
+    } else if (isDartCoreIterable) {
       if (self is ParameterizedType) {
         final mapped = self.typeArguments.last.mappedType;
         if (self != mapped) {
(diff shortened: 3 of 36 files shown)


generator/lib/src/class_element_ex.dart
```diff
@@ -161,6 +161,9 @@ extension ClassElementEx on ClassElement {
             todo: 'Remove the @PrimaryKey annotation from the field or set the model type to a value different from ObjectType.embeddedObject.');
       }
 
+      // Computed fields go last. This is important for the schema generation.
+      mappedFields.sort((a, b) => a.isComputed ^ b.isComputed ? (a.isComputed ? 1 : -1) : -1);
```
nirinchev, line 165: That's a bit too much math. Isn't easier to use `sortedby`? Like, on line 152, we could do something like:

```dart
      final mappedFields = fields.realmInfo.sortedBy<num>((f) => f.isComputed ? 0 : 1);
```

test/backlinks_test.dart
```diff
@@ -0,0 +1,139 @@
...
+class _Source {
+  String name = 'source';
+  _Target? oneTarget;
+  List<_Target> manyTargets = [];
+}
+
+@RealmModel()
+class _Target {
+  @Backlink(#oneTarget)
+  late Iterable<_Source> oneToMany; // computed property, so must go last in generated class!
```
nirinchev, line 37: Why not ask people to declare backlinks as `RealmResults<_Source>`? We're using results under the hood, so don't see much value in those being iterables in the model.

generator/lib/src/class_element_ex.dart
nielsenko: We would need to pull in `import 'package:collection/collection.dart` and I dislike the need to add `<num>` bit to `sortedBy` (`<int>` won't do).

Also `List.sort` is in-place, but apparently not stable 🤔  ... nor is `sortedBy`

test/backlinks_test.dart
nielsenko: You could say the same for `List`. 

However, they are generated as `RealmResults` on `Target`, so you have the full interface available, ie `target.oneToMany[0]`, `target.manyToMany.changes.listen(...`,  etc.

I think it match `List` and `Set` better. 

I did consider [UnmodifiableListView](https://api.dart.dev/stable/2.18.3/dart-collection/UnmodifiableListView/UnmodifiableListView.html), and can be persuaded to switch, but it is not perfect either.

nirinchev: Aren't we using RealmList for the models? Or do you mean that we allow people to declare them either as List or RealmList?

nielsenko: No. In the realm models we are using `List<T>`

nielsenko: But in the generated classes we also have realm specific stuff like `freeze()`,  `changes`, etc. So the generated class overrides with `RealmList`. The same goes for back-links, they become `RealmResults<T>`.

generator/test/test_util.dart
```diff
@@ -70,6 +65,11 @@ class LinesEqualsMatcher extends Matcher {
       }
     }
 
+    if (actualLines.length > expectedLines.length) {
```
blagoev, line 68: Why not check for equality. Now this will report if actual is more than expected but not if expected is more than actual.

test/backlinks_test.dart
```diff
@@ -0,0 +1,140 @@
...
+
+  String name = 'target';
+
+  @Backlink(#manyTargets)
+  late Iterable<_Source> manyToMany; // computed property, so must go last in generated class!
```
blagoev, line 43: why do we have this comment here

common/lib/src/realm_common_base.dart
```diff
@@ -97,3 +97,10 @@ class Indexed {
 class Ignored {
   const Ignored();
 }
+
+/// Indicates a backlink property.
+/// {@category Annotations}
+class Backlink {
+  final Symbol symbol;
```
blagoev, line 104: no API doc

blagoev (changes requested):
Its good. I have small suggestions.

generator/lib/src/dart_type_ex.dart
```diff
@@ -49,19 +50,20 @@ extension DartTypeEx on DartType {
   DartType? get nullIfDynamic => isDynamic ? null : this;
 
   DartType get basicType {
-    if (isRealmCollection) {
-      return (this as ParameterizedType).typeArguments.last;
+    final self = this;
```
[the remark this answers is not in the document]
nielsenko: It helps with type inference later on line 54 & 55. Otherwise I would have to cast `this` on line 55 to get to `typeArguments`.

One can hope that the dart compiler becomes a bit smarter one day.

generator/lib/src/dart_type_ex.dart
```diff
@@ -49,19 +50,20 @@ extension DartTypeEx on DartType {
...
+      return self.typeArguments.last;
     }
-    return asNonNullable;
+    return this;
   }
 
   String get basicMappedName => basicType.mappedName;
 
   DartType get mappedType {
     final self = this;
```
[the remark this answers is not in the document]
nielsenko: See previous reply

test/backlinks_test.dart
nielsenko: @blagoev The comment is their, because it is important that the generator re-orders these fields, so that the generated props go last. That is a requirement before handing the props over to core.

test/backlinks_test.dart
nielsenko: see previous reply

generator/lib/src/dart_type_ex.dart
```diff
@@ -49,19 +50,20 @@ extension DartTypeEx on DartType {
...
+    final provider = session.typeProvider;
```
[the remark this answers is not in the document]
nielsenko: Yes.. that is some leftover from a larger change I tried. Will move back.

test/backlinks_test.dart
nirinchev: Yeah, this is slightly different from the C# SDKs behavior, but I think that's fine. I got confused because I thought we're declaring properties as `RealmList`, but in reality we're doing it with `List`. Fine with resolving this on my end.

blagoev: Sorry for hijacking the thread here. 

I suggest we remove these comments from the tests. These serve no good in the tests here. I don't see how they help if I read the code. Should the next one that refactors do something about it.

common/lib/src/realm_common_base.dart
```diff
@@ -97,3 +97,11 @@ class Indexed {
 class Ignored {
   const Ignored();
 }
+
+/// Indicates that the field it decorates is the inverse end of a relationship.
+/// {@category Annotations}
+class Backlink {
+  /// The name of the field in the other class that links to this class.
```
blagoev, line 104: .NET has this as a doc `The property that is on the other end of the relationship.` Should we use
```suggestion
  /// The name of the field that is on the other end of the relationship.
```

nielsenko: I kind of think my sentence is a bit more precise? Also the suggestion is almost identical to the comment in line 101

test/backlinks_test.dart
nielsenko: What if I change to:
```dart
@RealmModel()
class _Target {
  @Backlink(#oneTarget)
  late Iterable<_Source> oneToMany; // will be reordered by the generator to go after name

  String name = 'target';

  @Backlink(#manyTargets)
  late Iterable<_Source> manyToMany;
}
```
?

generator/test/test_util.dart
nirinchev: That case is already handled by the for loop above. If expected has more lines than actual, we would end up in the if-clause on line 58.

test/backlinks_test.dart
blagoev: Actually I would suggest we change the generator comment to be more clear. Something like `Realm Core requires computed properties at the end so we sort them at generation time versus doing it at runtime every time`.

generator/lib/src/class_element_ex.dart
```diff
@@ -149,7 +152,8 @@ extension ClassElementEx on ClassElement {
 
       final objectType = ObjectType.values[modelInfo.value.getField('type')!.getField('index')!.toIntValue()!];
 
-      final mappedFields = fields.realmInfo.toList();
+      // Computed fields go last. This is important for the schema generation.
```
blagoev, line 155:
```suggestion
      // Realm Core requires computed properties at the end so we sort them at generation time versus doing it at runtime every time. 
```

blagoev: @nielsenko added this one for the generator. Imo, we should be more clear why we are doing the sort there

lib/src/realm_object.dart
```diff
@@ -142,24 +144,31 @@ class RealmCoreAccessor implements RealmAccessor {
...
-            case ObjectType.embeddedObject:
-              return object.realm.createList<EmbeddedObject>(handle, listMetadata);
-            default:
-              throw RealmError('List of ${listMetadata.schema.baseType} is not supported yet');
+        if (propertyMeta.propertyType == RealmPropertyType.linkingObjects) {
+          final sourceMeta = object.realm.metadata.getByName(propertyMeta.objectType!);
+          final sourceProperty = sourceMeta[propertyMeta.linkOriginProperty!];
+          final handle = realmCore.getBacklinks(object, sourceMeta.classKey, sourceProperty.key);
+          return RealmResultsInternal.create<T>(handle, object.realm, metadata);
+        } else {
```
nirinchev, line 152: Don't need this `else`.

test/backlinks_test.dart
nielsenko: I have done the comment change in the generator. I have left a comment here in case someone looks at the generated class, and wonder why the fields are re-ordered.

generator/test/test_util.dart
blagoev: why is this resolved

test/backlinks_test.dart
nielsenko: I have removed this

generator/test/test_util.dart
nielsenko: On my end it looks like:
```dart
if (actualLines.length != expectedLines.length) { ...
```

test/backlinks_test.dart
```diff
@@ -0,0 +1,140 @@
...
+  String name = 'source';
+  @MapTo('et mål') // to throw a curve ball..
+  _Target? oneTarget;
+  List<_Target> manyTargets = [];
+}
+
+@RealmModel()
+class _Target {
+  @Backlink(#oneTarget)
+  late Iterable<_Source> oneToMany; // will be reordered by the generator to go after name
```
blagoev, line 38:
```suggestion
  late Iterable<_Source> oneToMany;
```

generator/test/test_util.dart
nielsenko: @nirinchev I agree, but there is no harm in changing it to `!=`.. it reads a bit easier.

test/backlinks_test.dart
blagoev: yeah but why would they care. I don't remember one time I needed to check the generated class for the order. and there is the comment in the generator

nielsenko: I don't think users should care. It is more for the future me..

generator/lib/src/class_element_ex.dart
```diff
@@ -149,7 +152,8 @@ extension ClassElementEx on ClassElement {
 
       final objectType = ObjectType.values[modelInfo.value.getField('type')!.getField('index')!.toIntValue()!];
 
-      final mappedFields = fields.realmInfo.toList();
+      // Realm Core requires computed properties at the end so we sort them at generation time versus doing it at runtime every time.
+      final mappedFields = fields.realmInfo.toList()..sort((a, b) => a.isComputed ^ b.isComputed ? (a.isComputed ? 1 : -1) : -1);
```
nirinchev, line 156: I'm still not sure this sorting is super readable. How about:

```suggestion
      final mappedFields = fields.realmInfo.toList()..sort((a, b) => a.isComputed == b.isComputed ? 0 : (a.isComputed ? 1 : -1));
```

Or even simpler:

```suggestion
      final mappedFields = fields.realmInfo.toList()..sort((a, b) => a.isComputed ? 1 : -1);
```

nirinchev: Not a dealbreaker though - it should still be correct, just feels a little complicated to understand.

nirinchev (approved):

generator/lib/src/class_element_ex.dart
nielsenko: I like to read it like: If a and b are different, then if a is computed, a goes last, otherwise a goes first.

blagoev (approved):

Taco55:
After trying this new feature I get the following error after running the generator:

```
[SEVERE] realm:realm_generator on lib/schemas/category.dart:

Invalid argument(s): Element (FieldElementImpl) List<$Category> categories is not defined in this library.
#0      ResolvedLibraryResultImpl.getElementDeclaration.<anonymous closure> (package:analyzer/src/dart/analysis/results.dart:251:9)
#1      ListMixin.firstWhere (dart:collection/list.dart:166:38)
#2      ResolvedLibraryResultImpl.getElementDeclaration (package:analyzer/src/dart/analysis/results.dart:247:28)
#3      getDeclarationFromElement (package:realm_generator/src/element.dart:38:34)
#4      FieldElementEx.declarationAstNode (package:realm_generator/src/field_element_ex.dart:40:46)
#5      ElementEx.declarationAstNode (package:realm_generator/src/element.dart:72:43)
#6      ElementEx._annotationsInfoOfExact (package:realm_generator/src/element.dart:79:18)
#7      _SyncStarIterator.moveNext (dart:async-patch/async_patch.dart:710:21)
#8      new _GrowableList._ofOther (dart:core-patch/growable_array.dart:202:26)
#9      new _GrowableList.of (dart:core-patch/growable_array.dart:152:26)
#10     new List.of (dart:core-patch/array_patch.dart:51:28)
#11     Iterable.toList (dart:core/iterable.dart:470:12)
#12     ElementEx.annotationInfoOfExact (package:realm_generator/src/element.dart:89:58)
#13     FieldElementEx.realmInfo (package:realm_generator/src/field_element_ex.dart:250:44)
#14     _extension#0.realmInfo (package:realm_generator/src/class_element_ex.dart:37:22)
#15     _SyncStarIterator.moveNext (dart:async-patch/async_patch.dart:710:21)
#16     new _GrowableList._ofOther (dart:core-patch/growable_array.dart:202:26)
#17     new _GrowableList.of (dart:core-patch/growable_array.dart:152:26)
#18     new List.of (dart:core-patch/array_patch.dart:51:28)
#19     Iterable.toList (dart:core/iterable.dart:470:12)
#20     ClassElementEx.realmInfo (package:realm_generator/src/class_element_ex.dart:156:45)
#21     _extension#0.realmInfo.<anonymous closure> (package:realm_generator/src/realm_object_generator.dart:71:58)
#22     MappedIterator.moveNext (dart:_internal/iterable.dart:391:20)
#23     WhereIterator.moveNext (dart:_internal/iterable.dart:438:22)
#24     CastIterator.moveNext (dart:_internal/cast.dart:61:30)
#25     ExpandIterator.moveNext (dart:_internal/iterable.dart:477:21)
#26     Iterable.join (dart:core/iterable.dart:423:19)
#27     RealmObjectGenerator.generate.<anonymous closure>.<anonymous closure> (package:realm_generator/src/realm_object_generator.dart:61:69)
#28     _rootRun (dart:async/zone.dart:1391:13)
#29     _CustomZone.run (dart:async/zone.dart:1293:19)
#30     _runZoned (dart:async/zone.dart:1829:10)
#31     runZonedGuarded (dart:async/zone.dart:1817:12)
#32     scopeSession (package:realm_generator/src/session.dart:44:16)
#33     RealmObjectGenerator.generate.<anonymous closure> (package:realm_generator/src/realm_object_generator.dart:59:16)
<asynchronous suspension>
#34     measure.<anonymous closure> (package:realm_generator/src/measure.dart:50:18)
<asynchronous suspension>
#35     measure (package:realm_generator/src/measure.dart:47:7)
<asynchronous suspension>
#36     RealmObjectGenerator.generate (package:realm_generator/src/realm_object_generator.dart:56:12)
<asynchronous suspension>
#37     _generate (package:source_gen/src/builder.dart:352:23)
<asynchronous suspension>
#38     Stream.toList.<anonymous closure> (dart:async/stream.dart:1349:9)
<asynchronous suspension>


in: package:stockit/schemas/category.dart:33:33
    ╷
7   │ @RealmModel()
8   │ class $Category {
    │       ━━━━━━━━━ in realm model for 'Category'
... │
32  │   @Backlink(#categories)
33  │   late Iterable<$Item> items;
    │                                 ^^^^^^^^^^^^^^ !
    ╵
Unexpected error. Please open an issue on: https://github.com/realm/realm-dart
```

I get the same error for Many-to-one and many-to-many relationships. Still anything wrong?

nielsenko:
@Taco55 This is because the backlink is cross file. It was first reported here #1015. A fix is already merged, but not released yet. You can work around the issue by ensuring that the class with the backlink field is defined in the same file as the linking class.

Sorry for the inconvenience.

Taco55:
@nielsenko Thanks for the quick answer. I will check it out!

PR #998, opened 2022-10-28
Add timeouts for CI jobs

nirinchev:
These timeouts should be much higher than what we need, so if we start running into them, we should investigate.


nielsenko (approved):
These timeouts seems reasonable to me.

PR #999, opened 2022-10-28
Set the correct schema mode for disconnectedSync

nirinchev:
Noticed that while looking at https://docs.google.com/document/d/17cqTiriugPEJQjvk58t7kwcr9Cdkat5Le7khJOZju5k/edit#. Need to write some tests for it though.


blagoev (approved):

nielsenko (approved):
Wasn't aware of this flag 👍

PR #1003, opened 2022-11-02
Refactor client reset

blagoev:
refactor client reset

### DIFF
--- a/lib/src/configuration.dart
+++ b/lib/src/configuration.dart
@@ -429,18 +429,22 @@ extension SchemaObjectInternal on SchemaObject {
 /// This can happen if the server is rolled back or restored from backup.
 /// {@category Sync}
 abstract class ClientResetHandler {
-  /// Defines what should happen in case of a Client reset
-  ClientResyncModeInternal get _mode;
+  
+  // Defines what should happen in case of a Client reset
+  final ClientResyncModeInternal _mode;
 
-  BeforeResetCallback? get _onBeforeReset => null;
-  AfterResetCallback? get _onAfterDiscard => null;
-  AfterResetCallback? get _onAfterRecovery => null;
+  final BeforeResetCallback? _onBeforeReset;
+  final AfterResetCallback? _onAfterDiscard;
+  final AfterResetCallback? _onAfterRecovery;
 
   /// The callback that handles the [ClientResetError].
   final ClientResetCallback? onManualReset;
 
   /// Initializes a new instance of [ClientResetHandler].
-  const ClientResetHandler(this.onManualReset);
+  const ClientResetHandler._(this.onManualReset,  this._mode, {BeforeResetCallback? onBeforeReset, AfterResetCallback? onAfterDiscard, AfterResetCallback? onAfterRecovery})
+      : _onBeforeReset = onBeforeReset,
+        _onAfterDiscard = onAfterDiscard,
+        _onAfterRecovery = onAfterRecovery;
 }
 
 /// A client reset strategy where the user needs to fully take care of a client reset.
@@ -449,14 +453,10 @@ abstract class ClientResetHandler {
 /// that will enable full control of moving any unsynced changes to the synchronized `Realm`.
 /// {@category Sync}
 class ManualRecoveryHandler extends ClientResetHandler {
-
   /// Creates an instance of `ManualRecoveryHandler` with the supplied client reset handler.
   ///
   /// [onReset] callback is triggered when a manual client reset happens.
-  const ManualRecoveryHandler(ClientResetCallback onReset) : super(onReset);
-
-  @override
-  ClientResyncModeInternal get _mode => ClientResyncModeInternal.manual;
+  const ManualRecoveryHandler(ClientResetCallback onReset) : super._(onReset, ClientResyncModeInternal.manual);
 }
 
 /// A client reset strategy where any not yet synchronized data is automatically
@@ -468,10 +468,10 @@ class ManualRecoveryHandler extends ClientResetHandler {
 /// {@category Sync}
 class DiscardUnsyncedChangesHandler extends ClientResetHandler {
   /// The callback that will be executed just before the client reset happens.
-  final BeforeResetCallback? onBeforeReset;
+  BeforeResetCallback? get onBeforeReset => _onBeforeReset;
 
   /// The callback that will be executed just after the client reset happens.
-  final AfterResetCallback? onAfterReset;
+  AfterResetCallback? get onAfterReset => _onAfterDiscard;
 
   /// Creates an instance of `DiscardUnsyncedChangesHandler`.
   ///
@@ -479,16 +479,8 @@ class DiscardUnsyncedChangesHandler extends ClientResetHandler {
   /// The first two are invoked just before and after the client reset has happened,
   /// while the last one will be invoked in case an error occurs during the automated process and the system needs to fallback to a manual mode.
   /// The freshly downloaded copy of the synchronized Realm triggers all change notifications as a write transaction is internally simulated.
-  const DiscardUnsyncedChangesHandler({this.onBeforeReset, this.onAfterReset, ClientResetCallback? onManualResetFallback}) : super(onManualResetFallback);
-
-  @override
-  ClientResyncModeInternal get _mode => ClientResyncModeInternal.discardLocal;
-
-  @override
-  BeforeResetCallback? get _onBeforeReset => onBeforeReset;
-
-  @override
-  AfterResetCallback? get _onAfterDiscard => onAfterReset;
+  const DiscardUnsyncedChangesHandler({BeforeResetCallback? onBeforeReset, AfterResetCallback? onAfterReset, ClientResetCallback? onManualResetFallback})
+      : super._(onManualResetFallback, ClientResyncModeInternal.discardLocal, onAfterDiscard: onAfterReset, onBeforeReset: onBeforeReset);
 }
 
 /// A client reset strategy that attempts to automatically recover any unsynchronized changes.
@@ -499,26 +491,18 @@ class DiscardUnsyncedChangesHandler extends ClientResetHandler {
 /// {@category Sync}
 class RecoverUnsyncedChangesHandler extends ClientResetHandler {
   /// The callback that will be executed just before the client reset happens.
-  final BeforeResetCallback? onBeforeReset;
+  BeforeResetCallback? get onBeforeReset => _onBeforeReset;
 
   /// The callback that will be executed just after the client reset happens.
-  final AfterResetCallback? onAfterReset;
+  AfterResetCallback? get onAfterReset => _onAfterRecovery;
 
   /// Creates an instance of `RecoverUnsyncedChangesHandler`.
   ///
   /// This strategy supplies three callbacks: [onBeforeReset], [onAfterReset] and [onManualResetFallback].
   /// The first two are invoked just before and after the client reset has happened, while the last one is invoked
   /// in case an error occurs during the automated process and the system needs to fallback to a manual mode.
-  const RecoverUnsyncedChangesHandler({this.onBeforeReset, this.onAfterReset, ClientResetCallback? onManualResetFallback}) : super(onManualResetFallback);
-
-  @override
-  ClientResyncModeInternal get _mode => ClientResyncModeInternal.recover;
-
-  @override
-  BeforeResetCallback? get _onBeforeReset => onBeforeReset;
-
-  @override
-  AfterResetCallback? get _onAfterRecovery => onAfterReset;
+  const RecoverUnsyncedChangesHandler({BeforeResetCallback? onBeforeReset, AfterResetCallback? onAfterReset, ClientResetCallback? onManualResetFallback})
+      : super._(onManualResetFallback, ClientResyncModeInternal.recover, onBeforeReset: onBeforeReset, onAfterRecovery: onAfterReset);
 }
 
 /// A client reset strategy that attempts to automatically recover any unsynchronized changes.
@@ -532,15 +516,15 @@ class RecoverUnsyncedChangesHandler extends ClientResetHandler {
 /// {@category Sync}
 class RecoverOrDiscardUnsyncedChangesHandler extends ClientResetHandler {
   /// The callback that will be executed just before the client reset happens.
-  final BeforeResetCallback? onBeforeReset;
-
-  /// The callback that will be executed just after the client reset happens if the local changes
-  /// were successfully recovered.
-  final AfterResetCallback? onAfterRecovery;
+  BeforeResetCallback? get onBeforeReset => _onBeforeReset;
 
   /// The callback that will be executed just after the client reset happens if the local changes
   /// needed to be discarded.
-  final AfterResetCallback? onAfterDiscard;
+  AfterResetCallback? get onAfterDiscard => _onAfterDiscard;
+
+  /// The callback that will be executed just after the client reset happens if the local changes
+  /// were successfully recovered.
+  AfterResetCallback? get onAfterRecovery => _onAfterRecovery;
 
   /// Creates an instance of `RecoverOrDiscardUnsyncedChangesHandler`.
   ///
@@ -554,20 +538,9 @@ class RecoverOrDiscardUnsyncedChangesHandler extends ClientResetHandler {
   /// [onAfterDiscard] is invoked if and only if an automatic client reset failed and instead the discard unsynced one succeded.
   /// The callback is never called if the discard unsynced client reset fails.
   /// [onManualResetFallback] is invoked whenever an error occurs in either of the recovery stragegies and the system needs to fallback to a manual mode.
-  const RecoverOrDiscardUnsyncedChangesHandler({this.onBeforeReset, this.onAfterRecovery, this.onAfterDiscard, ClientResetCallback? onManualResetFallback})
-      : super(onManualResetFallback);
-
-  @override
-  ClientResyncModeInternal get _mode => ClientResyncModeInternal.recoverOrDiscard;
-
-  @override
-  BeforeResetCallback? get _onBeforeReset => onBeforeReset;
-
-  @override
-  AfterResetCallback? get _onAfterDiscard => onAfterDiscard;
-
-  @override
-  AfterResetCallback? get _onAfterRecovery => onAfterRecovery;
+  const RecoverOrDiscardUnsyncedChangesHandler(
+      {BeforeResetCallback? onBeforeReset, AfterResetCallback? onAfterRecovery, AfterResetCallback? onAfterDiscard, ClientResetCallback? onManualResetFallback})
+      : super._(onManualResetFallback, ClientResyncModeInternal.recoverOrDiscard, onBeforeReset: onBeforeReset, onAfterDiscard: onAfterDiscard, onAfterRecovery: onAfterRecovery);
 }
 
 /// @nodoc
--- a/lib/src/configuration.dart
+++ b/lib/src/configuration.dart
@@ -441,7 +441,7 @@ abstract class ClientResetHandler {
   final ClientResetCallback? onManualReset;
 
   /// Initializes a new instance of [ClientResetHandler].
-  const ClientResetHandler._(this.onManualReset,  this._mode, {BeforeResetCallback? onBeforeReset, AfterResetCallback? onAfterDiscard, AfterResetCallback? onAfterRecovery})
+  const ClientResetHandler._(this._mode, this.onManualReset,  {BeforeResetCallback? onBeforeReset, AfterResetCallback? onAfterDiscard, AfterResetCallback? onAfterRecovery})
       : _onBeforeReset = onBeforeReset,
         _onAfterDiscard = onAfterDiscard,
         _onAfterRecovery = onAfterRecovery;
@@ -456,7 +456,7 @@ class ManualRecoveryHandler extends ClientResetHandler {
   /// Creates an instance of `ManualRecoveryHandler` with the supplied client reset handler.
   ///
   /// [onReset] callback is triggered when a manual client reset happens.
-  const ManualRecoveryHandler(ClientResetCallback onReset) : super._(onReset, ClientResyncModeInternal.manual);
+  const ManualRecoveryHandler(ClientResetCallback onReset) : super._(ClientResyncModeInternal.manual, onReset);
 }
 
 /// A client reset strategy where any not yet synchronized data is automatically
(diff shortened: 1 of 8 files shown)


lib/src/native/realm_core.dart
```diff
@@ -500,52 +502,45 @@ class _RealmCore {
...
+      _realmLib.realm_register_user_code_callback_error(error.toPersistentHandle());
+    } finally {
+      _realmLib.realm_dart_invoke_unlock_callback(success, unlockCallbackFunc);
     }
   }
 
   static void _syncBeforeResetCallback(Object userdata, Pointer<shared_realm> realmHandle, Pointer<Void> unlockCallbackFunc) {
     final syncConfig = userdata as FlexibleSyncConfiguration;
-    final beforeResetCallback = syncConfig.clientResetHandler.onBeforeReset;
-    if (beforeResetCallback == null) {
```
blagoev, line 511: The callback is only registered if the dart callback was provided. So if it is invoked then we are sure this can't be null.

lib/src/native/realm_core.dart
```diff
@@ -226,13 +226,15 @@ class _RealmCore {
           _realmLib.realm_sync_config_set_error_handler(syncConfigPtr, _realmLib.addresses.realm_dart_sync_error_handler_callback, errorHandlerUserdata.cast(),
               _realmLib.addresses.realm_dart_userdata_async_free);
 
-          if (config.clientResetHandler is! ManualRecoveryHandler) {
```
blagoev, line 229: We don't switch on the type but rather check if the callback has been provided

src/realm_dart_sync.cpp
```diff
@@ -139,56 +139,46 @@ RLM_API void realm_dart_sync_on_subscription_state_changed_callback(realm_userda
...
+bool invoke_dart_and_await_result(realm::util::UniqueFunction<void(realm::util::UniqueFunction<void(bool)>*)>* userCallback)
 {
     std::condition_variable condition;
     std::mutex mutex;
-    bool complete = false;
     bool success = false;
+    bool completed = false;
 
     realm::util::UniqueFunction<void(bool)> unlockFunc = [&](bool result) {
         std::unique_lock<std::mutex> lock(mutex);
```
nielsenko, line 150:
```suggestion
    realm::util::UniqueFunction unlockFunc = [&](bool result) {
        std::unique_lock lock(mutex);
```
compiler is smart enough to infer this

src/realm_dart_sync.cpp
```diff
@@ -139,56 +139,46 @@ RLM_API void realm_dart_sync_on_subscription_state_changed_callback(realm_userda
...
-        complete = true;
         success = result;
+        completed = true;
         lock.unlock();
         condition.notify_one();
     };
     (*userCallback)(&unlockFunc);
 
     std::unique_lock<std::mutex> lock(mutex);
```
nielsenko, line 158:
```suggestion
    std::unique_lock lock(mutex);
```
let compiler infer

lib/src/configuration.dart
```diff
@@ -429,18 +429,22 @@ extension SchemaObjectInternal on SchemaObject {
...
+  
+  // Defines what should happen in case of a Client reset
+  final ClientResyncModeInternal _mode;
 
-  BeforeResetCallback? get _onBeforeReset => null;
-  AfterResetCallback? get _onAfterDiscard => null;
-  AfterResetCallback? get _onAfterRecovery => null;
+  final BeforeResetCallback? _onBeforeReset;
+  final AfterResetCallback? _onAfterDiscard;
+  final AfterResetCallback? _onAfterRecovery;
```
nielsenko, line 438: I think I prefer this as well, although it can be argued it allocates a bit more memory

nielsenko (approved):
Have a few suggestions in the c++ part, but otherwise fine with me

src/realm_dart_sync.cpp
blagoev: that's not something this PR is changing but I prefer your suggestion. will commit

lib/src/configuration.dart
blagoev: yes, but since there will be a limited number of instances of these classes throughout the app lifetime its should be negligible.

lib/src/native/realm_core.dart
```diff
@@ -500,52 +502,45 @@ class _RealmCore {
...
     final realm = RealmInternal.getUnowned(syncConfig, RealmHandle._unowned(realmHandle));
-    _continueWhenComplete(() => beforeResetCallback(realm), unlockCallbackFunc);
+
+    _callResetCallback(() => beforeResetCallback(realm), unlockCallbackFunc);
   }
 
   static void _syncAfterResetCallback(Object userdata, Pointer<shared_realm> beforeHandle, Pointer<realm_thread_safe_reference> afterReference, bool didRecover,
       Pointer<Void> unlockCallbackFunc) {
     final syncConfig = userdata as FlexibleSyncConfiguration;
-    final afterResetCallback = didRecover ? syncConfig.clientResetHandler.onAfterRecovery : syncConfig.clientResetHandler.onAfterDiscard;
```
nirinchev, line 522: Why is this check removed? You could still end up with a null callback if you do something like:


```dart
final handler = RecoveryOrDiscardHandler(onAfterRecovery: () {...});
```

If recovery fails, the `onAfterDiscard` needs to be invoked, but it'll be `null` here.

src/realm_dart.cpp
```diff
@@ -87,5 +87,11 @@ RLM_API void realm_dart_userdata_async_free(void* userdata)
     });
 }
 
+RLM_API void realm_dart_invoke_unlock_callback(bool success, void* unlockFunc)
+{
+    auto castFunc = (reinterpret_cast<realm::util::UniqueFunction<void(bool)>*>(unlockFunc));
+    (*castFunc)(success);
+}
```
nirinchev, line 94: I dislike that this is now so specific. There's nothing in the function implementation that indicates that this is somehow related to locking. It's just invoking a function that takes a boolean. The fact that in the client reset flow it'll unlock a lock is an implementation detail.

nirinchev (changes requested):

src/realm_dart.cpp
blagoev: This exactly a specific function to unlock and the argument is called unlockFunc, hence the otherway around was weird. This makes it more clear what it does and when it should be used.

lib/src/native/realm_core.dart
blagoev: which check was removed?

src/realm_dart.cpp
desistefanova: I think we can renamed only the argument name. I has left unchanged after I refactored the function to be general.

lib/src/native/realm_core.dart
nirinchev: The check for null

src/realm_dart_sync.cpp
```diff
@@ -139,56 +139,46 @@ RLM_API void realm_dart_sync_on_subscription_state_changed_callback(realm_userda
     });
 }
 
-bool wait_dart_before_continue(realm::util::UniqueFunction<void(realm::util::UniqueFunction<void(bool)>*)>* userCallback)
+bool invoke_dart_and_await_result(realm::util::UniqueFunction<void(realm::util::UniqueFunction<void(bool)>*)>* userCallback)
```
desistefanova, line 142: I prefer more `wait_dart_before_continue` since when you see the place where it is called it is clear that you stops the execution there and you wait something. Could be `wait_dart_result_then_continue`. But starting with `wait` makes it more obvious that you are waiting.

src/realm_dart.cpp
nirinchev: I also think we should rename the argument - it's a void* which is then cast to `function(bool)`. Nothing about it deals with unlocking anything.

src/realm_dart_sync.cpp
blagoev: There is a `wait` in the function name. Its just more natural language. The previous function name was not natural to me.

lib/src/native/realm_core.dart
```diff
@@ -226,13 +226,15 @@ class _RealmCore {
           _realmLib.realm_sync_config_set_error_handler(syncConfigPtr, _realmLib.addresses.realm_dart_sync_error_handler_callback, errorHandlerUserdata.cast(),
               _realmLib.addresses.realm_dart_userdata_async_free);
 
-          if (config.clientResetHandler is! ManualRecoveryHandler) {
+          if (config.clientResetHandler.onBeforeReset != null) {
```
desistefanova, line 229: OK! You have changed this, so I don't need to make additional PR, as you suggested in Client Reset PR.

blagoev: yes no need

src/realm_dart.cpp
desistefanova: It just call a function. You specify the pointer of that function as an argument. The method doesn't care what is the function just calls it. The specific this here is only that the function expects a bool argument. It could be renamed to `realm_dart_invoke_function_with_confirmation`

blagoev: the argument just names what the void* pointer points to. The name of the function names what it does.

lib/src/native/realm_core.dart
blagoev: the null check below 
```dart
if (afterResetCallback == null) {
```

was removed cause this funciton `_syncAfterResetCallback` now is only registered and invoked if either `onAfterRecovery` or `onAfterDiscard` is not null. 
hence the new implementation does a hard cast from null `onAfterRecovery!` and `onAfterDiscard!`

Is the question about these hard casts? Are they problematic?

lib/src/native/realm_core.dart
desistefanova: OK! You have changed this, so I don't need to make additional PR, as you suggested in Client Reset PR.

lib/src/native/realm_core.dart
blagoev: This check now is done here
https://github.com/realm/realm-dart/pull/1003/files#diff-cd0d6010ea23a618f899185aec525b9c186cc41c35679f48d21f2d295f5c8cfbR237

src/realm_dart.cpp
nirinchev: The `void*` pointer points to a `Function(bool)`. How is that related to locks?

lib/src/native/realm_core.dart
```diff
@@ -500,52 +502,45 @@ class _RealmCore {
...
   }
 
   static void _syncBeforeResetCallback(Object userdata, Pointer<shared_realm> realmHandle, Pointer<Void> unlockCallbackFunc) {
     final syncConfig = userdata as FlexibleSyncConfiguration;
-    final beforeResetCallback = syncConfig.clientResetHandler.onBeforeReset;
-    if (beforeResetCallback == null) {
-      _invokeNativeFunction(unlockCallbackFunc);
-      return;
-    }
+    final beforeResetCallback = syncConfig.clientResetHandler.onBeforeReset!;
```
desistefanova, line 525: I wouldn't rely on the implementation of another function to make sure that onBeforeReset is not null. We can change it after two version and to forget about that requirement. 
What bothers me is that if some error happens inside `_syncBeforeResetCallback` method we will never return and will hold C thread locked. It is good to have `try` block and in its `catch` to always unlock the C thread with `call_invokeNativeFunction(unlockCallbackFunc);`. An error could happen also when calling `RealmInternal.getUnowned`. It should be also in the `try`.
The same is for `_syncAfterResetCallback`

lib/src/native/realm_core.dart
nirinchev: As I said in my example - if `onRecoveryCallback` is not null, we're going to register: `config.clientResetHandler.onAfterRecovery != null || config.clientResetHandler.onAfterDiscard != null` returns `true`. But if `didRecover` is `false`, we'll try to execute `onAfterDiscard` which is `null`.

lib/src/native/realm_core.dart
```diff
@@ -500,52 +502,45 @@ class _RealmCore {
...
-  }
 
-  static Future<void> _continueWhenComplete(FutureOr<void> Function() callback, Pointer<Void> nativeFunction) async {
-    try {
-      await callback();
-      _invokeNativeFunction(nativeFunction);
-    } catch (error) {
-      _invokeNativeFunction(nativeFunction, error);
-    }
+     _callResetCallback(() => afterResetCallback(beforeRealm, afterRealm), unlockCallbackFunc);
```
desistefanova, line 543: Could we find some better name? There are too much uses of the name `ResetCallbacks`. It is not clear what we do in this method.

src/realm_dart.cpp
blagoev: it specifies that this function `realm_dart_invoke_unlock_callback` works with specific native callbacks expected to be passed as an argument.  You can pass any other function but that will be wrong usage and we can not express that other than the parameter name

lib/src/native/realm_core.dart
blagoev: `_callResetCallback` calls the reset call backs in a specific way and should not be used in any other place in the SDK. Do you have any suggestion for a better name?

lib/src/native/realm_core.dart
blagoev: That is exactly the idea and what it should do. This function  *demands* that this is not null if it is invoked. If it is it will throw an exception. That is the idea of the hard cast.

lib/src/native/realm_core.dart
```diff
@@ -500,52 +502,45 @@ class _RealmCore {
...
   static void _syncBeforeResetCallback(Object userdata, Pointer<shared_realm> realmHandle, Pointer<Void> unlockCallbackFunc) {
     final syncConfig = userdata as FlexibleSyncConfiguration;
-    final beforeResetCallback = syncConfig.clientResetHandler.onBeforeReset;
-    if (beforeResetCallback == null) {
-      _invokeNativeFunction(unlockCallbackFunc);
-      return;
-    }
+    final beforeResetCallback = syncConfig.clientResetHandler.onBeforeReset!;
+
     final realm = RealmInternal.getUnowned(syncConfig, RealmHandle._unowned(realmHandle));
```
nirinchev, line 527: This is not related to this PR, but we need to be releasing these handles at the end of the callback. At the end, this should look something like:

```dart
final beforeHandle = RealmHandle._unowned(realmHandle);
try {
  final realm = RealmInternal.getUnowned(...);
  await beforeResetCallback(realm);
  unlock(true);
} catch (e) {
  // log e
  unlock(false);
} finally {
  beforeHandle.release();
}
```

src/realm_dart.cpp
nirinchev: I disagree with this - it's a function that executes a native function. What the native function does is irrelevant as long as it takes a single boolean argument.

lib/src/native/realm_core.dart
nirinchev: As Desi points out, if it is null it will lead to a deadlock because we're doing this outside of the try block. Additionally, it's kind of dangerous if due to a bug in our SDK, users start losing data.

desistefanova: If you try to `throw Exception("test exception");` here, we will have a deadlock.

lib/src/native/realm_core.dart
blagoev: will fix

desistefanova: Maybe we can achieve it now with something like:
    _callResetCallback(() async { 
        final beforeHandle = RealmHandle._unowned(realmHandle);
        final realm = RealmInternal.getUnowned(syncConfig, RealmHandle._unowned(realmHandle));
        await beforeResetCallback(realm), unlockCallbackFunc).whenComplete(()=>beforeHandle.release());
        }, unlockCallbackFunc);

src/realm_dart.cpp
blagoev: We are not trying to implement a generic function that can call native functions with a bool argument. We are writing a very specific function that should do only one thing and is clear what it does and when it should be used. We don't have a case for anything else. 
If we are trying to make a generic function then the questions will be why this signature and not a different one. Why support bools and not even more. Then why don't we stop right now and make 100 functions with different generic meaning.
This is not useful since we don't have such case. Hence we should be explicit here when this function should be used.

lib/src/configuration.dart
```diff
@@ -429,18 +429,22 @@ extension SchemaObjectInternal on SchemaObject {
...
 
   /// The callback that handles the [ClientResetError].
   final ClientResetCallback? onManualReset;
 
   /// Initializes a new instance of [ClientResetHandler].
-  const ClientResetHandler(this.onManualReset);
+  const ClientResetHandler._(this.onManualReset,  this._mode, {BeforeResetCallback? onBeforeReset, AfterResetCallback? onAfterDiscard, AfterResetCallback? onAfterRecovery})
```
desistefanova, line 444:
```suggestion
const ClientResetHandler._(this._mode, this.onManualReset, {BeforeResetCallback? onBeforeReset, AfterResetCallback? onAfterDiscard, AfterResetCallback? onAfterRecovery})
```
I prefer to start with the mode and them to provide all the callBacks.

desistefanova: I'm OK with the rest of the other changes in the Configuration. The API doc also looks as good as before, so that we keep the user API the same.

src/realm_dart.cpp
nirinchev: I'm not sure I subscribe to this ideology. While I agree we're not trying to implement a general-purpose function calling mechanism, I firmly believe we should name functions for what they do and not for how they're being used. The name of the function as is implies that it somehow manages locking/unlocking of resources, but it doesn't do it - it just invokes another function with a boolean argument.

desistefanova: If it sounds to you better to be specific, why did you move it realm_dart.cpp? Let's specify it for reset callbacks and to leave it next to the locking functions in realm_dart_sync.cpp

blagoev: because it can be used for other invocations that expect a result synchronously as well. it is not tied to sync specifically.

blagoev: @nikola This function is named after what it does and not how it is used. It invokes a unlock function callback which was passed to a Dart function which accepts a unlock callback argument. This ties the whole concept together. So later when someone needs to do the same it is obvious what needs to be done. create a static dart function which should return a result using that function and continue the operation

lib/src/native/realm_core.dart
blagoev: .whenComplete is not avaialble since this is `FutureOr`

lib/src/native/realm_core.dart
blagoev: then the previous code was not written correctly either since we are casting the config. Which might fail as well.

blagoev: I will be fixing this

src/realm_dart.cpp
nirinchev: Again, the function being passed as an argument is not an unlock function. It's a function(bool).

nirinchev: Consider this extreme example:

```dart
void bark(Object dog) {
  final animal = dog as Animal;
  animal.produceSound();
}

final dog = Dog();
dog.bark();
```

Do you think the bark function is named correctly here? Because this is what is essentially happening with this C function. We accept a generic void* pointer, we cast it to a function(bool), then invoke that function. And we call the argument `unlockFunc` because we expect that consumers are going to invoke us with a function that unlocks something.

lib/src/native/realm_core.dart
blagoev: I will be fixing it

src/realm_dart.cpp
blagoev: I don't think we are getting anywhere here. I would like to move away from bike shading the function name. The function is called `invoke_unlock_callback` and it does what it says and expect a specific argument to be passed in and be used in a specific case. Lets move the discussion to another place where we can gain more value.
The other feedback for deadlocks is more important in this PR. I am trying to handle it.

lib/src/native/realm_core.dart
blagoev: handled

lib/src/native/realm_core.dart
blagoev: fixed

lib/src/native/realm_core.dart
blagoev: fixed

lib/src/configuration.dart
blagoev: fixed

lib/src/native/realm_core.dart
```diff
@@ -500,52 +502,59 @@ class _RealmCore {
...
+      _realmLib.realm_dart_invoke_unlock_callback(success, unlockCallbackFunc);
     }
-    final realm = RealmInternal.getUnowned(syncConfig, RealmHandle._unowned(realmHandle));
-    _continueWhenComplete(() => beforeResetCallback(realm), unlockCallbackFunc);
+  }
+
+  static void _syncBeforeResetCallback(Object userdata, Pointer<shared_realm> realmHandle, Pointer<Void> unlockCallbackFunc) {
+    _guardSynchCallback(() async {
+      final syncConfig = userdata as FlexibleSyncConfiguration;
+      var beforeResetCallback = syncConfig.clientResetHandler.onBeforeReset!;
```
desistefanova, line 526: I really prefer to have the same check here:
```dart
 if (beforeResetCallback== null) {
        return Future<void>.value();
      }
```
If we throw if for some reason this callback is null then we will always fail the automatics recovery and discard, just because there is no attached callback.

blagoev: this should fail and silence the error. This callback should never be called with onBeforeReset null

src/realm_dart.cpp
nirinchev: I'm fine with moving the discussion elsewhere, but I disagree with changing the function name here. Since both Desi and I believe the old name is better and the name change is not fixing an actual bug, let's keep the original name and we can rename it once we've completed bikeshedding.

lib/src/native/realm_core.dart
desistefanova: I don't understand your point. The callback that the users set is clientResetHandler.onBeforeReset but the callback that we pass to core is _syncBeforeResetCallback eventually. I know we have a check while we are creating the config, but should we rely on it? I think that it is better and safety to have the correct check here. If the user hadn't provide a callback we shouldn't break the recover and discard to continue.

desistefanova (approved):

lib/src/native/realm_core.dart
```diff
@@ -500,52 +502,59 @@ class _RealmCore {
...
+    syncConfig.syncErrorHandler(syncError);
   }
 
-  static void _syncBeforeResetCallback(Object userdata, Pointer<shared_realm> realmHandle, Pointer<Void> unlockCallbackFunc) {
-    final syncConfig = userdata as FlexibleSyncConfiguration;
-    final beforeResetCallback = syncConfig.clientResetHandler.onBeforeReset;
-    if (beforeResetCallback == null) {
-      _invokeNativeFunction(unlockCallbackFunc);
-      return;
+  static void _guardSynchCallback(FutureOr<void> Function() callback, Pointer<Void> unlockCallbackFunc) async {
```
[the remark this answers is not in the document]
blagoev: _guardSynchonousCallback ?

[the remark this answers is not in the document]
desistefanova: I will rename it to _guardSynchronousCallback in [Client reset PR](https://github.com/realm/realm-dart/pull/925)

PR #1005, opened 2022-11-03
realm compact

### LINKED ISSUE #803: Realm Compaction support
realm_compact,  https://www.mongodb.com/docs/realm/sdk/dotnet/advanced-guides/compact-realm/

blagoev:
Support realm compact. 

fixes #803


nielsenko (approved):
LGTM 

I think it is prudent not to spawn an isolate - users can do that themselves. Would be nice if realm-core exposed an async interface for this though.

blagoev:
@nielsenko  added a test for a worker isolate.

test/realm_test.dart
```diff
@@ -1394,6 +1395,58 @@ Future<void> main([List<String>? args]) async {
...
   });
+
+  test('Realm - local realm can be compacted', () {
+    var config = Configuration.local([Car.schema]);
+    var realm = getRealm(config);
+    final compacted = realm.compact();
+    expect(compacted, true);
+  });
+
+  test('Realm - local realm can be compacted in worker isolate', () async {
```
blagoev, line 1406: Should we provide a public API for this? Or perhaps this could be the only API ie 
Future<bool> compact()  // works in a worker isolate. This could trigger migration callbacks in that isolate, which could bring problems.

nielsenko: I suggest we keep it as a test for now. It gets tricky quickly. For one realms cannot be passed between isolates since they hold references to native types.

nielsenko (approved):

desistefanova (commented):
I think we need to test if:

- a compacted realm could be opened
- a synced realm populated with data could be compacted and the opened with the same number of rows.
- the size of compacted realm file is less than the size of a synced populated ream that had been frozen before the compaction.

lib/src/realm_class.dart
```diff
@@ -483,6 +483,40 @@ class Realm implements Finalizable {
...
     }
   }
+
+  /// Compacts a Realm file. A Realm file usually contains free/unused space.
+  ///
+  /// This method removes this free space and the file size is thereby reduced.
+  /// Objects within the Realm file are untouched.
+  /// Note: The file system should have free space for at least a copy of the Realm file. This method must not be called inside a transaction.
+  /// The Realm file is left untouched if any file operation fails.
+  static Future<bool> compact(Configuration config) {
```
nirinchev, line 493: Why is this a future?

blagoev: This is cause this operation is long running and also we may decide to to that on a background isolate later. This is why I changed the method signature to accept a Configuration object.

test/test.dart
```diff
@@ -465,13 +465,20 @@ Future<void> tryDeleteRealm(String path) async {
...
   for (var i = 0; i < 5; i++) {
     try {
       Realm.deleteRealm(path);
-      await File('$path.lock').delete();
+      
+      //delete lock file
+      await File('$path.lock').delete().onError((error, stackTrace) => dummy);
+
+      //delete compaction space file
+      await File('$path.tmp_compaction_space').delete().onError((error, stackTrace) => dummy);
```
desistefanova, line 478: Why do we hide the errors when deleting these files?

lib/src/realm_class.dart
blagoev: I don't understand what you mean. How a Future result is a non blocking operation. If you await on it when you invoke the method you are blocking at the point of the call and waiting for the result to continue from that point forward.

lib/src/realm_class.dart
```diff
@@ -483,6 +483,40 @@ class Realm implements Finalizable {
...
+    late Configuration compactConfig;
+    if (config is LocalConfiguration) {
```
desistefanova, line 495: I'm wondering why we need to recreate the configs before compact.

lib/src/realm_class.dart
blagoev: In other words a method that returns a Future does not guarantee it will not execute synchronously

test/realm_test.dart
```diff
@@ -1470,11 +1470,14 @@ Future<void> main([List<String>? args]) async {
...
-    final configuration =
+    final config =
         Configuration.flexibleSync(user, [Task.schema], encryptionKey: key, path: p.join(Configuration.defaultStoragePath, "${generateRandomString(8)}.realm"));
 
-    final compacted = await Realm.compact(configuration);
+    final compacted = await Realm.compact(config);
     expect(compacted, true);
+
+    //test the realm can be opened. This also allows the compacted realm to be deleted after the test
+    final realm = getRealm(config);
```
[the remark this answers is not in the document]
blagoev: we have already the infrastructure in place and also every test is compacting a realm with different config.

test/realm_test.dart
```diff
@@ -1470,11 +1470,14 @@ Future<void> main([List<String>? args]) async {
...
     List<int> key = List<int>.generate(encryptionKeySize, (i) => random.nextInt(256));
-    final configuration =
+    final config =
         Configuration.flexibleSync(user, [Task.schema], encryptionKey: key, path: p.join(Configuration.defaultStoragePath, "${generateRandomString(8)}.realm"));
 
-    final compacted = await Realm.compact(configuration);
+    final compacted = await Realm.compact(config);
     expect(compacted, true);
+
+    //test the realm can be opened. This also allows the compacted realm to be deleted after the test
```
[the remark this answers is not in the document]
blagoev: no need to do a special case for that now that every test is doing it with different config

nirinchev (changes requested):
The API should be sync

test/test.dart
```diff
@@ -465,13 +465,20 @@ Future<void> tryDeleteRealm(String path) async {
...
+
```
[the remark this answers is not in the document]
blagoev: I am not sure its worth it. The performance difference will be negligible.

test/test.dart
blagoev: these files might not exists. So this is operatio is "try deleting the file and ingore if it exists" In one operation. Otherwise I will need check if the file exists first and then delete it, which is two operations to the file system which should be slower than just trying and ignoring the error.

lib/src/realm_class.dart
blagoev: the idea is not to invoke any callbacks while opening from within this method. We copy only the relevant fields in the new configs

lib/src/realm_class.dart
blagoev: we will continue the discussion offline

nielsenko: I think people are right to expect that the thread will not be blocked for more than a few ms when awaiting an async function. Even if no such guarantees exists, it is still the expectation. 

Perhaps we can wait with exposing this until we figure out how to move the work onto a background isolate?

lib/src/realm_class.dart
nielsenko: It is because `disableFormatUpgrade` and `isReadOnly` is set differently

test/realm_test.dart
```diff
@@ -1394,6 +1395,90 @@ Future<void> main([List<String>? args]) async {
...
+    final config = Configuration.local([Friend.schema],
+        encryptionKey: generateEncryptionKey(), path: p.join(Configuration.defaultStoragePath, "${generateRandomString(8)}.realm"));
+    final compacted = await Realm.compact(config);
+    expect(compacted, true);
+
+    //test the realm can be opened. This also allows the compacted realm to be deleted after the test
+    final realm = getRealm(config);
+  });
+
+  test('Realm - in-memory realm can be compacted', () async {
```
desistefanova, line 1437: We have add a test for Read-Only realm and to expect an exception.

nielsenko (changes requested):

lib/src/realm_class.dart
blagoev: wonder if we should not throw if `isReadOnly` is set?

lib/src/realm_class.dart
```diff
@@ -483,6 +483,40 @@ class Realm implements Finalizable {
...
+      compactConfig = Configuration.local(config.schemaObjects.toList(),
+          schemaVersion: config.schemaVersion,
+          fifoFilesFallbackPath: config.fifoFilesFallbackPath,
+          path: config.path,
+          encryptionKey: config.encryptionKey,
+          disableFormatUpgrade: true,
+          isReadOnly: false);
```
desistefanova, line 502: This will break the rule that read-only realms can not be compacted. You will always compact them. It must be `isReadOnly: config.isReadOnly`

blagoev: I agree. will fix

test/realm_test.dart
blagoev: done

lib/src/realm_class.dart
blagoev: fixed

lib/src/realm_class.dart
```diff
@@ -483,6 +483,40 @@ class Realm implements Finalizable {
...
+          schemaVersion: config.schemaVersion,
+          fifoFilesFallbackPath: config.fifoFilesFallbackPath,
+          path: config.path,
+          encryptionKey: config.encryptionKey,
+          disableFormatUpgrade: true,
+          isReadOnly: config.isReadOnly);
+    } else if (config is FlexibleSyncConfiguration || config is DisconnectedSyncConfiguration) {
+      compactConfig = Configuration.disconnectedSync(config.schemaObjects.toList(),
+          fifoFilesFallbackPath: config.fifoFilesFallbackPath, path: config.path, encryptionKey: config.encryptionKey);
+    } else if (config is InMemoryConfiguration) {
```
nirinchev, line 506: Why can we even compact an in memory realm? Should this be an exception instead?

test/realm_test.dart
```diff
@@ -1394,6 +1395,103 @@ Future<void> main([List<String>? args]) async {
...
+    var config = Configuration.inMemory([Car.schema], path: p.join(Configuration.defaultStoragePath, "${generateRandomString(8)}.realm"));
+    final compacted = Realm.compact(config);
+    expect(compacted, true);
+
+    //test the realm can be opened. This also allows the compacted realm to be deleted after the test
+    final realm = getRealm(config);
+  });
+
+    test('Realm - readonly realm can not be compacted', () async {
```
nirinchev, line 1446: Indentation issue

test/realm_test.dart
```diff
@@ -1394,6 +1395,103 @@ Future<void> main([List<String>? args]) async {
...
+    realm.close();
+
+    config = Configuration.local([Car.schema], isReadOnly: true, path: path);
+    expect(() => Realm.compact(config), throws<RealmException>("Can't compact a read-only Realm"));
+
+    //test the realm can be opened. This also allows the compacted realm to be deleted after the test
+    realm = getRealm(config);
+  });
+
+   test('Realm - disconnected sync realm can be compacted', () async {
```
nirinchev, line 1459: Indentation issue

nirinchev (changes requested):
All these tests are testing you can compact a non-existent realm - we don't open the file before compacting it. We don't test the behavior of calling compact on a Realm that is still open and we don't test for the effects of compact. Right now, the API could return `true` without any call to Core and the tests will pass. The way we do it in the .net tests is to add and remove a bunch of data, then verify that compact reduces the file size.

lib/src/realm_class.dart
blagoev: I was contemplating that but decided to support it since core can do it

nirinchev: Can you ping them and see if it actually does something? The thing is, in-memory Realms only live while somebody is referencing them, while compact requires that all realm instances are closed. So if I close the in-memory realm, it goes away and then compacting it will recreate it, compact the empty realm, then close it, thus destroying it.

blagoev: Why can we even compact an in memory realm?
>Free some memory perhaps

blagoev: I will throw an exception

blagoev: fixed

test/realm_test.dart
```diff
@@ -1394,6 +1396,166 @@ Future<void> main([List<String>? args]) async {
...
+
+    realm.close();
+    return beforeSize;
+  }
+
+  void validateCompact(bool compacted, String realmPath, int before) async {
+    expect(compacted, true);
+    final afterSize = await File(realmPath).stat().then((value) => value.size);
+    expect(before, greaterThan(afterSize));
+  }
```
desistefanova, line 1443:
```suggestion
  void validateCompact(bool compacted, String realmPath, int beforeCompactSize) async {
    expect(compacted, true);
    final afterCompactSize = await File(realmPath).stat().then((value) => value.size);
    expect(beforeCompactSize, compacted? greaterThan(afterCompactSize) : afterCompactSize);
  }
```

test/realm_test.dart
```diff
@@ -1394,6 +1396,166 @@ Future<void> main([List<String>? args]) async {
...
+
+  Future<int> createRealmForCompact(Configuration config) async {
+    var realm = getRealm(config);
+    
+    if (config is FlexibleSyncConfiguration) {
+      realm.subscriptions.update((mutableSubscriptions) {
+        mutableSubscriptions.add(realm.all<Task>());
+      });
+      await realm.subscriptions.waitForSynchronization();
```
nirinchev, line 1423: We should avoid doing that as this introduces unpleasant dependencies between tests. I.e. if we have other tests that are adding `Task` objects, those will get synchronized here and we'll start the test with non-0 number of tasks. The way to avoid this is for example adding a `differentiator` field which works similar to a partition. So you do something like:

```dart
final differentiator = ObjectId();
realm.subscriptions.update((mutableSubs) {
  mutableSubs.add(realm.all<Task>().filter('differentiator = $0', differentiator);
});

addDataFroCompact(realm, differentiator);
```

That way, you're certain that every time you execute the test, it'll do the same thing and have the exact same number of objects.

test/test.dart
nirinchev: Why is this file not automatically deleted by Core? It sounds like a bug that we need to clean it up ourselves.

nirinchev (commented):
Looks mostly there - the main remaining thing is to make sure we're not polluting tests by subscribing to all objects from a collection.

test/test.dart
desistefanova: Isn't it better to delete inside the implementation it Realm.compact? So that our users will not have to care about it.

test/realm_test.dart
desistefanova: I agree. For this reason I created the Product class. So that I can add the differentiator in product name.

test/test.dart
nirinchev: Core should not leave these around. If it does, it's a bug on their end.

test/realm_test.dart
blagoev: will fix

test/realm_test.dart
blagoev: this is not needed since the first expect will throw.

test/realm_test.dart
```diff
@@ -1394,6 +1396,173 @@ Future<void> main([List<String>? args]) async {
...
+    }
+
+    addDataForCompact(realm);
+
+    final beforeSize = await File(config.path).stat().then((value) => value.size);
```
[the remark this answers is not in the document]
blagoev: there is no data to be downloaded this is should be an empty realm.

test/realm_test.dart
```diff
@@ -1394,6 +1396,166 @@ Future<void> main([List<String>? args]) async {
...
+    realm.write(() {
+      for (var i = 0; i < 2500; i++) {
+        realm.add(Task(ObjectId()));
+      }
+    });
+
+    realm.write(() => realm.deleteAll<Task>());
+
+    realm.write(() {
+      for (var i = 10; i < 20; i++) {
```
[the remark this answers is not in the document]
blagoev: fixed

test/realm_test.dart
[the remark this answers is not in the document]
blagoev: fixed

lib/src/realm_class.dart
```diff
@@ -497,6 +497,44 @@ class Realm implements Finalizable {
...
+
+    late Configuration compactConfig;
+
+    if (!File(config.path).existsSync()) {
+      return false;
+    }
+
+    if (config is LocalConfiguration) {
+      //compact opens the realm file so it can triger schema version upgrade, file format upgrade, migration and initial data callbacks etc.
+      //We must to allow that to happen so use the local config as is.
```
desistefanova, line 520:
```suggestion
      // `compact` opens the realm file so it can triger schema version upgrade, file format upgrade, migration and initial data callbacks etc.
      // We must allow that to happen so use the local config as it is.
```

test/realm_test.dart
desistefanova: What about renaming the variable and parameter names to beforeCompactSize and afterCompactSize?

blagoev: I don't think it matters much.

test/realm_test.dart
```diff
@@ -1394,6 +1396,173 @@ Future<void> main([List<String>? args]) async {
...
+
+  void validateCompact(bool compacted, String realmPath, int before) async {
+    expect(compacted, true);
+    final afterSize = await File(realmPath).stat().then((value) => value.size);
+    expect(before, greaterThan(afterSize));
+  }
+
+  test('Realm - local realm can be compacted', () async {
+    var config = Configuration.local([Product.schema], path: p.join(Configuration.defaultStoragePath, "${generateRandomString(8)}.realm"));
+    final beforeCompact = await createRealmForCompact(config);
```
desistefanova, line 1448: Let's say that this is the size:  beforeCompactSize

test/realm_test.dart
blagoev: renamed. Hope it is more clear and readable now

test/realm_test.dart
blagoev: alright. as you wish.

nirinchev (approved):

test/realm_test.dart
```diff
@@ -1394,6 +1396,173 @@ Future<void> main([List<String>? args]) async {
...
+      var config = Configuration.local([Product.schema], path: path);
+      final compacted = Realm.compact(config);
+      Isolate.exit(sendPort, compacted);
+    }, [receivePort.sendPort, config.path]);
+
+    final compacted = await receivePort.first as bool;
+
+    validateCompact(compacted, config.path, beforeCompactSizeSize);
+
+    //test the realm can be opened.
```
desistefanova, line 1481:
```suggestion
// Test the realm can be opened.
```
In all the tests.

test/realm_test.dart
```diff
@@ -1394,6 +1396,173 @@ Future<void> main([List<String>? args]) async {
...
+    final realm = getRealm(config);
+  });
+
+  baasTest('Realm - synced realm can be compacted', (appConfiguration) async {
+    final app = App(appConfiguration);
+    final credentials = Credentials.anonymous();
+    final user = await app.logIn(credentials);
+    final path = p.join(Configuration.defaultStoragePath, "${generateRandomString(8)}.realm");
```
desistefanova, line 1533: We don't have to generate path here. The path is already unique by the user id. FlexibleSyncConfiguration will generate it unique.

test/realm_test.dart
```diff
@@ -1394,6 +1396,173 @@ Future<void> main([List<String>? args]) async {
...
+    final beforeCompactSize = await createRealmForCompact(config);
+    user.logOut();
+    Future<void>.delayed(Duration(seconds: 5));
+
+    final compacted = Realm.compact(config);
+
+    validateCompact(compacted, config.path, beforeCompactSize);
+
+    //test the realm can be opened.
+    final realm = getRealm(Configuration.disconnectedSync([Product.schema], path: path, encryptionKey: key));
```
desistefanova, line 1564: I would add `expect(realm.isClosed, false);` at end of each test, but it is not so important.

test/test.dart
```diff
@@ -467,13 +467,21 @@ Future<void> tryDeleteRealm(String path) async {
...
       Realm.deleteRealm(path);
-      await File('$path.lock').delete();
+      
+      //delete lock file
+      await File('$path.lock').delete().onError((error, stackTrace) => dummy);
+
+      //Bug in Core https://github.com/realm/realm-core/issues/5997. Remove when fixed
+      //delete compaction space file
+      await File('$path.tmp_compaction_space').delete().onError((error, stackTrace) => dummy);
+
```
desistefanova, line 482:
```suggestion
      // Delete lock file.
      await File('$path.lock').delete().onError((error, stackTrace) => dummy);

      // Bug in Core https://github.com/realm/realm-core/issues/5997. Remove when fixed.
      // Delete compaction space file.
      await File('$path.tmp_compaction_space').delete().onError((error, stackTrace) => dummy);
```
Just formatting comments.

test/test.dart
```diff
@@ -467,13 +467,21 @@ Future<void> tryDeleteRealm(String path) async {
...
     try {
       Realm.deleteRealm(path);
-      await File('$path.lock').delete();
+      
+      //delete lock file
+      await File('$path.lock').delete().onError((error, stackTrace) => dummy);
+
+      //Bug in Core https://github.com/realm/realm-core/issues/5997. Remove when fixed
+      //delete compaction space file
+      await File('$path.tmp_compaction_space').delete().onError((error, stackTrace) => dummy);
```
desistefanova, line 481:
```suggestion
      // Delete lock file.
      await File('$path.lock').delete().onError((error, stackTrace) => dummy);

      // Bug in Core https://github.com/realm/realm-core/issues/5997. Remove when fixed.
      // Delete compaction space file.
      await File('$path.tmp_compaction_space').delete().onError((error, stackTrace) => dummy);
```
Just formatting comments.

desistefanova (approved):
I have added some small suggestions about formatting and naming. It looks good.

test/realm_test.dart
blagoev: I don't agree. Is there an issue with generating it.

blagoev: I will leave it as it is.

PR #1007, opened 2022-11-04
Make path required for DisconnectedSync

blagoev:


nirinchev (commented):
I think a good test would be one where you don't specify the path explicitly:

```dart
final config = Configuration.flexibleSync(user);
final realm = await createRealmForCompact(config);

final compacted = Realm.compact(config);
```

That way we verify that the disconnected sync change really works. Additionally, we could write a simple test that verifies that the path generated via `Configuration.flexibleSync` is the same as the path for `Configuration.disconnectedSync`.

desistefanova (approved):

nielsenko (approved):
It is getting late. If it is important to add to 0.7.0+rc then you have my approval. I don't fully grok why we need a user, when a user is not needed by core.

blagoev:
@nielsenko I changed this to require a path

PR #1037, opened 2022-11-15
List.move

nielsenko:
- Use updated realm-core
- Update bindings
- Implement List.move
- Add tests of List.move
- Update CHANGELOG

Resolves: #794

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 ### Enhancements
 * Support setting `maxNumberOfActiveVersions` when creating a `Configuration`. ([#1036](https://github.com/realm/realm-dart/pull/1036))
+* Add ManagedRealmList.move method that allows notifications to correctly report moves, as opposed to reporting moves as deletes + inserts. ([#1037](https://github.com/realm/realm-dotnet/issues/1037))
 
 ### Fixed
 * Support mapping into `SyncSessionErrorCode` for "Compensating write" with error code 231. ([#1022](https://github.com/realm/realm-dart/pull/1022))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 
 ### Enhancements
 * Support setting `maxNumberOfActiveVersions` when creating a `Configuration`. ([#1036](https://github.com/realm/realm-dart/pull/1036))
-* Add ManagedRealmList.move method that allows notifications to correctly report moves, as opposed to reporting moves as deletes + inserts. ([#1037](https://github.com/realm/realm-dotnet/issues/1037))
+* Add List.move extension method that moves an element from one index to another. Delegates to ManagedRealmList.move for managed lists. This allows notifications to correctly report moves, as opposed to reporting moves as deletes + inserts. ([#1037](https://github.com/realm/realm-dart/issues/1037))
 
 ### Fixed
 * Support mapping into `SyncSessionErrorCode` for "Compensating write" with error code 231. ([#1022](https://github.com/realm/realm-dart/pull/1022))
--- a/lib/src/list.dart
+++ b/lib/src/list.dart
@@ -145,9 +145,9 @@ class ManagedRealmList<T extends Object?> with RealmEntity, ListMixin<T> impleme
     return result;
   }
 
-  /// Move the element at [from] to [to].
-  void move(int oldIndex, int newIndex) {
-    realmCore.listMoveElement(handle, oldIndex, newIndex);
+  /// Move the element at index [from] to index [to].
+  void move(int from, int to) {
+    realmCore.listMoveElement(handle, from, to);
   }
 
   /// Removes all objects from this list; the length of the list becomes zero.
@@ -334,7 +334,7 @@ class ListNotificationsController<T extends Object?> extends NotificationsContro
 }
 
 extension ListExtension<T> on List<T> {
-  /// Move the element at [from] to [to].
+  /// Move the element at index [from] to index [to].
   void move(int from, int to) {
     RangeError.checkValidIndex(from, this, 'from', length);
     RangeError.checkValidIndex(to, this, 'to', length);
--- a/lib/src/list.dart
+++ b/lib/src/list.dart
@@ -16,6 +16,7 @@
 //
 ////////////////////////////////////////////////////////////////////////////////
 
+import 'dart:core';
 import 'dart:async';
 import 'dart:collection';
 import 'dart:ffi';
@@ -142,6 +143,10 @@ class ManagedRealmList<T extends Object?> with RealmEntity, ListMixin<T> impleme
     final result = this[index];
     realmCore.listRemoveElementAt(handle, index);
     return result;
+  }
+
+  void move(int oldIndex, int newIndex) {
+    realmCore.listMoveElement(handle, oldIndex, newIndex);
   }
 
   /// Removes all objects from this list; the length of the list becomes zero.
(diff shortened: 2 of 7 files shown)


nirinchev (approved):

test/list_test.dart
```diff
@@ -1066,4 +1066,82 @@ Future<void> main([List<String>? args]) async {
...
+    final list = [0, 1, 2, 3];
+    list.move(1, 0);
+    expect(list, [1, 0, 2, 3]);
+    list.move(2, 3);
+    expect(list, [1, 0, 3, 2]);
+    final length = list.length;
+    expect(() => list.move(-1, 0), throwsRangeError);
+    expect(() => list.move(0, -1), throwsRangeError);
+    expect(() => list.move(length, 0), throwsRangeError);
+    expect(() => list.move(0, length), throwsRangeError);
```
desistefanova, line 1087: What about moving from one index to the same index. Can we add it to the tests? Also for managed realm list.

test/list_test.dart
```diff
@@ -1066,4 +1066,82 @@ Future<void> main([List<String>? args]) async {
...
+    expect(team.players, [bob, alice, carol, dan]);
+
+    realm.write(() => team.players.move(2, 3));
+    expect(team.players, [bob, alice, dan, carol]);
+
+    final length = team.players.length;
+    expect(() => realm.write(() => team.players.move(-1, 0)), throwsRangeError);
+    expect(() => realm.write(() => team.players.move(0, -1)), throwsRangeError);
+    expect(() => realm.write(() => team.players.move(length, 0)), throwsRangeError);
+    expect(() => realm.write(() => team.players.move(0, length)), throwsRangeError);
```
desistefanova, line 1115: We could check moving outside write block. Is it throwing the correct exception?

test/list_test.dart
```diff
@@ -1066,4 +1066,82 @@ Future<void> main([List<String>? args]) async {
...
+    list.move(2, 3);
+    expect(list, [1, 0, 3, 2]);
+    final length = list.length;
+    expect(() => list.move(-1, 0), throwsRangeError);
+    expect(() => list.move(0, -1), throwsRangeError);
+    expect(() => list.move(length, 0), throwsRangeError);
+    expect(() => list.move(0, length), throwsRangeError);
+  });
+
+  test('ManagedRealmList.move', () {
```
desistefanova, line 1090: What about trying to move in Frozen Realm, RealmList, or in Read-only Realm? Do you think we can test them?

lib/src/list.dart
```diff
@@ -144,6 +145,10 @@ class ManagedRealmList<T extends Object?> with RealmEntity, ListMixin<T> impleme
     return result;
   }
 
+  void move(int oldIndex, int newIndex) {
```
desistefanova, line 148: This method `move` is not visible in the API doc neither for List or RealmList. It is visible on the class `ListExtension` and it is not documented. Is it possible to have in the documentation `move` method described for `RealmList` and `List`?

lib/src/list.dart
```diff
@@ -326,3 +331,18 @@ class ListNotificationsController<T extends Object?> extends NotificationsContro
     streamController.addError(error);
   }
 }
+
+extension ListExtension<T> on List<T> {
+  void move(int from, int to) {
```
desistefanova, line 336: Missing API doc. Looks strange in the documentation.

test/list_test.dart
nielsenko: That is a no-op, but sure I can add a test

test/list_test.dart
nielsenko: Throws a RealmException, if not a RangeError, since you need to be in a write transaction. I can add a test.

test/list_test.dart
nielsenko: You need a write transaction to modify a realm object. Frozen realms and read-only realms won't allow that. It should not be tested her

lib/src/list.dart
nielsenko: Will add

test/list_test.dart
nielsenko: It will look odd here

desistefanova: Ah, I see now. OK. Leave it as it is.

desistefanova (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 
 ### Enhancements
 * Support setting `maxNumberOfActiveVersions` when creating a `Configuration`. ([#1036](https://github.com/realm/realm-dart/pull/1036))
+* Add ManagedRealmList.move method that allows notifications to correctly report moves, as opposed to reporting moves as deletes + inserts. ([#1037](https://github.com/realm/realm-dotnet/issues/1037))
```
blagoev, line 10: realm-dotnet?

blagoev: Also while this is technically true move also rearranges items. Why not mention the real feature here. Isnt the notifications more like a side effect of this feature

blagoev (changes requested):

CHANGELOG.md
nielsenko: I agree, I should also mention that the element is actually moved 😄 

How about:
> * Add List.move extension method that moves an element from one index to another. Delegates to ManagedRealmList.move for managed lists. This allows notifications to correctly report moves, as opposed to reporting moves as deletes + inserts. ([#1037](https://github.com/realm/realm-dart/issues/1037))

I do kind of think the notification part is the important feature, since that is the reason we don't just implement `move` as:
```dart
void move(int from, int to) => insert(to, removeAt(from));
```

lib/src/list.dart
```diff
@@ -326,3 +332,18 @@ class ListNotificationsController<T extends Object?> extends NotificationsContro
     streamController.addError(error);
   }
 }
+
+extension ListExtension<T> on List<T> {
+  /// Move the element at [from] to [to].
+  void move(int from, int to) {
```
blagoev, line 338: should we rename to `fromIndex` `toIndex` and also the other one that is oldIndex to be `fromIndex` as well

blagoev (changes requested):
just wonder should we rename the parameters

CHANGELOG.md
blagoev: it is good to mention the notification as well

lib/src/list.dart
blagoev: here https://github.com/realm/realm-dart/pull/1037/files#diff-847b6223dda3c6c9a63c5b47a2dcc1b983c5035fc0cd6ccc152286c5c8449ef7R149

lib/src/list.dart
```diff
@@ -326,3 +332,18 @@ class ListNotificationsController<T extends Object?> extends NotificationsContro
     streamController.addError(error);
   }
 }
+
+extension ListExtension<T> on List<T> {
+  /// Move the element at [from] to [to].
```
blagoev, line 337:
```suggestion
  /// Move the element at [from] index to [to] index.
```

nielsenko: I could also rename the arguments to `fromIndex` and `toIndex`. Would that read better?

lib/src/list.dart
blagoev: or maybe use `from` and `to` everywhere since the API help says these are `indexes`

lib/src/list.dart
```diff
@@ -144,6 +145,11 @@ class ManagedRealmList<T extends Object?> with RealmEntity, ListMixin<T> impleme
     return result;
   }
 
+  /// Move the element at [from] to [to].
```
blagoev, line 148:
```suggestion
  /// Move the element at [from] index to [to] index.
```

blagoev: the current API help does not correctly state the parameter names

lib/src/list.dart
nielsenko: Done

blagoev (approved):

nielsenko:
> just wonder should we rename the parameters

Use `from`/`to` consistently now.

PR #1044, opened 2022-11-24
implement NativeFinalizers in Realm

blagoev:
implement `NativeFinalizers` in Realm cause the Dart VM `NativeFinalizers` have an issue which causes the Dart GC to run too often cause of a bug while collecting memory. 
Dart VM issue here https://github.com/dart-lang/sdk/issues/50537

### DIFF
--- a/lib/src/native/realm_core.dart
+++ b/lib/src/native/realm_core.dart
@@ -85,6 +85,11 @@ class _RealmCore {
   static const libraryVersion = '0.7.0+rc';
   late String nativeLibraryVersion = _realmLib.realm_dart_library_version().cast<Utf8>().toDartString();
 
+  // for debugging only. Enable in realm_dart.cpp
+  // void invokeGC() {
+  //   _realmLib.realm_dart_gc();
+  // }
+
   LastError? getLastError(Allocator allocator) {
     final error = allocator<realm_error_t>();
     final success = _realmLib.realm_get_last_error(error);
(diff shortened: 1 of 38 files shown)


lib/src/native/realm_core.dart
```diff
@@ -85,6 +85,11 @@ class _RealmCore {
   static const libraryVersion = '0.7.0+rc';
   late String nativeLibraryVersion = _realmLib.realm_dart_library_version().cast<Utf8>().toDartString();
 
+  // for debugging only. Enable in realm_dart.cpp
```
blagoev, line 88: Should we leave this commented out here for the future?

src/realm_dart.cpp
```diff
@@ -94,4 +94,24 @@ RLM_API void realm_dart_invoke_unlock_callback(bool success, void* unlockFunc)
 }
 
 // stamped into the library by the build system (see prepare-release.yml)
-RLM_API const char* realm_dart_library_version() { return "0.8.0+rc"; }
+RLM_API const char* realm_dart_library_version() { 
+    return "0.8.0+rc"; 
+}
```
nirinchev, line 99: This will break the prepare-release workflow - it does simple string replacement without parsing the syntax tree.

blagoev: thanks for noticing will revert and put a comment there

nielsenko (approved):

desistefanova (approved):

PR #1046, opened 2022-11-29
support realm refresh()

blagoev:
Add support for Realm.refresh() and onRefresh callbacks
#793


lib/src/realm_class.dart
```diff
@@ -539,6 +539,21 @@ class Realm implements Finalizable {
...
+  /// Typically you don't need to call this method since Realm has auto-refresh built-in.
+  bool refresh() {
+    return realmCore.realmRefresh(this);
+  }
+
+  /// Returns a [Future] that will complete the first time the realm is refreshed to the version which is the
+  /// latest version at the time when this method is called.
+  Future<void> onRefresh(void Function() action) {
+    return realmCore.onRealmRefresh(this).whenComplete(action);
+  }
```
nirinchev, line 556: Why is this taking an action? I'd imagine an API like `Future<void> refreshAsync()` and if you want to do something after the refresh you use:

```
await realm.refreshAsync();
action();
```

It seems redundant to be exposing this as both a promise and a method that takes a callback.

blagoev: This is not refreshAsync implementation but rather a add_refresh_callback implementation. This takes a callback which will get executed only once and returns a Future since you might want to await it or most probably not. Just register the callback.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### Enhancements
 * Support setting `maxNumberOfActiveVersions` when creating a `Configuration`. ([#1036](https://github.com/realm/realm-dart/pull/1036))
 * Add List.move extension method that moves an element from one index to another. Delegates to ManagedRealmList.move for managed lists. This allows notifications to correctly report moves, as opposed to reporting moves as deletes + inserts. ([#1037](https://github.com/realm/realm-dart/issues/1037))
+* Add `Realm.refresh()` support. ([#1046](https://github.com/realm/realm-dart/pull/1046))
```
desistefanova, line 12: Why not to mention also `refreshAsync`?

desistefanova, line 12: Why not to mention also `refreshAsync`?

test/realm_test.dart
```diff
@@ -1567,6 +1566,75 @@ Future<void> main([List<String>? args]) async {
...
+    expect(called, true, reason: "beginWrite onRefresh failed");
+
+    called = false;
+    realm.write(() {
+      realm.refreshAsync().then((_) => called = true);
+      realm.add(Person("name"));
+    });
+
+    await Future<void>.delayed(Duration(milliseconds: 1));
+    expect(called, true, reason: "write onRefresh failed");
```
desistefanova, line 1594: This test looks strange to me. Why "write onRefresh failed" when called == true? Is it expected the Person not to be stored in the realm?

test/realm_test.dart
```diff
@@ -1567,6 +1566,75 @@ Future<void> main([List<String>? args]) async {
...
+
+  test('Realm.onRefresh sync transaction', () async {
+    final realm = getRealm(Configuration.local([Person.schema]));
+    var called = false;
+    final transaction = realm.beginWrite();
+    realm.refreshAsync().then((_) => called = true);
+    realm.add(Person("name"));
+    transaction.commit();
+
+    await Future<void>.delayed(Duration(milliseconds: 1));
```
desistefanova, line 1584: Is it correct to relly on the delay for 1 milisecond to ensure that the `refreshAsync` is completed?

test/realm_test.dart
```diff
@@ -1567,6 +1566,75 @@ Future<void> main([List<String>? args]) async {
...
+
+    expect(called, true);
+  });
+
+   test('Realm.onRefresh on frozen realm should be no-op', () async {
+    var realm = getRealm(Configuration.local([Person.schema]));
+    bool called = false;
+    realm = realm.freeze();
+
+    await Future.any([realm.refreshAsync().then((_) => called = true), Future<void>.delayed(Duration(milliseconds: 10))]);
```
desistefanova, line 1634: Probably you could use `waitFutureWithTimeout`.

desistefanova (commented):
I have added some questions.

test/realm_test.dart
blagoev: This is a message if the expect fails. Not that it expect it to fail

CHANGELOG.md
blagoev: will add it

test/realm_test.dart
blagoev: yes, this makes sure to release the event loop.

test/realm_test.dart
```diff
@@ -1567,6 +1566,75 @@ Future<void> main([List<String>? args]) async {
...
+  });
+
+  test('Realm.onRefresh async transaction', () async {
+    final realm = getRealm(Configuration.local([Person.schema]));
+    bool called = false;
+    final transaction = await realm.beginWriteAsync();
+    realm.refreshAsync().then((_) => called = true);
+    realm.add(Person("name"));
+    await transaction.commitAsync();
+    expect(called, true);
```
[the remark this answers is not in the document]
blagoev: yes since it releases the event loop.

test/realm_test.dart
blagoev: This is test specific. The problem with these functions is they become part of the internal testing API. And here problem is highlighted. I prefer to use the common Future API instead of increasing the size of the internal API which then one needs to remember that it exists

lib/src/native/realm_core.dart
```diff
@@ -764,8 +764,35 @@ class _RealmCore {
     _realmLib.invokeGetBool(() => _realmLib.realm_rollback(realm.handle._pointer), "Could not rollback write");
   }
 
-  void realmRefresh(Realm realm) {
+  bool realmRefresh(Realm realm) {
     _realmLib.invokeGetBool(() => _realmLib.realm_refresh(realm.handle._pointer), "Could not refresh");
+    //TODO: Remove this and return the correct result of calling `realm_refresh`.
+    //Currently there is a bug in C-API that does not return the refresh result correctly.
+    //Issue https://github.com/realm/realm-core/pull/6068
```
nielsenko, line 771: Should we await this fix before merging?

lib/src/native/realm_core.dart
```diff
@@ -764,8 +764,35 @@ class _RealmCore {
...
+    return true;
+  }
+  
+  Future<void> realmRefreshAsync(Realm realm) {
```
nielsenko, line 775: Maybe not for this PR, as this is a pattern we have used all over `realm_core.dart`..

Anyway.. Why not make this 
```suggestion
  Future<void> realmRefreshAsync(Realm realm) async {
```
and await the `completer.future` at the end. That way we are sure that any exceptions raised by this functions before that point is correctly wrapped in a `Future.error(ex)`

desistefanova (approved):

test/realm_test.dart
```diff
@@ -1567,6 +1566,75 @@ Future<void> main([List<String>? args]) async {
...
+  });
+
+   test('Realm.onRefresh on frozen realm should be no-op', () async {
+    var realm = getRealm(Configuration.local([Person.schema]));
+    bool called = false;
+    realm = realm.freeze();
+
+    await Future.any([realm.refreshAsync().then((_) => called = true), Future<void>.delayed(Duration(milliseconds: 10))]);
+    
+    expect(called, true);
```
nielsenko, line 1636:
```suggestion
    var realm = getRealm(Configuration.local([Person.schema]));
    realm = realm.freeze();

    expect(realm.refreshAsync().timeout(const Duration(milliseconds: 10)), completes);
```

test/realm_test.dart
```diff
@@ -1567,6 +1566,75 @@ Future<void> main([List<String>? args]) async {
...
+
+
+    realm.refreshAsync().then((_) => called = true);
+
+    realm.write(() {});
+
+    await Future<void>.delayed(Duration(milliseconds: 1));
+
+    expect(called, true);
```
nielsenko, line 1626:
```suggestion
  final realm = getRealm(Configuration.local([Person.schema]));

    final first = expectLater(realm.refreshAsync(), completes);
    realm.beginWrite().commit();
    await (first);

    final second = expectLater(realm.refreshAsync(), completes);
    realm.write(() {});
    await (second);
```

test/realm_test.dart
nielsenko, line 1604:
```suggestion
    final realm = getRealm(Configuration.local([Person.schema]));
    final transaction = await realm.beginWriteAsync();
    expectLater(realm.refreshAsync(), completes);
    realm.add(Person("name"));
    await transaction.commitAsync();
```

test/realm_test.dart
nielsenko, line 1594:
```suggestion
    final realm = getRealm(Configuration.local([Person.schema]));
    final transaction = realm.beginWrite();
    final first = expectLater(realm.refreshAsync(), completes, reason: "beginWrite onRefresh failed");
    realm.add(Person("name"));
    transaction.commit();
    await first;

    realm.write(() {
      expectLater(realm.refreshAsync(), completes, reason: "write onRefresh failed");
      realm.add(Person("name"));
    });
```

nielsenko (changes requested):
Mostly test related suggestions

lib/src/native/realm_core.dart
blagoev: It seems we will anyway, but merging this without it also will work

lib/src/native/realm_core.dart
blagoev: I was debating to the same but for the public API which is the entry point. so `Realm.refreshAsync`. Let's discuss that and we can do it for all. But not in this PR :)

test/realm_test.dart
blagoev: I will use the suggestion

test/realm_test.dart
blagoev: This changes the test too much as it is not readable what is happening and what is the expectation. Lets leave it as I have written so it is more clear what goes after the other.

test/realm_test.dart
blagoev: As above

test/realm_test.dart
blagoev: as above

test/realm_test.dart
nielsenko: I believe it is clearer like I write it. It is also shorter. I would like us to stop using the flag based approach, and spread the knowledge of how to do it right.

test/realm_test.dart
```diff
@@ -1567,6 +1566,75 @@ Future<void> main([List<String>? args]) async {
...
+  });
+
+   test('Realm.onRefresh on frozen realm should be no-op', () async {
+    var realm = getRealm(Configuration.local([Person.schema]));
+    bool called = false;
+    realm = realm.freeze();
+
+    await realm.refreshAsync().timeout(const Duration(microseconds: 10));
+    
+    expect(called, true);
```
nielsenko, line 1636: I think you introduced an error here. It should be:
```suggestion
    realm = realm.freeze();

    expect(realm.refreshAsync().timeout(const Duration(milliseconds: 10)), completes);
```
you are not setting the flag anymore, and you are not expecting `completes`

test/realm_test.dart
nirinchev: I'm not sure what this test is supposed to test (in either of the forms). `realm.refreshAsync()` should complete synchronously when we're already at the latest version, which is pretty much always when using a single isolate. So it will have resolved already before we even get to `realm.write`. If we want to verify that it actually does something, we need to spin up a second isolate, open the same realm, write a bunch of data in it, then commit the transaction, signal the first isolate, then assert the data is not visible, call refresh async, assert that the data is visible.

blagoev: this test is the same as what we have in Core. Just using Dart

blagoev:
blocked on https://github.com/realm/realm-core/issues/6075

test/realm_test.dart
nielsenko: @nirinchev Currently calling `await realm.refreshAsync()` today will block until something is actually written to the realm. So it tests that.. which may well be a bug in the C-api.

nirinchev: Yeah, it's not the intended behavior though. There's no obvious situation in which this is going to be useful behavior.

nielsenko: I can see value in knowing when the next transaction completes, but I agree that is not what this API is intended for.

desistefanova: The test suggested by @nirinchev is added.

test/realm_test.dart
desistefanova: Outdated.

test/realm_test.dart
desistefanova: Outdated

test/realm_test.dart
desistefanova: Outdated.

lib/src/realm_class.dart
```diff
@@ -562,6 +562,24 @@ class Realm implements Finalizable {
 
     realmCore.writeCopy(this, config);
   }
+
+  /// Update the `Realm` instance and outstanding objects to point to the most recent persisted version.
+  ///
+  /// If another process or thread has made changes to the realm file, this causes
```
blagoev, line 568:
```suggestion
  /// If another process or [Isolate] has made changes to the realm file, this causes
```

lib/src/realm_class.dart
```diff
@@ -562,6 +562,24 @@ class Realm implements Finalizable {
...
+  /// those changes to become visible in this realm instance.
+  /// Typically you don't need to call this method since Realm has auto-refresh built-in.
+  /// Returns `true` only if there are changed data to refresh.
```
blagoev, line 571: This drastically differs from what .NET has for example. Here is the doc API .NET has currently
https://www.mongodb.com/docs/realm-sdks/dotnet/latest/reference/Realms.Realm.html#Realms_Realm_Refresh

> Note that this may return true even if no data has actually changed.

Also this is not what the native code is doing. Native code is returning `did_refresh` and it does not return `true` if any data actually changed

lib/src/realm_class.dart
```diff
@@ -562,6 +562,24 @@ class Realm implements Finalizable {
...
+  bool refresh() {
+    return realmCore.realmRefresh(this);
+  }
+
+  /// Returns a [Future<bool>] that will complete when the `Realm` is refreshed to the version which is the
+  /// latest version at the time when this method is called.
+  ///
+  /// Returns `true` only if there are changed data to refresh.
```
blagoev, line 579: same as above

nielsenko (commented):
I think we should consider dropping the synchronous version, and renaming the async version to just `refresh`

test/realm_test.dart
```diff
@@ -1756,6 +1755,125 @@ Future<void> main([List<String>? args]) async {
       });
     }
   }
+  test('Realm.refresh no changes', () async {
+    final realm = getRealm(Configuration.local([Person.schema]));
+    final result = realm.refresh();
+    expect(result, false);
+  });
+
+  test('Realm.refresh into sync transaction', () async {
```
blagoev, line 1764: what does it mean `into sync transaction`. This test name is not clear. Also it is named `Realm.refresh()` but executes `refreshAsync`
Should we rename to 
```suggestion
  test('Realm.refreshAsync() sync transaction', () async {
```

test/realm_test.dart
```diff
@@ -1756,6 +1755,125 @@ Future<void> main([List<String>? args]) async {
...
+    final transaction = realm.beginWrite();
+    realm.refreshAsync().then((refreshed) {
+      called = true;
+      isRefreshed = refreshed;
+    });
+    realm.add(Person("name"));
+    transaction.commit();
+
+    await Future<void>.delayed(Duration(milliseconds: 1));
+    expect(isRefreshed, false, reason: "There are no new data");
```
blagoev, line 1777:
```suggestion
    expect(isRefreshed, false, reason: "There is no new data");
```

blagoev: I think the reason should state why we have this expecations. So this should be 
"There should be no new data"

blagoev: In general adding a reason like this here is not adding much info. I see using a `reason` when the expectation is not clear enough

test/realm_test.dart
```diff
@@ -1756,6 +1755,125 @@ Future<void> main([List<String>? args]) async {
...
+    expect(called, true, reason: "Refresh called after transaction commit");
```
blagoev, line 1778:
```suggestion
    expect(called, true);
```

test/realm_test.dart
```diff
@@ -1756,6 +1755,125 @@ Future<void> main([List<String>? args]) async {
...
+    expect(realm.all<Person>().length, 1);
+  });
+
+  test('Realm.refresh into a write block', () async {
```
blagoev, line 1782:
```suggestion
  test('Realm.refreshAsync from within a write block', () async {
```

test/realm_test.dart
```diff
@@ -1756,6 +1755,125 @@ Future<void> main([List<String>? args]) async {
...
+    realm.write(() {
+      realm.refreshAsync().then((refreshed) {
+        called = true;
+        isRefreshed = refreshed;
+      });
+      realm.add(Person("name"));
+    });
+
+    await Future<void>.delayed(Duration(milliseconds: 1));
+    expect(isRefreshed, false, reason: "There are no new data");
```
blagoev, line 1795:
```suggestion
    expect(isRefreshed, false);
```

test/realm_test.dart
```diff
@@ -1756,6 +1755,125 @@ Future<void> main([List<String>? args]) async {
...
+    expect(called, true, reason: "Refresh called after write block completes");
```
blagoev, line 1796:
```suggestion
    expect(called, true);
```

test/realm_test.dart
```diff
@@ -1756,6 +1755,125 @@ Future<void> main([List<String>? args]) async {
...
+    expect(realm.all<Person>().length, 1);
+  });
+
+  test('Realm.refresh into async transaction', () async {
```
blagoev, line 1800:
```suggestion
  test('Realm.refreshAsync from within an async transaction', () async {
```

test/realm_test.dart
```diff
@@ -1756,6 +1755,125 @@ Future<void> main([List<String>? args]) async {
...
+    bool called = false;
+    bool isRefreshed = false;
+    final transaction = await realm.beginWriteAsync();
+    realm.refreshAsync().then((refreshed) {
+      called = true;
+      isRefreshed = refreshed;
+    });
+    realm.add(Person("name"));
+    await transaction.commitAsync();
+    expect(isRefreshed, false, reason: "There are no new data");
```
blagoev, line 1811: remove `reason`

test/realm_test.dart
```diff
@@ -1756,6 +1755,125 @@ Future<void> main([List<String>? args]) async {
...
+    expect(called, true, reason: "Refresh called after transaction commit");
```
blagoev, line 1812: remove `reason`

test/realm_test.dart
```diff
@@ -1756,6 +1755,125 @@ Future<void> main([List<String>? args]) async {
...
+    expect(realm.all<Person>().length, 1);
+  });
+
+  test('Realm.refresh is called if registered outside a transaction', () async {
```
blagoev, line 1816:
```suggestion
  test('Realm.refreshAsync outside a transaction', () async {
```

blagoev, line 1816: This test is not need imo. Or the name is really not explaining what we are testing in the test.

test/realm_test.dart
```diff
@@ -1756,6 +1755,125 @@ Future<void> main([List<String>? args]) async {
...
+    });
+
+    realm.write(() {});
+    await Future<void>.delayed(Duration(milliseconds: 1));
+
+    expect(isRefreshed, false, reason: "There are no new data");
+    expect(called, true);
+  });
+
+  test('Realm.refresh on frozen realm should be no-op', () async {
```
blagoev, line 1833:
```suggestion
  test('Realm.refreshAsync on frozen realm should be no-op', () async {
```

blagoev: why are we testing with refreshAsync in all tests? Why not use the sync version here?

test/realm_test.dart
```diff
@@ -1756,6 +1755,125 @@ Future<void> main([List<String>? args]) async {
...
+    var realm = getRealm(Configuration.local([Person.schema]));
+    bool called = false;
+    realm = realm.freeze();
+    final isRefreshed = await realm.refreshAsync();
+    expect(isRefreshed, false);
+  });
+
+  test('Realm.refresh is updating data', () async {
```
blagoev, line 1841:
```suggestion
  test('Realm.refresh', () async {
```

test/realm_test.dart
```diff
@@ -1756,6 +1755,125 @@ Future<void> main([List<String>? args]) async {
...
+    Completer<void> completer = Completer<void>();
```
blagoev, line 1842: I would move the completer creation close to where it is first used

test/realm_test.dart
```diff
@@ -1756,6 +1755,125 @@ Future<void> main([List<String>? args]) async {
...
+    final realm = getRealm(Configuration.local([Person.schema]));
+    String personName = generateRandomString(5);
+    final path = realm.config.path;
+    final results = realm.query<Person>(r"name == $0", [personName]);
+
+    expect(realm.refresh(), false);
+    realm.setAutoRefresh(false);
+
+    ReceivePort receivePort = ReceivePort();
```
blagoev, line 1851: This test seems unnecessary complex. Why are we sending booleans and waiting for 100 mils?

blagoev: I would expect this to work as well

```dart
 final externalRealm = Realm(Configuration.local([Person.schema], path: path));
      externalRealm.write(() {
        externalRealm.add(Person(personName));
      });
    externalRealm.close();
```

blagoev (commented):
[CHANGES REQUESTED]. 
I can't not make my own PRs RED in GH

test/realm_test.dart
```diff
@@ -1815,15 +1811,35 @@ Future<void> main([List<String>? args]) async {
...
   });
+
+  test('Realm.refresh is updating data', () async {
+    final realm = getRealm(Configuration.local([Person.schema]));
+    bool called = false;
+    String personName = generateRandomString(5);
+    final path = realm.config.path;
+    final results = realm.query<Person>(r"name == $0", [personName]);
+
+    await Isolate.run(() async {
```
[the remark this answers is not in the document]
desistefanova: I removed `Isolate.run` because it doesn't pass for MacOS where we are testing with Flutter 3.3.10.

lib/src/native/realm_core.dart
```diff
@@ -581,6 +581,10 @@ class _RealmCore {
     });
   }
 
+  void realmSetAutoRefresh(Realm realm, bool enable) {
```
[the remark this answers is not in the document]
desistefanova: It is by purpose. We may need to enable the auto refresh in some tests after it has been disabled.

test/realm_test.dart
desistefanova: You suggested to close the second realm after the refresh was called. That's why I close it a little bit later. If you think it is not need I can simplify the test.

test/realm_test.dart
desistefanova: I agree. Shall I remove it?

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
blagoev: We don't need to currently. Are you planning on adding more tests?

test/realm_test.dart
blagoev: I think it should not be needed if we disable autorefresh. My suggestion was how to workaround autorefresh. Which is not needed now

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
desistefanova: OK

test/realm_test.dart
desistefanova: Done.

lib/src/realm_class.dart
```diff
@@ -562,6 +562,24 @@ class Realm implements Finalizable {
...
+  /// Typically you don't need to call this method since Realm has auto-refresh built-in.
+  /// Note that this may return true even if no data has actually changed.
+  bool refresh() {
+    return realmCore.realmRefresh(this);
+  }
+
+  /// Returns a [Future] that will complete when the `Realm` is refreshed to the version which is the
+  /// latest version at the time when this method is called.
+  ///
+  /// Note that this may return true even if no data has actually changed.
```
blagoev, line 579:
```suggestion
  /// Note that this may return `true` even if no data has actually changed.
```

test/realm_test.dart
```diff
@@ -1756,6 +1755,95 @@ Future<void> main([List<String>? args]) async {
...
+    final realm = getRealm(Configuration.local([Person.schema]));
+    String personName = generateRandomString(5);
+    final path = realm.config.path;
+    final results = realm.query<Person>(r"name == $0", [personName]);
+
+    expect(realm.refresh(), false);
+    realm.disableAutoRefreshForTesting();
+
+    ReceivePort receivePort = ReceivePort();
+    Isolate.spawn((SendPort sendPort) async {
```
blagoev, line 1833: Can we use `Isolate.run` which returns a future with a result we can `await` here. Then we can remove the recieving port and the weird await for below?

blagoev (commented):
[SUGGESTED CHANGES]

nielsenko (approved):

Issue #684, opened 2022-06-23
Support AnyRealmValue


dotjon0:
Within MongoDB Atlas you can have schema validation that supports more then 1 x type for a field - i.e.:

```
multi_type_field: {
    title: `Multi Type Field`,
    bsonType: [ `date`, `bool`, `decimal`, `int`, `string` ],
  },
```

Within Dart/Flutter you can have the same effect by using the 'dynamic' type i.e.:

`final dynamic dynamicVariable = '123';`

We have also noticed Realm's Swift SDK supports the type **AnyRealmValue** which appears to support any type (Realm supports) (https://www.mongodb.com/docs/realm/sdk/swift/data-types/anyrealmvalue/#overview).

Of course in a high quality application (in an ideal world) every class property should have a specific type (i.e. not 'dynamic' type), however there will always be edge cases. Would the Realm team consider supporting 'dynamic' types for the classes/models that realm-dart auto generates with perhaps one of the approaches below (no 2 is our preference as it strictly declares what data types can be stored):

 **1. Declare the 'dynamic' type** 

```
@RealmModel()
class _Car {

    late dynamic make;

    late String model;

    int? kilometers = 500;
  }
```
  
OR

**2. Declare the 'dynamic' type & specify the multiple types accepted**

Use json_annotation package https://pub.dev/packages/json_annotation via '@JsonKey(' or similar to pass in the bsonTypes accepted - if the incorrect type is handed into Realm local DB, throwing an exception would be amazing, but this may reach too far for Realm local DB remit:

```
@RealmModel()
class _Car {

    @JsonKey( bsonType: [ `date`, `bool`, `decimal`, `int`, `string` ] )
    late dynamic make;

    late String model;

    int? kilometers = 500;
 }
```

The use case for us is that we have some models that are flexible on the data type we are storing so the user can customise their instance of the application.

Looking forward to any thoughts!

nielsenko:
@dotjon0 Our policy prevents me from talking about timelines. Obviously we are heading into summer vacation territory, so I would be surprised if it happens this side of summer.

dotjon0:
@nielsenko complete get that, we are the same - thanks for the heads up re summer its appreciated!

dotjon0:
Just wondering if there have been any advances on this?

nielsenko:
@dotjon0 I know that the kotlin team was working through some issues in the realm-core C-API on this as late as today. I assume it will be one of the next thing on our agenda.

[...]

PR #1051, opened 2022-11-30
Support RealmValue (aka mixed)

nielsenko:
Support `RealmValue` that allows users to store different types in the same field
```dart
@RealmModel()
class _AnythingGoes {
  @Indexed()
  late RealmValue any;
  late List<RealmValue> manyAny;
}

void main() {
  final realm = Realm(Configuration.local([AnythingGoes.schema]));

  final something = realm.write(() => realm.add(AnythingGoes(any: RealmValue.string('text'))));

  realm.write(() {
    something.manyAny.addAll([null, true, 'text', 42, 3.14].map(RealmValue.from));
  });
}
```
Resolves: #684

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 * Add List.move extension method that moves an element from one index to another. Delegates to ManagedRealmList.move for managed lists. This allows notifications to correctly report moves, as opposed to reporting moves as deletes + inserts. ([#1037](https://github.com/realm/realm-dart/issues/1037))
 * Support setting `shouldDeleteIfMigrationNeeded` when creating a `Configuration.local`. ([#1049](https://github.com/realm/realm-dart/issues/1049))
 * Add `unknown` error code to all SyncErrors: `SyncSessionErrorCode.unknown`, `SyncConnectionErrorCode.unknown`, `SyncClientErrorCode.unknown`, `GeneralSyncErrorCode.unknown`. Use `unknown` error code instead of throwing a RealmError. ([#1052](https://github.com/realm/realm-dart/pull/1052))
-* Support `RealmValue` that allows users to store different types in the same field. ([#1051](https://github.com/realm/realm-dart/pull/1051))
+* Add support for `RealmValue` data type. This new type can represent any valid Realm data type, including objects. Lists of `RealmValue` are also supported, but `RealmValue` itself cannot contain collections. Please note that a property of type `RealmValue` cannot be nullable, but can contain null, represented by the value `RealmValue.nullValue()`. ([#1051](https://github.com/realm/realm-dart/pull/1051))
 
 ### Fixed
 * Support mapping into `SyncSessionErrorCode` for "Compensating write" with error code 231. ([#1022](https://github.com/realm/realm-dart/pull/1022))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Add List.move extension method that moves an element from one index to another. Delegates to ManagedRealmList.move for managed lists. This allows notifications to correctly report moves, as opposed to reporting moves as deletes + inserts. ([#1037](https://github.com/realm/realm-dart/issues/1037))
 * Support setting `shouldDeleteIfMigrationNeeded` when creating a `Configuration.local`. ([#1049](https://github.com/realm/realm-dart/issues/1049))
 * Add `unknown` error code to all SyncErrors: `SyncSessionErrorCode.unknown`, `SyncConnectionErrorCode.unknown`, `SyncClientErrorCode.unknown`, `GeneralSyncErrorCode.unknown`. Use `unknown` error code instead of throwing a RealmError. ([#1052](https://github.com/realm/realm-dart/pull/1052))
+* Support `RealmValue` that allows users to store different types in the same field. ([#1051](https://github.com/realm/realm-dart/pull/1051))
 
 ### Fixed
 * Support mapping into `SyncSessionErrorCode` for "Compensating write" with error code 231. ([#1022](https://github.com/realm/realm-dart/pull/1022))
--- a/common/lib/src/realm_types.dart
+++ b/common/lib/src/realm_types.dart
@@ -131,8 +131,26 @@ class RealmValue {
   const RealmValue.realmObject(RealmObjectBaseMarker o) : this._(o);
   const RealmValue.dateTime(DateTime timestamp) : this._(timestamp);
   const RealmValue.objectId(ObjectId id) : this._(id);
-  const RealmValue.decimal128(Decimal128 decimal) : this._(decimal);
+  // const RealmValue.decimal128(Decimal128 decimal) : this._(decimal); // Not supported yet
   const RealmValue.uuid(Uuid uuid) : this._(uuid);
+
+  factory RealmValue.from(Object o) {
+    if (o is bool ||
+        o is String ||
+        o is int ||
+        o is Float ||
+        o is double ||
+        o is Uint8List ||
+        o is RealmObjectBaseMarker ||
+        o is DateTime ||
+        o is ObjectId ||
+        // o is Decimal128 || // not supported yet
+        o is Uuid) {
+      return RealmValue._(o);
+    } else {
+      throw ArgumentError.value(o, 'o', 'Unsupported type');
+    }
+  }
 }
 
 /// The category of a [SyncError].
--- a/common/lib/src/realm_types.dart
+++ b/common/lib/src/realm_types.dart
@@ -46,7 +46,7 @@ enum RealmPropertyType {
   _3, // ignore: unused_field, constant_identifier_names
   binary,
   _5, // ignore: unused_field, constant_identifier_names
-  mixed,
+  mixed(Mapping<RealmValue>(indexable: true)),
   _7, // ignore: unused_field, constant_identifier_names
   timestamp(Mapping<DateTime>(indexable: true)),
   float,
@@ -114,25 +114,25 @@ class EmbeddedObjectMarker extends RealmObjectBaseMarker {}
 
 // Union type
 /// @nodoc
-class RealmAny {
-  final dynamic value;
+class RealmValue {
+  final Object value;
   T as<T>() => value as T; // better for code completion
 
   // This is private, so user cannot accidentally construct an invalid instance
-  const RealmAny._(this.value);
+  const RealmValue._(this.value);
 
-  const RealmAny.bool(bool b) : this._(b);
-  const RealmAny.string(String text) : this._(text);
-  const RealmAny.int(int i) : this._(i);
-  const RealmAny.float(Float f) : this._(f);
-  const RealmAny.double(double d) : this._(d);
-  const RealmAny.uint8List(Uint8List data) : this._(data);
+  const RealmValue.bool(bool b) : this._(b);
+  const RealmValue.string(String text) : this._(text);
+  const RealmValue.int(int i) : this._(i);
+  const RealmValue.float(Float f) : this._(f);
+  const RealmValue.double(double d) : this._(d);
+  const RealmValue.uint8List(Uint8List data) : this._(data);
   // TODO: RealmObjectMarker introduced to avoid dependency inversion. It would be better if we could use RealmObject directly. https://github.com/realm/realm-dart/issues/701
-  const RealmAny.realmObject(RealmObjectBaseMarker o) : this._(o);
-  const RealmAny.dateTime(DateTime timestamp) : this._(timestamp);
-  const RealmAny.objectId(ObjectId id) : this._(id);
-  const RealmAny.decimal128(Decimal128 decimal) : this._(decimal);
-  const RealmAny.uuid(Uuid uuid) : this._(uuid);
+  const RealmValue.realmObject(RealmObjectBaseMarker o) : this._(o);
+  const RealmValue.dateTime(DateTime timestamp) : this._(timestamp);
+  const RealmValue.objectId(ObjectId id) : this._(id);
+  const RealmValue.decimal128(Decimal128 decimal) : this._(decimal);
+  const RealmValue.uuid(Uuid uuid) : this._(uuid);
 }
 
 /// The category of a [SyncError].
--- a/common/lib/src/realm_types.dart
+++ b/common/lib/src/realm_types.dart
@@ -111,9 +111,6 @@ abstract class RealmObjectMarker implements RealmObjectBaseMarker {}
 /// @nodoc
 abstract class EmbeddedObjectMarker implements RealmObjectBaseMarker {}
 
-/// @nodoc
-abstract class AsymmetricObjectMarker implements RealmObjectBaseMarker {}
-
 /// A type that can represent any valid realm data type, except collections and embedded objects.
 ///
 /// You can use [RealmValue] to declare fields on realm models, in which case it must be non-nullable,
--- a/common/lib/src/realm_types.dart
+++ b/common/lib/src/realm_types.dart
@@ -115,12 +115,13 @@ class EmbeddedObjectMarker extends RealmObjectBaseMarker {}
 // Union type
 /// @nodoc
 class RealmValue {
-  final Object value;
+  final Object? value;
   T as<T>() => value as T; // better for code completion
 
   // This is private, so user cannot accidentally construct an invalid instance
   const RealmValue._(this.value);
 
+  const RealmValue.nullValue() : this._(null);
   const RealmValue.bool(bool b) : this._(b);
   const RealmValue.string(String text) : this._(text);
   const RealmValue.int(int i) : this._(i);
@@ -134,8 +135,9 @@ class RealmValue {
   // const RealmValue.decimal128(Decimal128 decimal) : this._(decimal); // Not supported yet
   const RealmValue.uuid(Uuid uuid) : this._(uuid);
 
-  factory RealmValue.from(Object o) {
-    if (o is bool ||
+  factory RealmValue.from(Object? o) {
+    if (o == null ||
+        o is bool ||
         o is String ||
         o is int ||
         o is Float ||
@@ -151,6 +153,18 @@ class RealmValue {
       throw ArgumentError.value(o, 'o', 'Unsupported type');
     }
   }
+
+  @override
+  operator ==(Object? other) {
+    if (other is RealmValue) {
+      return value == other.value;
+    } else {
+      return value == other;
+    }
+  }
+
+  @override
+  int get hashCode => value.hashCode;
 }
 
 /// The category of a [SyncError].
--- a/common/lib/src/realm_types.dart
+++ b/common/lib/src/realm_types.dart
@@ -125,9 +125,8 @@ class RealmValue {
   const RealmValue.bool(bool b) : this._(b);
   const RealmValue.string(String text) : this._(text);
   const RealmValue.int(int i) : this._(i);
-  const RealmValue.float(Float f) : this._(f);
   const RealmValue.double(double d) : this._(d);
-  const RealmValue.uint8List(Uint8List data) : this._(data);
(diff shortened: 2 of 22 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * Add List.move extension method that moves an element from one index to another. Delegates to ManagedRealmList.move for managed lists. This allows notifications to correctly report moves, as opposed to reporting moves as deletes + inserts. ([#1037](https://github.com/realm/realm-dart/issues/1037))
 * Support setting `shouldDeleteIfMigrationNeeded` when creating a `Configuration.local`. ([#1049](https://github.com/realm/realm-dart/issues/1049))
 * Add `unknown` error code to all SyncErrors: `SyncSessionErrorCode.unknown`, `SyncConnectionErrorCode.unknown`, `SyncClientErrorCode.unknown`, `GeneralSyncErrorCode.unknown`. Use `unknown` error code instead of throwing a RealmError. ([#1052](https://github.com/realm/realm-dart/pull/1052))
+* Support `RealmValue` that allows users to store different types in the same field. ([#1051](https://github.com/realm/realm-dart/pull/1051))
```
blagoev, line 15: It would be better if we add more details here like what .NET did https://github.com/realm/realm-dotnet/blob/main/CHANGELOG.md?plain=1#L825

common/lib/src/realm_types.dart
```diff
@@ -104,35 +103,60 @@ class RealmStateError extends StateError implements RealmError {
...
-
-/// @nodoc
-class EmbeddedObjectMarker extends RealmObjectBaseMarker {}
+abstract class RealmObjectBaseMarker {}
 
 // Union type
 /// @nodoc
-class RealmAny {
-  final dynamic value;
+class RealmValue {
```
blagoev, line 110: should we have an API doc here?

common/lib/src/realm_types.dart
```diff
@@ -104,35 +103,60 @@ class RealmStateError extends StateError implements RealmError {
...
+      throw ArgumentError.value(o, 'o', 'Unsupported type');
+    }
+  }
+
+  @override
+  operator ==(Object? other) {
+    if (other is RealmValue) {
+      return value == other.value;
+    } else {
+      return value == other;
```
blagoev, line 154: why not return `false` here. In what case this can be `true`

common/lib/src/realm_types.dart
```diff
@@ -104,35 +103,60 @@ class RealmStateError extends StateError implements RealmError {
...
 
 /// @nodoc
-class RealmObjectBaseMarker {}
-
-/// @nodoc
-class RealmObjectMarker extends RealmObjectBaseMarker {}
-
-/// @nodoc
-class EmbeddedObjectMarker extends RealmObjectBaseMarker {}
+abstract class RealmObjectBaseMarker {}
```
blagoev, line 106: Since we are removing the two marker classes should we rename to `RealmObjectMarker` instead. Since there is no hierarchy now

blagoev (approved):
Looks good as initial version, 
Added some comments. 
Also it would be good to check what additional tests .NET has and implement here. I think they have more elaborate tests like using embedded objects etc.

common/lib/src/realm_types.dart
nielsenko: This allows you to write `RealmValue.int(1) == 1` and get `true`

blagoev: got it

CHANGELOG.md

common/lib/src/realm_types.dart
nielsenko: Fixed

common/lib/src/realm_types.dart
nielsenko: Fixed for class. Will add for props as well

generator/lib/src/realm_model_info.dart
```diff
@@ -48,10 +48,13 @@ class RealmModelInfo {
...
 
         final notRequired = allSettable.where((f) => !f.isRequired && !f.isPrimaryKey);
-        final collections = fields.where((f) => f.type.isRealmCollection).toList();
+        final collections = fields.where((f) => f.isRealmCollection).toList();
         if (notRequired.isNotEmpty || collections.isNotEmpty) {
           yield '{';
-          yield* notRequired.map((f) => '${f.mappedTypeName} ${f.name}${f.hasDefaultValue ? ' = ${f.fieldElement.initializerExpression}' : ''},');
+          yield* notRequired.map(
+            (f) =>
+                '${f.mappedTypeName} ${f.name}${f.hasDefaultValue ? ' = ${f.fieldElement.initializerExpression}' : (f.isMixed ? ' = const RealmValue.nullValue()' : '')},',
```
nirinchev, line 56: That's kinda convoluted. Can we make a few adjustments to `RealmFieldInfo` to accommodate mixed support? E.g. we can change things to:

```dart
class RealmFieldInfo {
  // ...
  bool get hasDefaultValue => fieldElement.hasInitializer || isMixed;

  String? get defaultValueExpression {
    if (fieldElement.hasInitializer) {
      return fieldElement.initializerExpression.toString();
    }

    if (isMixed) {
      return 'const RealmValue.nullValue()';
    }

    return null;
  }
}
```

Then this will simplify to:

```dart
yield* notRequired.map((f) => '${f.mappedTypeName} ${f.name}${f.hasDefaultValue ? ' = ${f.defaultValueExpression}' : ''},');
```

common/lib/src/realm_types.dart
```diff
@@ -104,35 +103,87 @@ class RealmStateError extends StateError implements RealmError {
...
+///       'text',
+///       42,
+///       3.14,
+///     ].map(RealmValue.from));
+///   });
+/// }
+/// ```
+class RealmValue {
+  final Object? value;
+  Type get type => value.runtimeType;
```
nirinchev, line 139: What if `value` is `null`?

lib/src/list.dart
```diff
@@ -111,12 +111,23 @@ class ManagedRealmList<T extends Object?> with RealmEntity, ListMixin<T> impleme
...
-
+      var value = realmCore.listGetElementAt(this, index);
       if (value is RealmObjectHandle) {
-        return realm.createObject(T, value, _metadata!) as T;
+        late RealmObjectMetadata targetMetadata;
+        late Type type;
+        if (T == RealmValue) {
+          final tuple = realm.metadata.getByClassKey(realmCore.getClassKey(value));
+          type = tuple.item1;
+          targetMetadata = tuple.item2;
```
nirinchev, line 121: I don't think this handles the case where you open a Realm with an incomplete schema that contains `RealmValue` properties linking to tables not in the user schema.

common/lib/src/realm_types.dart
```diff
@@ -104,35 +103,87 @@ class RealmStateError extends StateError implements RealmError {
...
+  operator ==(Object? other) {
+    if (other is RealmValue) {
+      return value == other.value;
+    } else {
+      return value == other;
+    }
+  }
+
+  @override
+  int get hashCode => value.hashCode;
```
nirinchev, line 186: Should we override `toString` to give this a nicer string representation?

common/lib/src/realm_types.dart
```diff
@@ -104,35 +103,87 @@ class RealmStateError extends StateError implements RealmError {
...
+    } else {
+      throw ArgumentError.value(o, 'o', 'Unsupported type');
+    }
+  }
+
+  @override
+  operator ==(Object? other) {
+    if (other is RealmValue) {
+      return value == other.value;
+    } else {
```
nirinchev, line 180: [nit] We can drop the else, but not a dealbreaker - I know some people prefer this style.

nirinchev (changes requested):
We probably need to expose `ObjectType` similarly to what we're doing in C#.

test/realm_value_test.dart
```diff
@@ -0,0 +1,158 @@
...
+
+import 'test.dart';
+
+part 'realm_value_test.g.dart';
+
+@RealmModel()
+class _AnythingGoes {
+  @Indexed()
+  late RealmValue oneAny;
+  late List<RealmValue> manyAny;
```
desistefanova, line 32:
```suggestion
late RealmValue oneValue;
late List<RealmValue> manyValues;
```
Maybe will be better if you rename all the places where you have members `any` to `value`. 'Any' is confusing to me.

test/realm_value_test.dart
```diff
@@ -0,0 +1,158 @@
...
+      'text',
+      42,
+      3.14,
+      AnythingGoes(),
+      now,
+      ObjectId.fromTimestamp(now),
+      Uuid.v4(),
+    ];
+
+    final config = Configuration.inMemory([AnythingGoes.schema]);
```
desistefanova, line 148: We are testing only with inMemory Realm. Is that enough?

test/realm_value_test.dart
```diff
@@ -0,0 +1,158 @@
...
+    final realm = getRealm(config);
+
+    test('Roundtrip', () {
+      final something = realm.write(() => realm.add(AnythingGoes(manyAny: values.map(RealmValue.from))));
+      expect(something.manyAny.map((e) => e.value), values);
+      expect(something.manyAny, values.map(RealmValue.from));
+//      expect(something.manyAny.cast<Object>(), [true, 42]);
```
desistefanova, line 155: This comment maybe has to be deleted.

test/realm_value_test.dart
```diff
@@ -0,0 +1,158 @@
...
+part 'realm_value_test.g.dart';
+
+@RealmModel()
+class _AnythingGoes {
+  @Indexed()
+  late RealmValue oneAny;
+  late List<RealmValue> manyAny;
+}
+
+void main() {
```
desistefanova, line 35: Will be good to make more tests for example:
- Query by RealmValue;
- Frozen objects with RealmValue;
- Embedded objects with RealmValue;
- Flexible sync with RealmValue;
- Subscriptions by RealmValue;

generator/lib/src/field_element_ex.dart
```diff
@@ -134,12 +133,12 @@ extension FieldElementEx on FieldElement {
 
       // Validate indexes
       if ((((primaryKey ?? indexed) != null) && !(type.realmType?.mapping.indexable ?? false)) || //
-          (primaryKey != null && type.isDartCoreBool)) {
+          (primaryKey != null && (type.isDartCoreBool || type.isExactly<RealmValue>()))) {
```
desistefanova, line 136: Lets add tests in the generator for @Indexed and @PrimaryKey annotations on a RealmValue member.

common/lib/src/realm_types.dart
nielsenko: `runtimeType` is a method on `Object?`, it is safe to call on `null`
```dart
void main() {
  print(null.runtimeType);
}
```
prints: `Null` which is the runtime type of a `null` expression

generator/test/error_test_data/nullable_realm_value.expected
```diff
@@ -0,0 +1,12 @@
+RealmValue fields cannot be nullable
+
+in: asset:pkg/test/error_test_data/nullable_realm_value.dart:7:3
+  ╷
+5 │ @RealmModel()
+6 │ class _Bad {
+  │       ━━━━ in realm model for 'Bad'
+7 │   RealmValue? wrong;
+  │   ^^^^^^^^^^^ RealmValue? is nullable
```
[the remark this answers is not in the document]
nielsenko: In general I agree that we should have long descriptive error messages, but we need to agree on the team. Previously I was asked to simplify them.

generator/lib/src/realm_model_info.dart
nielsenko: I will rework this to make intent more clear

test/realm_value_test.dart
nielsenko: Yes, I don't see a reason test with other configurations.

test/realm_value_test.dart
nielsenko: This would test realm-core more than realm-dart.

test/realm_value_test.dart
nielsenko: > We probably need to expose `ObjectType` similarly to what we're doing in C#.

I will re-work to disallow non top-level objects by type in RealmValue. This will take a bit of time.

nielsenko: Much easier than anticipated. We now prevent storing embedded objects in a `RealmValue` as they implement `EmbeddedObjectMarker` and not `TopLevelObjectMarker`

generator/lib/src/field_element_ex.dart
nielsenko: Tests added

nielsenko:
> We probably need to expose `ObjectType` similarly to what we're doing in C#.

I assume this is catch the error case that you cannot store embedded objects in a `RealmValue` (or asymmetric objects for that matter, when we eventually support those).

I have reworked the code to catch this misuse in the type system.

test/realm_value_test.dart
nielsenko: .. hmm these comments are misplaced.. will delete

lib/src/list.dart
nirinchev: We should return dynamic objects.

generator/test/error_test_data/nullable_realm_value.expected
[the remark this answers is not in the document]
blagoev: I would vote for short and clear messages. We could reword this to be more clear, but long error messages stand in the way after a while. So imo, this could be 
```
`RealmValue?` is declared nullable. Use `RealmValue` or `RealmValue myfield = RealmValue.nullValue()` to assign a default null value.
```

common/lib/src/realm_types.dart
```diff
@@ -105,6 +105,15 @@ class Decimal128 {} // TODO Support decimal128 datatype https://github.com/realm
...
 abstract class RealmObjectMarker {}
 
+/// @nodoc
+abstract class TopLevelObjectMarker implements RealmObjectMarker {}
+
+/// @nodoc
+abstract class EmbeddedObjectMarker implements RealmObjectMarker {}
+
+/// @nodoc
+abstract class AsymmetricObjectMarker implements RealmObjectMarker {}
```
blagoev, line 115: If this is not used I suggest not to introduce it here. or comment it out. Generally anticipating code changes is not good practice imo, since we might end up not needing that when we get to implementing it. This was the case with `RealmObjectBaseMarker`. 
Why not revert the change to `RealmObjectBaseMarker` since it seems we need it.

blagoev: So in a sense I want to say that I prefer the `RealmObjectBaseMarker` and `RealmIObjectMarker` than naming this `ToplLevelMarker` which is a new concept which is not clear to me.

blagoev (approved):

test/realm_value_test.dart
nirinchev: Some of these have the potential to surface symmetric bugs and may be valuable. Let's say you make a mistake and you store double values as negative the original value (e.g. `4.56 -> RealmValue(-4.56)`). This kind of bug will always pass a test like:

```
final value = 4.56;
realm.write(() {
  myObj.any = value;
});

expect(myObj.any, value);
```

Ideally, that should be tested by serializing the object to json and comparing the stringified representation as seen by Core, but those other tests may also be valuable. While I agree it's testing Core, historically, SDKs have been the primary source of bug reports for the Core team and it doesn't hurt to have a few extra tests.

blagoev:
@nirinchev I couldn't get your meaning for this one
```
We probably need to expose ObjectType similarly to what we're doing in C#.
```
Could you clarify this. Especially important if it is a reason to block a PR.

common/lib/src/realm_types.dart
nielsenko: Okay, was just following core
```c++
enum class ObjectType : uint8_t { TopLevel = 0, Embedded = 0x1, TopLevelAsymmetric = 0x2 };
```
I will rename `RealmObjectMarker` to `RealmObjectBaseMarker` and `TopLevelObjectMarker` as `RealmObjectMarker`

generator/lib/src/realm_model_info.dart
nielsenko: Simplified

common/lib/src/realm_types.dart
nielsenko: Fixed

common/lib/src/realm_types.dart
nielsenko: Fixed

nirinchev:
Okay, I had forgotten some things in dart, so I believe there's more work needed to support RealmValue, but that can probably be split into a different PR. Here's an example of an issue that we can't handle today. Let's say you have a synchronized Realm with the following schema (this is v1 of your app).

```dart
class MyDynamic {
  RealmValue value;
}

class Person {
  String name;
}
```

In v2, you add a new type - `Dog`. Now, a new client comes in and adds a `Dog` instance as a value for `MyDynamic.value`. A client in v1 of your app will have no way to do something meaningful with that object as it won't have the schema for `Dog` and there's no way to inspect what properties exist. In C# and other SDKs we expose the object schema on RealmObject instances, so you could do something like:

```dart
final obj = myDynamic.value.as<RealmObjectBase>();
switch (obj.schema.name) {
  case "Person": // do something
  default:
    final nameProp = obj.schema.properties.firstOrNull(p => p.name == "name");
    if (nameProp != null && nameProp.propertyType == RealmPropertyType.string) {
      // do something with the name
    }
}
```

My thinking with exposing `ObjectType` on `RealmValue` was to support use cases like:

```dart
public String getDisplayName(RealmValue value) {
  switch (value.objectType) {
    case null:
      // value is not an object or is null - do nothing
	  return "unknown";
    case "Person":
      final person = value.as<Person>();
      return '${person.firstName} ${person.lastName}';
    case "Company":
      final company = value.as<Company>();
      return company.legalEntity.name;
    default:
      final obj = value.as<RealmObjectBase>();
      final nameProp = obj.schema.properties.firstOrNull(p => p.name == "name");
      if (nameProp != null || nameProp.propertyType == 'string') {
        return obj.dynamic.get<string>('name');
      }

      return "unknown";
  }
}
```

But again, this depends on the schema being available on the realm object.

test/realm_value_test.dart
nielsenko: As the 6th SDK I see less value in tests that essentially test realm-core. Adding them now will delay this PR, and require maintenance down the line.

generator/test/error_test_data/nullable_realm_value.expected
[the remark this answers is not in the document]
nielsenko: I don't think it gets in the way. The short version is at the top, the rest you can read if you are confused.

lib/src/list.dart
nielsenko: Hmm .. added a test, but it crashes in
```dart
realmCore.getProperty(object, propertyMeta.key);
```
At least the c-api don't support this

test/realm_value_test.dart
nielsenko: > Okay, I had forgotten some things in dart, so I believe there's more work needed to support RealmValue, but that can probably be split into a different PR...

I will just comment that even trying to read a mixed property from core containing a realm object with unknown schema will crash, and since the property is mixed we really have no way to detect this, so the crash is inevitable 😢

test/realm_value_test.dart
nielsenko: .. an argument can be made that we are only the 2nd SDK to use the c-api though

desistefanova (approved):

nirinchev (approved):
As discussed, I think it's fine to ship the feature as is while the C API limitation is addressed. I'd still recommend adding a json verification for the roundtrip test to protect us from symmetric bugs.

PR #1052, opened 2022-11-30
Set missing SyncError codes to Unknown

### LINKED ISSUE #1041: Set missing SyncError codes to Unknown
Instead of throwing an Exception when the Error Category is known but the error code is unknown it is better to set the code to Unknown.

[https://github.com/realm/realm-dart/blob/64cde9f57795a1d3ea39985a2550c2303a9cac77/lib/src/configuration.dart#L746](https://github.com/realm/realm-dart/blob/64cde9f57795a1d3ea39985a2550c2303a9cac77/lib/src/configuration.dart#L746)

[https://github.com/realm/realm-dart/blob/64cde9f57795a1d3ea39985a2550c2303a9cac77/common/lib/src/realm_types.dart#L326](https://github.com/realm/realm-dart/blob/64cde9f57795a1d3ea39985a2550c2303a9cac77/common/lib/src/realm_types.dart#L326)

Check also all the other error enums.

desistefanova:
Fixes #1041

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 * Support setting `maxNumberOfActiveVersions` when creating a `Configuration`. ([#1036](https://github.com/realm/realm-dart/pull/1036))
 * Add List.move extension method that moves an element from one index to another. Delegates to ManagedRealmList.move for managed lists. This allows notifications to correctly report moves, as opposed to reporting moves as deletes + inserts. ([#1037](https://github.com/realm/realm-dart/issues/1037))
 * Support setting `shouldDeleteIfMigrationNeeded` when creating a `Configuration.local`. ([#1049](https://github.com/realm/realm-dart/issues/1049))
-* Add `unknown` error code to all SyncErrors. `SyncSessionErrorCode.unknown`, `SyncConnectionErrorCode.unknown`, `SyncClientErrorCode.unknown` or `GeneralSyncErrorCode.unknown`. Use `unknown` error code instead of throwing a RealmError. ([#1052](https://github.com/realm/realm-dart/pull/1052))
+* Add `unknown` error code to all SyncErrors: `SyncSessionErrorCode.unknown`, `SyncConnectionErrorCode.unknown`, `SyncClientErrorCode.unknown`, `GeneralSyncErrorCode.unknown`. Use `unknown` error code instead of throwing a RealmError. ([#1052](https://github.com/realm/realm-dart/pull/1052))
 
 ### Fixed
 * Support mapping into `SyncSessionErrorCode` for "Compensating write" with error code 231. ([#1022](https://github.com/realm/realm-dart/pull/1022))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 * Support setting `maxNumberOfActiveVersions` when creating a `Configuration`. ([#1036](https://github.com/realm/realm-dart/pull/1036))
 * Add List.move extension method that moves an element from one index to another. Delegates to ManagedRealmList.move for managed lists. This allows notifications to correctly report moves, as opposed to reporting moves as deletes + inserts. ([#1037](https://github.com/realm/realm-dart/issues/1037))
 * Support setting `shouldDeleteIfMigrationNeeded` when creating a `Configuration.local`. ([#1049](https://github.com/realm/realm-dart/issues/1049))
+* If `syncErrorHandler` occurs with a new unknown error code then return `SyncSessionErrorCode.unknown`, `SyncConnectionErrorCode.unknown`, `SyncClientErrorCode.unknown` or `GeneralSyncErrorCode.unknown` instead of throwing exception. ([#1052](https://github.com/realm/realm-dart/pull/1052))
```
blagoev, line 13: What does it mean for a `syncErrorHandler to occur`.

I suggest we reword this to something like 
```suggestion
* Add `unknown` error code to all SyncErrors. `SyncSessionErrorCode.unknown`, `SyncConnectionErrorCode.unknown`, `SyncClientErrorCode.unknown` or `GeneralSyncErrorCode.unknown`. Use `unknown` error code instead of throwing a RealmError. ([#1052](https://github.com/realm/realm-dart/pull/1052))
```

common/lib/src/realm_types.dart
```diff
@@ -316,17 +313,14 @@ enum SyncConnectionErrorCode {
   switchToFlxSync(113),
 
   /// Connected with wrong wire protocol - should switch to PBS
-  switchToPbs(114);
+  switchToPbs(114),
+
+  unknown(9999);
```
blagoev, line 318: no api doc

blagoev: and all others are missing api docs

lib/src/configuration.dart
```diff
@@ -252,20 +252,19 @@ abstract class Configuration implements Finalizable {
...
-    super.encryptionKey,
-    this.disableFormatUpgrade = false,
-    this.isReadOnly = false,
-    this.shouldCompactCallback,
-    this.migrationCallback,
-    super.maxNumberOfActiveVersions,
-    this.shouldDeleteIfMigrationNeeded = false
-  }) : super._();
+  LocalConfiguration._(super.schemaObjects,
+      {this.initialDataCallback,
```
blagoev, line 256: this seems not formatted correctly?

blagoev (changes requested):

lib/src/configuration.dart
nirinchev: The formatter started changing these for me as well - not sure what happened. I do prefer the old style, but it's also annoying to always revert back to it.

common/lib/src/realm_types.dart
desistefanova: What should it be. "Unknown error code" ?

lib/src/configuration.dart
desistefanova: It is already required to have a "," after the last parameter to order it as it was before.

common/lib/src/realm_types.dart
```diff
@@ -316,17 +314,15 @@ enum SyncConnectionErrorCode {
   switchToFlxSync(113),
 
   /// Connected with wrong wire protocol - should switch to PBS
-  switchToPbs(114);
+  switchToPbs(114),
+
+  /// Unknown error code
```
blagoev, line 319:
```suggestion
  /// Unknown Sync connection error code
```

blagoev: instead of generic api doc, we could mention the category here. like the suggestion above. I suggest we do this for each sync error type

blagoev (changes requested):

blagoev (approved):

nielsenko (approved):
LGTM

Issue #1070, opened 2022-12-18
[Bug]: Can't generate code


ebelevics:
### What happened?

After generating code I get this error. Everything was working until upgrading flutter to 3.3.10.
`dependency_overrides: analyzer: ^4.7.0` would not work any more with never flutter version, but upgrading to `dependency_overrides: analyzer: ^5.2.0` causes this exception, and don't know what to do now

### Repro steps

```
% flutter pub run build_runner build --delete-conflicting-outputs
[INFO] Generating build script...
[INFO] Generating build script completed, took 192ms

[INFO] Precompiling build script......
[WARNING] ../../FlutterSDK/flutter/.pub-cache/hosted/pub.dartlang.org/realm_generator-0.8.0+rc/lib/src/pseudo_type.dart:13:7: Error: The non-abstract class 'PseudoType' is missing implementations for these members:
 - DartType.element
Try to either
 - provide an implementation,
 - inherit an implementation from a superclass or mixin,
 - mark the class as abstract, or
 - provide a 'noSuchMethod' implementation.

class PseudoType extends TypeImpl {
      ^^^^^^^^^^
../../FlutterSDK/flutter/.pub-cache/hosted/pub.dartlang.org/analyzer-5.2.0/lib/dart/element/type.dart:38:16: Context: 'DartType.element' is defined here.
  Element? get element;
               ^^^^^^^
../../FlutterSDK/flutter/.pub-cache/hosted/pub.dartlang.org/realm_generator-0.8.0+rc/lib/src/pseudo_type.dart:18:84: Error: Too many positional arguments: 0 allowed, but 1 found.
Try removing the extra positional arguments.
  PseudoType(this._name, {this.nullabilitySuffix = NullabilitySuffix.none}) : super(null);
                                                                                   ^
[INFO] Precompiling build script... completed, took 413ms

[SEVERE] Failed to precompile build script .dart_tool/build/entrypoint/build.dart.
This is likely caused by a misconfigured builder definition.
```

### Version

Flutter 3.3.10 / Dart 2.18.6 

### What Realm SDK flavor are you using?

Local Database only

### What type of application is this?

Flutter Application

### Client OS and version

macOS Monterey 12.6

dotjon0:
[...]

PR #1072, opened 2022-12-19
Update to analyzer 5+

nielsenko:
Fixes: #1070, fixes: #1073

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@
 ### Internal
 * Using Core 13.1.2.
 * No longer use vcpkg ([#1069](https://github.com/realm/realm-dart/pull/1069))
+* Upgraded analyzer dependency to ^5.0.0. ([#1072](https://github.com/realm/realm-dart/pull/1072))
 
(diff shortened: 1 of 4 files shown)


desistefanova:
In this case we will have to remove the changelog item :
"Bump minimum Dart SDK version to 2.17.5 due to an issue with the Dart virtual machine when implementing `Finalizable`. ([dart-lang/sdk#49075](https://github.com/dart-lang/sdk/issues/49075))"
And also I have to update the docs: https://github.com/mongodb/docs-realm/pull/2416

blagoev (changes requested):
Not sure we should do this yet. Lets discuss when is the appropriate time to do it.

nielsenko:
If we want to work with latest stable, ie. 3.3.10 we need to do this.

nielsenko:
> If we want to work with analyzer ^5.0.0 (which was released 3 month ago) we need latest stable, ie. 3.3.10, to do this. Community seems to want this.

My mistake, I'll drop this requirement

desistefanova (approved):

blagoev (approved):

PR #1078, opened 2022-12-21
Correctly propagate public name to the database

nirinchev:
Fixes https://github.com/realm/realm-dart/issues/697

### DIFF
--- a/lib/src/realm_property.dart
+++ b/lib/src/realm_property.dart
@@ -50,6 +50,9 @@ class SchemaProperty {
   final String? mapTo;
 
   /// @nodoc
+  String get internalName => mapTo ?? name;
+
+  /// @nodoc
   const SchemaProperty(
     this.name,
     this.propertyType, {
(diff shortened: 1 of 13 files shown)


desistefanova (approved):

lib/src/realm_property.dart
```diff
@@ -49,6 +49,9 @@ class SchemaProperty {
   /// Indicates that the property should be persisted under a different name
   final String? mapTo;
 
+  /// @nodoc
+  String get internalName => mapTo ?? name;
```
blagoev, line 53: That's introducing the problem which one should we use and when. Why not make `name` a getter which returns mapTo instead of name value?
```dart
String get name => mapTo ?? _name;
```

blagoev (changes requested):

lib/src/realm_property.dart
blagoev: or maybe we do need to differentiate in certain cases?

nielsenko, line 53:
```suggestion
  final String? _mapTo;

  /// @nodoc
  String get mapTo => _mapTo ?? name;
```
might also be a better option, but I'm okay with `internalName`

nielsenko (approved):

lib/src/realm_property.dart
nirinchev: We need the public name when setting up the Core schema.

nirinchev:
Those test failures seem unrelated to these changes.

blagoev (approved):

dotjon0:
Thanks so much @nielsenko

Issue #1087, opened 2023-01-03
[Bug]: non-zero custom status code considered fatal


Navil:
### What happened?

Our sentry reports a few fatal errors, when signing in:

> AppException: non-zero custom status code considered fatal, link to server logs: null

We do not get this on our devices, it is visible to us in sentry. We could reproduce it on a GT-I9505 with Android Version 5.0.1. So maybe it is related to old versions of Android?

The error message does not help much to find the issue.

### Repro steps

Sign in with Firebase (Phone Auth)
Sign in Realm (JWT)
Error

### Version

Channel stable, 3.3.10, on Microsoft Windows [Version 10.0.19045.2364], locale de-AT

### What Realm SDK flavor are you using?

MongoDB Atlas (i.e. Sync, auth, functions)

### What type of application is this?

Flutter Application

### Client OS and version

Android 5.0.1

### Code snippets

```
debugPrint('Verifying phone code'); 
try {
      // Sign the user in (or link) with the credential
      firebase.UserCredential firebaseCredential =
          await firebase.FirebaseAuth.instance.signInWithCredential(credential);
       debugPrint('Signed in with Firebase');

      String jwt = await firebaseCredential.user!.getIdToken();
      user = await app.logIn(Credentials.jwt(jwt));
      _verificationId = null;
      notifyListeners();
      debugPrint('Signed in with Realm');
    } catch (error, stackTrace) {
      Sentry.captureException(error, stackTrace: stackTrace);
    }
```

### Stacktrace of the exception/crash you're getting

```shell
AppException: AppException: non-zero custom status code considered fatal, link to server logs: null
```

### Relevant log output

```shell
Verifying phone code
Signed in with Firebase
AppException: AppException: non-zero custom status code considered fatal, link to server logs: null
```

nielsenko:
How does the jwt look?

Navil:
I do not have those tokens in my logs so I have no token from the crashing device. I could however try to reproduce it on the development environment or provide a working jwt if this helps.

nielsenko:
It might help. Remember to strip the signature part before posting the JWTs here.

I'm also interested in how you have setup Audience (aud claim) and Verification Method on the backend (https://www.mongodb.com/docs/atlas/app-services/authentication/custom-jwt/)

Navil:
So, I can reproduce the issue with an emulator running on 5.0.1.

Here is the content of the jwt:
```
{
  "alg": "RS256",
[...]

PR #1088, opened 2023-01-06
Add ISRG X1 Root certificate (needed for Android 7.0 and earlier)

nielsenko:
The certificates from https://realm.mongodb.com/ are signed by lets-encrypt, which is dependent on the root certificate ISRG Root X1.

The list of platforms that support the ISRG Root X1 root certificate can be seen here: https://letsencrypt.org/docs/certificate-compatibility/. In particular you need Android >= 7.1.1

There is a dirty work-around to make these certificates work with old Android apps (>= 2.3.6), that you can read about here: https://letsencrypt.org/docs/dst-root-ca-x3-expiration-september-2021/. Unfortunately I don't think this will work with BoringSSL that is used by Dart / Flutter.

Fixes: #1087

### DIFF
--- a/lib/src/app.dart
+++ b/lib/src/app.dart
@@ -27,8 +27,8 @@ import 'credentials.dart';
 import 'native/realm_core.dart';
 import 'user.dart';
 
-final _pinnedClient = () {
-  const isrgRootX1CertPEM = // The root certificate used by lets encrypt
+final _defaultClient = () {
+  const isrgRootX1CertPEM = // The root certificate used by lets encrypt and hence MongoDB
       '''
 subject=CN=ISRG Root X1,O=Internet Security Research Group,C=US
 issuer=CN=DST Root CA X3,O=Digital Signature Trust Co.
@@ -64,7 +64,7 @@ he8Y4IWS6wY7bCkjCWDcRQJMEhg76fsO3txE+FiYruq9RUWhiF1myv4Q6W+CyBFC
 Dfvp7OOGAN6dEOM4+qR9sdjoSYKEBpsr6GtPAQw4dy753ec5
 -----END CERTIFICATE-----''';
 
-  final context = SecurityContext();
+  final context = SecurityContext(withTrustedRoots: true);
   context.setTrustedCertificatesBytes(const AsciiEncoder().convert(isrgRootX1CertPEM));
   return HttpClient(context: context);
 }();
@@ -143,7 +143,7 @@ class AppConfiguration {
     HttpClient? httpClient,
   })  : baseUrl = baseUrl ?? Uri.parse('https://realm.mongodb.com'),
         baseFilePath = baseFilePath ?? Directory(_path.dirname(Configuration.defaultRealmPath)),
-        httpClient = httpClient ?? _pinnedClient;
+        httpClient = httpClient ?? _defaultClient;
 }
 
 /// An [App] is the main client-side entry point for interacting with an [Atlas App Services](https://www.mongodb.com/docs/atlas/app-services/) application.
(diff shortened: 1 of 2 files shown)


lib/src/app.dart
```diff
@@ -26,6 +27,48 @@ import 'credentials.dart';
 import 'native/realm_core.dart';
 import 'user.dart';
 
+final _pinnedClient = () {
```
nirinchev, line 30: This is not pinning the certificate, rather it's adding it to the set of trusted certificates. This doesn't prevent other certificates from being trusted by the app.

nirinchev (commented):
Looks good, assuming you verified it works. I have comments about the pinning wording, but those don't affect the actual implementation.

lib/src/app.dart
nielsenko: It pins the root certificate. No other root certificates are allowed

nielsenko: .. otherwise I should have done:
```dart
  final context = SecurityContext(withTrustedRoots: true);
```

nielsenko:
> Looks good, assuming you verified it works. I have comments about the pinning wording, but those don't affect the actual implementation.

It has been verified to work. Should we configure CI to run Android tests on an older image as well?

lib/src/app.dart
nielsenko: But only for the `HttpClient` we use ofcourse.

blagoev: `No other root certificates are allowed` this doesn't sound as the best approach imo.

nielsenko: https://security.stackexchange.com/questions/224246/certficate-pinning-should-i-pin-the-leaf-or-intermediate#:~:text=By%20pinning%20against%20the%20root,not%20to%20mis%2Dissue%20certificates.

.. but it is trivial to change

nielsenko:
> Should we check the android version and do this only on platforms that are problematic. Also isn't this a double edged sword `Conveniently it also ensure we are not dependent on the installed root certificates on the device.` If for some reasons server and client certificates are not working or updated. This means we will not work by default on newer Android versions. Then people will need to manually fix it for newer platforms by default. And this is just for supporting a low number of users using outdated Android versions.

Yes we will work everywhere, as long as realm.mongodb.com signs with a certificate that is in turn signed by ISRG X1 Root.

lib/src/app.dart
nielsenko: We restrict to only trust certificates signed (transitively) by ISRG X1 Root, which is less than what we trust today, but still a whole lot.. basically anything signed by lets-encrypt.

nielsenko: .. and different from what is trusted on Android prior to 7.1.1

nielsenko: After a long slack discussion I have added the root certificate instead of pinning it.

desistefanova (approved):
It is not a good practice to hardcode the certificate that will expire. But still it will work for some time. Probably we will have to update that at some point.

nielsenko:
> It is not a good practice to hardcode the certificate that will expire. But still it will work for some time. Probably we will have to update that at some point.

Root certificates are always "hardcoded". They are updated with system updates, if not added explicitly by the user.

Navil:
@nielsenko When adding
`withTrustedRoots: true`
it does not work anymore on my old testing device. Anything else to keep in mind when adding that flag?

nielsenko:
> @nielsenko When adding `withTrustedRoots: true` it does not work anymore on my old testing device. Anything else to keep in mind when adding that flag?

Interesting .. I will need to check up on why. You don't need that flag. I only added it in the PR, due to a discussion on slack.

lib/src/app.dart
nielsenko: This turns out to break it for @Navil. I was a bit quick to accept this change. Will revert to draft.

Issue #676, opened 2022-06-18
syncErrorHandler - provide better details in SyncError


jesusrp98:
I'm using the `flutter_flexible_sync` example as the basis of a POC I'm working on, and after adding sync capabilities with Atlas, I'm not able to make the app work in offline mode.

Am I doing something wrong?

This is the code I'm using to create the `Realm` object.
``` dart
Future<Realm> createRealm(String appId) async {
  final user = await App(AppConfiguration(appId)).logIn(
    Credentials.anonymous(),
  );

  final realm = Realm(
    Configuration.flexibleSync(
      user,
      [
        // ...
      ],
    ),
  );
  print("Created local realm db at: ${realm.config.path}");

  realm.subscriptions.update((mutableSubscriptions) {
    mutableSubscriptions.add(realm.all<Car>());
    mutableSubscriptions.add(realm.all<Person>());
  });
  await realm.subscriptions.waitForSynchronization();
  print("Syncronization completed for realm: ${realm.config.path}");

  return realm;
}
```

And this is the issue generated when trying to launch the app in offline mode:

```
flutter: [INFO] Realm: Connection[1]: Session[1]: client_reset_config = false, Realm exists = true, client reset = false
flutter: [ERROR] Realm: Failed to resolve 'ws.realm.mongodb.com:443': Host not found (authoritative)
```

desistefanova:
Hello @ jesusrp98,
Sorry we delayed the answer.
The flexible sync area can work offline. The error you see is just a background warning and it does not crash the application. 
It is the line 
     `await realm.subscriptions.waitForSynchronization()` 
that prevents your code from continuing to work with the offline realm.
Use this method `waitForSynchronization` only in case you insist to sync all data before to continue.
Hope this will help.

nirinchev:
Generally speaking, you don't need to recreate the subscriptions every time you run the app as those are persisted. You could do something like:

```dart
final cars = realm.all<Car>();
final people = realm.all<Person>();

if (realm.subscriptions.find(cars) == null || realm.subscriptions.find(people) == null)
{
  realm.subscriptions.update((mutableSubscriptions) {
    mutableSubscriptions.add(cars);
    mutableSubscriptions.add(people);
  });

  await realm.subscriptions.waitForSynchronization();
}
```

jesusrp98:
@nirinchev  @desistefanova 

So, in the end, the `realm.subscriptions.update` should only be make if realm cant find existing subscriptions, since they persist over sessions.

[...]

PR #1093, opened 2023-01-10
Update realm-core to v13.2.0

nielsenko:
- Update realm-core to v13.2.0
- Return bool from ClientResetError.resetRealm to indicate if reset was initiated or not
- Fill in new metrics required by core
- Update CHANGELOG

### DIFF
--- a/lib/src/configuration.dart
+++ b/lib/src/configuration.dart
@@ -612,9 +612,6 @@ class ClientResetError extends SyncError {
   /// Initiates the client reset process.
   ///
   /// Returns> `true` if actions were run successfully, `false` otherwise.
-  ///
-  /// All Realm instances for that path must be closed before this method is called or an
-  /// [RealmException] will be thrown.
   bool resetRealm() {
     if (_config is! FlexibleSyncConfiguration) {
       throw RealmException("The current configuration is not FlexibleSyncConfiguration.");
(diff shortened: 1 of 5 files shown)


nielsenko:
Note that we don't fill in the metrics correct yet, just the bare minimum to get it working with realm-core v13.2.0.

lib/src/native/realm_bindings.dart
```diff
@@ -11119,12 +11244,13 @@ abstract class realm_sync_error_category {
   static const int RLM_SYNC_ERROR_CATEGORY_CLIENT = 0;
   static const int RLM_SYNC_ERROR_CATEGORY_CONNECTION = 1;
   static const int RLM_SYNC_ERROR_CATEGORY_SESSION = 2;
+  static const int RLM_SYNC_ERROR_CATEGORY_RESOLVE = 3;
```
desistefanova, line 11247: @nielsenko  I will create a new branch for adding this category in relation to #676

lib/src/configuration.dart
```diff
@@ -611,14 +611,13 @@ class ClientResetError extends SyncError {
 
   /// Initiates the client reset process.
   ///
-  /// All Realm instances for that path must be closed before this method is called or an
-  /// [RealmException] will be thrown.
-  void resetRealm() {
+  /// Returns> `true` if actions were run successfully, `false` otherwise.
```
desistefanova, line 614:
```suggestion
  /// Returns `true` if actions were run successfully, `false` otherwise.
```

blagoev (approved):

PR #1102, opened 2023-01-15
Realm Set support

### LINKED ISSUE #779: Support Set data types.
realm_get_set,
realm_set_resolve_in,
realm_set_is_valid,
realm_set_size,
realm_set_get_property,
realm_set_find,
realm_set_get_property,
realm_set_insert,realm_set_erase,
realm_set_clear,
realm_set_remove_all,
realm_set_assign,
realm_set_add_notification_callback,
realm_set_from_thread_safe_reference,
+ generator

blagoev:
* Add support for Realm Set. 
* Updates generator with Realm Set support

```dart
@RealmModel()
class _TestRealmSets {
  @PrimaryKey()
  late int key;

  late Set<bool> boolSet;
  late Set<int> intSet;
  late Set<String> stringSet;
  late Set<double> doubleSet;
  late Set<DateTime> dateTimeSet;
  late Set<ObjectId> objectIdSet;
  late Set<Uuid> uuidSet;
  late Set<RealmValue> realmValueSet;
  late Set<_Car> objectsSet;

  late Set<bool?> nullableBoolSet;
  late Set<int?> nullableIntSet;
  late Set<String?> nullableStringSet;
  late Set<double?> nullableDoubleSet;
  late Set<DateTime?> nullableDateTimeSet;
  late Set<ObjectId?> nullableObjectIdSet;
  late Set<Uuid?> nullableUuidSet;
}
```

fixes #779 

This also fixes the generator tests to output the expected and the actual files. Also adds a launch config to debug the generator in VS Code.


nielsenko (approved):

lib/src/realm_object.dart
```diff
@@ -242,6 +248,28 @@ class RealmCoreAccessor implements RealmAccessor {
...
 
+      //TODO: set from ManagedRealmList is not supported yet
+      if (value is UnmanagedRealmSet) {
+        final handle = realmCore.getSetProperty(object, propertyMeta.key);
+        if (update) {
+          realmCore.realmSetClear(handle);
+        }
+
+        // TODO: use realmSetAssign when available in C-API
+        //realmCore.realmSetAssign(handle, value.toList());
```
desistefanova, line 259: It will be good to have links to issues in the TODO comments.

generator/lib/src/field_element_ex.dart
```diff
@@ -204,6 +212,34 @@ extension FieldElementEx on FieldElement {
...
+                  todo: 'Did you mean to use Set<RealmValue> instead.');
+            }
+
+            final initExpression = initializerExpression;
+            if (initExpression != null) {
+              throw RealmInvalidGenerationSourceError('Default values for set are not supported',
+                  primarySpan: initializerExpressionSpan(file, initExpression),
+                  primaryLabel: 'Remove the default value',
+                  element: this,
+                  todo: 'Remove the default value for field $displayName');
```
desistefanova, line 240: Not all the error messages have "." at the end.

lib/src/realm_class.dart
```diff
@@ -92,6 +93,7 @@ export "configuration.dart"
         SyncSessionError;
 export 'credentials.dart' show AuthProviderType, Credentials, EmailPasswordAuthProvider;
 export 'list.dart' show RealmList, RealmListOfObject, RealmListChanges, ListExtension;
+export 'set.dart' show RealmSet;
```
desistefanova, line 96: `RealmSetChanges` is not exposed.

desistefanova (changes requested):
It looks good. Here are some small comments.

generator/test/error_test_data/unsupported_realm_set_with_default_values.expected
```diff
@@ -0,0 +1,13 @@
+Default values for set are not supported
+
+in: asset:pkg/test/error_test_data/unsupported_realm_set_with_default_values.dart:10:27
+    ╷
+5   │ @RealmModel()
+6   │ class _Bad {
+    │       ━━━━ in realm model for 'Bad'
+... │
+10  │   late Set<bool> wrong1 = {true, false};
+    │                           ^^^^^^^^^^^^^ Remove the default value
```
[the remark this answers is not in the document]
blagoev: I had this working but decided to not allow it for now. We can add it later if we decide. There is nothing missing to just allow it except that it works for certain types and for others one can not use `const` at all. I will create an issue for later

generator/test/good_test_data/realm_set.dart
```diff
@@ -0,0 +1,39 @@
...
+  late Set<double?> nullabledoubleSet;
+
+  late Set<DateTime> dateTimeSet;
+  late Set<DateTime?> nullabledateTimeSet;
+
+  late Set<ObjectId> objectIdSet;
+  late Set<ObjectId?> nullableobjectIdSet;
+
+  late Set<Uuid> uuidSet;
+  late Set<Uuid?> nullableuuidSet;
```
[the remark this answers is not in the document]
blagoev: Thanks for the suggestion. I prefer to leave the grouping as it is

common/lib/src/realm_types.dart
```diff
@@ -69,6 +69,7 @@ enum RealmCollectionType {
   none,
   list,
   set,
+  _3, // ignore: unused_field, constant_identifier_names
```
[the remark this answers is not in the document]
blagoev: we can change but it won't improve much.

lib/src/realm_class.dart
blagoev: good catch

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
...
+  Sets(this.set, this.values, [this.getRealmObject]);
+}
+
+Future<void> main([List<String>? args]) async {
+  await setupTests(args);
+
+  for (var type in supportedTypes) {
+    test('RealmSet<$type> unmanged set add', () {
+      final testSet = TestRealmSets(1);
+      final set = testSet.setByType(type).set;
```
desistefanova, line 164: The values are initialized on this row, but they are not used. They are initialized on the next row again.

desistefanova (commented):
The helper methods and classes that are added for the purpose of the tests are a little bit confusing and the values of the sets are reinitialized too much. I prepared a suggestion how we can simplify the helper methods. I haven't touched the logic of the tests.
https://github.com/realm/realm-dart/blob/ds/realm_set_tests/test/realm_set_test.dart

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
...
+
+      expect(realm.find<TestRealmSets>(1), isNotNull);
+
+      testSet = realm.find<TestRealmSets>(1)!;
+      var set = testSet.setByType(type).set;
+
+      expect(set.length, equals(0));
+    });
+
+    test('RealmSet<$type> create from unmanaged', () {
```
desistefanova, line 218:
```suggestion
    test('RealmSet<$type> create set with values', () {
```

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
...
+
+      var testSet = TestRealmSets(1);
+      var set = testSet.setByType(type).set;
+      var values = testSet.values(type);
+
+      realm.write(() {
+        realm.add(testSet);
+      });
+
+      set = testSet.setByType(type).set;
```
desistefanova, line 256: Why is this repeated here?

desistefanova: It seems that `testSet.setByType(type).set` for unmanage `testSets` returns different instance than `testSet.setByType(type).set` when `testSets` is managed.

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
...
+
+      expect(set.contains(values.first), false);
+
+      realm.write(() {
+        set.add(values.first);
+      });
+
+      testSet = realm.find<TestRealmSets>(1)!;
+      set = testSet.setByType(type).set;
```
desistefanova, line 265: This shouldn't be needed. We already have the right `set` instance and it is refreshed.

desistefanova (commented):
I didn't checked all the tests, but getting `testSet.setByType(type).set;` all the time really confused me. Probably because I expect that once I have an unmanaged object with field empty `set` it will become managed after I add the main object to the realm. I might be wrong, but currently the new managed set is different instance than the unmanaged set.
For example, for the RealmList, if I create unmanaged team and players and then add the them to the realm, team.players collection is already managed.

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
+////////////////////////////////////////////////////////////////////////////////
+//
+// Copyright 2022 Realm Inc.
```
nirinchev, line 3: Lubo, stop living in the past 😜

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
...
+      set.add(values.first);
+      expect(set.length, equals(1));
+      expect(set.contains(values.first), true);
+
+      set.remove(values.first);
+      expect(set.length, equals(0));
+      expect(set.contains(values.first), false);
+    });
+
+    test('RealmSet<$type> unmanged set elementAt', () {
```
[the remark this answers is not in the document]
nirinchev: I guess it's confusing because `set` means both "a set of values" and "set a value". In this case I believe Lubo is using it as the "unmanaged set" noun. Perhaps it may be clearer if the test was renamed to something like `'RealmSet<$type>.elementAt (unmanaged)`.

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
...
+      final testSet = TestRealmSets(1);
+      final set = testSet.setByType(type).set;
+      final values = testSet.values(type);
+
+      set.add(values.first);
+      expect(set.length, equals(1));
+      expect(set.elementAt(0), values.first);
```
nirinchev, line 197: This is not testing error conditions, such as `elementAt(1)`, `elementAt(-1)`.

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
...
+      var values = testSet.values(type);
+
+      realm.write(() {
+        set.add(values.first);
+      });
+
+      expect(set.length, 1);
+
+      realm.write(() {
+        expect(set.remove(values.first), true);
```
nirinchev, line 310: This should probably also test the case where we remove a value that's missing from the set.

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
...
+      realm.write(() {
+        realm.add(testSet);
+        var set = testSet.setByType(type).set;
+        set.add(values.first);
+      });
+
+      var set = testSet.setByType(type).set;
+
+      expect(() => set.elementAt(-1), throws<RealmException>("Index out of range"));
+      expect(() => set.elementAt(800), throws<RealmException>());
```
nirinchev, line 361: Why don't we check the exception string here?

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
...
+    test('RealmSet<$type> unmanged set remove', () {
+      final testSet = TestRealmSets(1);
+      final set = testSet.setByType(type).set;
+      final values = testSet.values(type);
+
+      set.add(values.first);
+      expect(set.length, equals(1));
+      expect(set.contains(values.first), true);
+
+      set.remove(values.first);
```
nirinchev, line 185: Should this test the return value of `remove`? Also, should we have a test for removing a non-existent item?

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
...
+Future<void> main([List<String>? args]) async {
+  await setupTests(args);
+
+  for (var type in supportedTypes) {
+    test('RealmSet<$type> unmanged set add', () {
+      final testSet = TestRealmSets(1);
+      final set = testSet.setByType(type).set;
+      final values = testSet.values(type);
+
+      set.add(values.first);
```
nirinchev, line 167: Should we be testing the return value of `add`? Also, should we test adding a duplicate item?

lib/src/set.dart
```diff
@@ -0,0 +1,316 @@
...
+
+  @override
+  bool add(T value) {
+    //It is valid to call `add` with managed objects already in the set.
+    if (value is RealmObject) {
+      if (value.isManaged) {
+        _ensureManagedByThis(value, "add");
+      } else {
+        // might be updating an existing realm object
+        realm.add<RealmObject>(value, update: true);
```
nirinchev, line 143: Should we have `update: true` here? This may lead to unintended behavior.

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
...
+      case double:
+        return Sets(doubleSet as RealmSet<double>, [-1.1, 0.1, 1.1, 2.2, 3.3, 3.14]);
+      case DateTime:
+        return Sets(dateTimeSet as RealmSet<DateTime>, [DateTime(2023).toUtc(), DateTime(1981).toUtc()]);
+      case ObjectId:
+        return Sets(objectIdSet as RealmSet<ObjectId>, [ObjectId.fromTimestamp(DateTime(2023).toUtc()), ObjectId.fromTimestamp(DateTime(1981).toUtc())]);
+      case Uuid:
+        return Sets(uuidSet as RealmSet<Uuid>, [Uuid.fromString("12345678123456781234567812345678"), Uuid.fromString("82345678123456781234567812345678")]);
+      case RealmValue:
+        return Sets(mixedSet as RealmSet<RealmValue>, [RealmValue.nullValue(), RealmValue.bool(true), RealmValue.int(1), RealmValue.string("text")]);
```
nirinchev, line 111: It'll be a good idea to also test with a RealmValue containing a RealmObject.

nirinchev (commented):
I did a super cursory review of the tests and the public API of the collection

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
...
+    }
+
+    return set.values.map<Object?>((value) {
+      RealmObject? realmValue = set.getRealmObject!(realm, value as RealmObject);
+      return realmValue;
+    }).toList();
+  }
+}
+
+class Sets {
```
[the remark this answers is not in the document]
blagoev: Can you elaborate why is it adding more complexity?

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
...
+        return Sets(nullableObjectIdSet as RealmSet<ObjectId?>, [...setByType(ObjectId).values, null]);
+      case _NullableUuid:
+        return Sets(nullableUuidSet as RealmSet<Uuid?>, [...setByType(Uuid).values, null]);
+      default:
+        throw RealmError("Unsupported type $type");
+    }
+  }
+
+  List<Object?> values(Type type) {
+    return setByType(type).values;
```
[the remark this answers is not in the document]
blagoev: I believe we have discussed this, but again.  Do you see an issue with this? Is it a performance concern we should be addressing

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
...
+      final testSet = TestRealmSets(1);
+      final set = testSet.setByType(type).set;
+      final values = testSet.values(type);
+
+      set.add(values.first);
+      expect(set.length, equals(1));
+      expect(set.elementAt(0), values.first);
+    });
+
+    test('RealmSet<$type> create', () {
```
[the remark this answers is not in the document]
blagoev: will fix

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
...
+      default:
+        throw RealmError("Unsupported type $type");
+    }
+  }
+
+  List<Object?> values(Type type) {
+    return setByType(type).values;
+  }
+
+  List<Object?> getValuesOrManagedValues(Realm realm, Type type) {
```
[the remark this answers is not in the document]
blagoev: I don't know what you mean. As the name suggests this returns a list of values or managed values depending on the type requested

blagoev:
@nirinchev Thanks for the review. I will try addressing your suggestions above

test/realm_set_test.dart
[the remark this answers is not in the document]
desistefanova: It introduces one level more, which is not needed. I think it is simpler in the example that I sent. But it is just a suggestion.

desistefanova:
> I didn't checked all the tests, but getting `testSet.setByType(type).set;` all the time really confused me. Probably because I expect that once I have an unmanaged object with field empty `set` it will become managed after I add the main object to the realm. I might be wrong, but currently the new managed set is different instance than the unmanaged set. For example, for the RealmList, if I create unmanaged team and players and then add the them to the realm, team.players collection is already managed.

Please, ignore this comment: I checked it in details and it works as it is expected.

test/realm_set_test.dart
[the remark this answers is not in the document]
nirinchev: I also found it confusing when I read it - I think a more straightforward approach would be to create `setByType(type)` in your test and then call `.set` and `.values` on it.

test/realm_set_test.dart
```diff
@@ -0,0 +1,535 @@
...
+    test('RealmSet<$type> unmanged set add', () {
+      final testSet = TestRealmSets(1);
+      final set = testSet.setByType(type).set;
+      final values = testSet.values(type);
+
+      set.add(values.first);
+      expect(set.length, equals(1));
+      expect(set.contains(values.first), true);
+
+      set.remove(values.first);
```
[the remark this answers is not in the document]
blagoev: fixed

test/realm_set_test.dart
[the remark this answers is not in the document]
blagoev: fixed

test/realm_set_test.dart
blagoev: I will leave as it is. it is creating a managed RealmSet from unmanaged RealmSet

test/realm_set_test.dart
blagoev: I don't think we should be testing these. At least not here. Unmanged RealmSet is just an implementation of Dart Set. Did you mean to test these somewhere else?

test/realm_set_test.dart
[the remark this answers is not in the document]
blagoev: These tests are combing multiple concepts into one. `TestRealmSets` can be both unmanaged and managed and it contains RealmSet of different types so in order this to work in all cases one needs to go through the concrete instance of `TestRealmSets` every time to get the correct set and value instances.
so each test needs to call 
```
var set = testSet.setByType(type).set;
var values = testSet.values(type);
```
each time it needs to work with the correct data. 

Also `testSet.values` is just a convenience method for `testSet.setByType(type).values`.

I am not getting an local var of the `testSet.setByType(type)` result since then I will need three code lines each time to get the `set` and `values` from that local var. And I just don't care too much of the extra object created since this will get collected by the GC as soon as the method exits, since its a gen0 object. 

In general I agree these auto generated tests in all test files in the project that we are creating are hard to work with. Especially amplified by the lack of good tooling support for such tests. But then I don't think we have an option not to auto generate these. 

I will go with the tests that we have here and handle the feedback for the missing tests coverage.

test/realm_set_test.dart
blagoev: We already have tests for these in the managed case in the `elementAt` tests

test/realm_set_test.dart
blagoev: fixed

test/realm_set_test.dart
blagoev: fixed

test/realm_set_test.dart
blagoev: this is the unmanaged set implementation which is just inherited from a Dart set. I don't think we should be testing that here. 
I have changed the managed test to test for the `add` result

test/realm_set_test.dart
blagoev: see above for unmanaged `add`

lib/src/set.dart
blagoev: changed to false. good point

test/realm_set_test.dart
blagoev: good catch. This unraveled a lot of issues. Fixed in the latest commit

lib/src/set.dart
```diff
@@ -169,14 +177,12 @@ class ManagedRealmSet<T extends Object?> with RealmEntity, SetMixin<T> implement
...
-      if (!element.isManaged) {
-        return false;
-      }
-
-      _ensureManagedByThis(element, "contains");
+    if (!_isManagedRealmObject(element)) {
+      return false;
     }
 
+    _ensureManagedByThis(element, "contains");
```
desistefanova, line 184: "contains operation" will probably sounds better after the whole error message is built.

test/realm_set_test.dart
```diff
@@ -277,7 +286,10 @@ Future<void> main([List<String>? args]) async {
       realm.write(() {
         realm.add(testSet);
         var set = testSet.setByType(type).set;
-        set.add(values.first);
+        expect(set.add(values.first), true);
+
+        //adding an already existing value is a no operation
+        expect(set.add(values.first), false);
```
desistefanova, line 292: Looking at this I'm wondering if it would be valuable to test add with upsert option for the parent object. Imagine we want to create a set with managed objects that are already added in the realm and then we create the parent with the set.

desistefanova (commented):
Just a reminder to add `realm_set_test.dart` "flutter\realm_flutter\tests\test_driver\realm_test.dart".

blagoev:
@desistefanova 

> Just a reminder to add realm_set_test.dart "flutter\realm_flutter\tests\test_driver\realm_test.dart".

Good catch. Will add this

lib/src/set.dart
blagoev: I used what we use for lists as well. So lets keep it as is

test/realm_set_test.dart
blagoev: I will add a such test

lib/src/set.dart
```diff
@@ -220,18 +224,39 @@ class ManagedRealmSet<T extends Object?> with RealmEntity, SetMixin<T> implement
...
     }
 
-    if (element.realm != realm) {
-      if (element.isFrozen) {
+    RealmObject realmObject = value;
+
+    if (realmObject.realm != realm) {
+      if (realmObject.isFrozen) {
         throw RealmError('Cannot do $operation because the object is managed by a frozen Realm');
       }
```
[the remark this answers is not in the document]
blagoev: This will be handled by Core so we don't need to handle it in the SDK.

lib/src/set.dart
blagoev: fixed the error text

desistefanova (approved):
Looks good!

lib/src/set.dart
```diff
@@ -0,0 +1,341 @@
...
+  /// final characters = <String>{'A', 'B', 'C'};
+  /// final didRemoveB = characters.remove('B'); // true
+  /// final didRemoveD = characters.remove('D'); // false
+  /// print(characters); // {A, C}
+  /// ```
+  @override
+  bool remove(covariant T value); //Note: explicitly overriding remove() to change parameter type
+
+  /// Clear the set and if it is a set of Realm objects, delete all realm objects from the realm.
+  void deleteAll();
```
nirinchev, line 100: That's weird - does it really delete the objects from Realm? I don't believe such an API belongs on the set - we don't have it for list and I don't think we expose a similar API in any of the other SDKs.

nirinchev (commented):
Re: unmanaged set tests - I think we should treat the unmanaged set implementation as an implementation detail and not rely that it'll never change, so adding some tests for the behavior of add/remove/elementAt makes sense. We've had issues on the .NET SDK with inconsistent behavior between managed and unmanaged sets, especially with `RealmValue` containing different value types, but the same value (e.g. 0 integer, 0 double, and 0 decimal). I don't know if that'll be an issue for dart, but maybe worth adding a test.

lib/src/set.dart
blagoev: Well the test expectations show that objects are indeed delete from the realm and this is exposed in the C-API. 
I don't see why not adding it. Perhaps we can extends List to support it as well later. I think it's a good usability improvement. Don't know how much we should sync on this. Perhaps we could leave it and see if it is requested for the other SDKs.

blagoev:
We really redirect all UnmanagedSet operations to a base Set class which works with `equals` and `hashcode`. `RealmValue` overrides both so I don't think we will have issues there.

lib/src/set.dart
nirinchev: I don't think it's a good idea to have it - generally, we want to expose operations that mutate the Realm on the Realm itself. We already have `Realm.deleteMany` - if we want to optimize performance we could detect that the iterable is a RealmSet and call the C API to delete all elements. But having both `clear` and `deleteAll` on RealmSet is not good design.

PR #1103, opened 2023-01-15
Write copy

### LINKED ISSUE #796: realm.WriteCopyLocalToSync and SyncToLocal, LocalToLocal, SyncToSync
realm_convert_with_config, realm_convert_with_path 
https://www.mongodb.com/docs/realm-sdks/swift/latest/Structs/Realm.html#/s:10RealmSwift0A0V9writeCopy13configurationyAC13ConfigurationV_tKF

desistefanova:
Implement `writeCopy` that copies realm file:
Local to Local
Local to Local read-only
Local in-memory to Local
Local to Sync  - not supported
Sync to Sync
Sync to Local

Copy to file path is about to be deprecated.
Fixes #796


lib/src/native/realm_core.dart
```diff
@@ -2647,6 +2652,14 @@ Pointer<realm_value_t> _toRealmValue(Object? value, Allocator allocator) {
...
 }
 
+Pointer<realm_binary_t> _toRealmBinary(List<int>? value, Allocator allocator) {
+  final realm_binary = allocator<realm_binary_t>();
+  if (value != null) {
+    realm_binary.ref.data = value.toUint8Ptr(allocator);
+    realm_binary.ref.size = value.length;
+  }
+  return realm_binary;
+}
```
nielsenko, line 2662: Is this used anywhere?

test/realm_test.dart
```diff
@@ -1344,7 +1343,7 @@ Future<void> main([List<String>? args]) async {
     final app = App(appConfiguration);
     final queryDifferentiator = generateRandomString(10);
     const itemCount = 200;
```
nielsenko, line 1345:
```suggestion
    const itemsCount = 200;
```
for consistency

nielsenko (approved):

lib/src/native/realm_core.dart
desistefanova: It is not user anymore. I will delete this.

lib/src/realm_class.dart
```diff
@@ -530,6 +530,20 @@ class Realm implements Finalizable {
...
     }
   }
+
+  /// Writes a compacted copy of the `Realm` to the path in the specified config. If the configuration object has
+  /// non-null [Configuration.encryptionKey], the copy will be encrypted with that key.
+  ///
+  /// 1. The destination file cannot already exist.
+  /// 2. When using a local `Realm` and this is called from within a transaction it writes the current data,
+  ///    and not the data as it was when the last transaction was committed.
+  /// 3. When using Sync, it is required that all local changes are synchronized with the server before the copy can be written.
```
blagoev, line 540:
```suggestion
  /// 3. When using synced Realm, it is required that all local changes are synchronized with the server before the copy can be written.
```

lib/src/realm_class.dart
```diff
@@ -530,6 +530,20 @@ class Realm implements Finalizable {
...
+  ///    This is to be sure that the file can be used as a starting point for a newly installed application.
+  ///    The function will throw if there are pending uploads.
+  /// 4. Converting a Local `Realm` to Sync `Realm` is not supported.
```
blagoev, line 543:
```suggestion
  /// 4. copying a local `Realm` to a synced `Realm` is not supported.
```
Since we expose this as `copy` and not `convert` I think the better wording here is `copying a local Realm`

lib/src/native/realm_core.dart
```diff
@@ -2332,6 +2332,11 @@ class _RealmCore {
       return out_did_run.value;
     });
   }
+
+  void writeCopy(Realm realm, Configuration config) {
+    final configHandle = _createConfig(config);
+    _realmLib.invokeGetBool(() => _realmLib.realm_convert_with_config(realm.handle._pointer, configHandle._pointer, false));
```
blagoev, line 2338: Since we are using `convert_with_config` why don't we document this

> When merging, all classes must have a pk called '_id" otherwise an exception is thrown.

blagoev, line 2338: `realm_convert_with_config` supports merging with existing file. Should we be exposing this functionality as well?

test/realm_test.dart
```diff
@@ -1551,6 +1550,210 @@ Future<void> main([List<String>? args]) async {
     //test the realm can be opened.
     final realm = getRealm(Configuration.disconnectedSync([Product.schema], path: path, encryptionKey: key));
   });
+
+  test('Realm writeCopy Local to existing file', () {
```
blagoev, line 1554:
```suggestion
  test('Realm writeCopy local to existing file', () {
```
nit

test/realm_test.dart
```diff
@@ -1551,6 +1550,210 @@ Future<void> main([List<String>? args]) async {
...
+
+    final app = App(appConfiguration);
+    final credentials = Credentials.anonymous(reuseCredentials: false);
+    var user = await app.logIn(credentials);
+    final configCopy = Configuration.flexibleSync(user, [Product.schema]);
+    expect(() => originalRealm.writeCopy(configCopy),
+        throws<RealmException>("Realm cannot be converted to a flexible sync realm unless flexible sync is already enabled"));
+  });
+
+  test('Realm writeCopy Local->Local inside a write block', () {
```
blagoev, line 1579: we should disallow this

test/realm_test.dart
```diff
@@ -1551,6 +1550,210 @@ Future<void> main([List<String>? args]) async {
...
+    final pathCopy = originalConfig.path.replaceFirst(p.basenameWithoutExtension(originalConfig.path), generateRandomString(10));
+    final configCopy = Configuration.local([Car.schema], path: pathCopy, isReadOnly: true);
+
+    originalRealm.writeCopy(configCopy);
+    originalRealm.close();
+
+    expect(File(pathCopy).existsSync(), isTrue);
+    final copiedRealm = getRealm(configCopy);
+    expect(copiedRealm.isClosed, isFalse);
+    _expectAllWritesToThrow<RealmException>(copiedRealm, "Can't perform transactions on read-only Realms.");
```
blagoev, line 1612: This is testing readonly realms becuase we are just using a readonly config. These checks are not needed here. We should rather be checking if the data was copied correctly

test/realm_test.dart
```diff
@@ -1551,6 +1550,210 @@ Future<void> main([List<String>? args]) async {
...
+    final originalRealm = getRealm(originalConfig);
+    final pathCopy = originalConfig.path.replaceFirst(p.basenameWithoutExtension(originalConfig.path), generateRandomString(10));
+    final configCopy = Configuration.local([Car.schema], path: pathCopy);
+
+    originalRealm.writeCopy(configCopy);
+    originalRealm.close();
+
+    expect(File(pathCopy).existsSync(), isTrue);
+    final copiedRealm = getRealm(configCopy);
+    expect(copiedRealm.isClosed, isFalse);
```
blagoev, line 1629: why do we need this check exactly after we have opend a realm. That really not needed imo

test/realm_test.dart
```diff
@@ -1551,6 +1550,210 @@ Future<void> main([List<String>? args]) async {
...
+    final itemsCount = 100;
```
blagoev, line 1630: I would suggest reducing this to 2 items

test/realm_test.dart
```diff
@@ -1551,6 +1550,210 @@ Future<void> main([List<String>? args]) async {
...
+    copiedRealm.write(() {
+      for (var i = 0; i < itemsCount; i++) {
+        copiedRealm.add(Car("make_${i + 1}"));
```
blagoev, line 1633: I am not sure why we are adding new data to the copied realm

blagoev: shouldn't we be testing that the new realm has the copied data?

test/realm_test.dart
```diff
@@ -1551,6 +1550,210 @@ Future<void> main([List<String>? args]) async {
...
+        await _addSubscriptions(copiedRealm, productNamePrefix);
+        await copiedRealm.syncSession.waitForUpload();
+        await copiedRealm.syncSession.waitForDownload();
+        expect(copiedRealm.all<Product>().length, itemsCount);
+        expect(copiedRealm.isClosed, isFalse);
+        copiedRealm.close();
+
+        // Create another user's realm and download the data
+        var anotherUser = await app.logIn(Credentials.anonymous(reuseCredentials: false));
+        final anotherUserRealm = getRealm(Configuration.flexibleSync(anotherUser, [Product.schema]));
```
blagoev, line 1718: are we sure we are not opening the same realm here. isn't path required for what we are trying to do?

blagoev (changes requested):
We should disallow calling `copy` from within write transactions and migrations.

lib/src/realm_class.dart
```diff
@@ -530,6 +530,20 @@ class Realm implements Finalizable {
       realm.close();
     }
   }
+
+  /// Writes a compacted copy of the `Realm` to the path in the specified config. If the configuration object has
+  /// non-null [Configuration.encryptionKey], the copy will be encrypted with that key.
+  ///
+  /// 1. The destination file cannot already exist.
+  /// 2. When using a local `Realm` and this is called from within a transaction it writes the current data,
```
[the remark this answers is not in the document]
desistefanova: Ok, I will disable it.

lib/src/native/realm_core.dart
desistefanova: I don't understand this comment.

lib/src/realm_class.dart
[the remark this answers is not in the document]
desistefanova: It was aligned to .Net and Swift

test/realm_test.dart
desistefanova: This test is testing whether the new realm is really read-only.

test/realm_test.dart
desistefanova: I usually add this to all the tests, which have to prove that the realm can be opened.

test/realm_test.dart
desistefanova: This proves that the new realm is in memory. I add new data then close it and I expect the newly added data to disapper.

test/realm_test.dart
desistefanova: When using `reuseCredentials: false` a new user is created with a new id. The path to the opened flexible synced realm with this user will contain the id of the new user.

lib/src/realm_class.dart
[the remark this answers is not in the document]
desistefanova: I'm not sure how can I check if the copy is inside a write block. I will check that.

lib/src/native/realm_core.dart
desistefanova: It is not exposed by the other SDKs.

test/realm_test.dart
desistefanova: OK

test/realm_test.dart
desistefanova: Will remove that.

lib/src/realm_class.dart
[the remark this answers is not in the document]
desistefanova: Fixed. A RealmError is thrown if the copying is in a transaction or in a migration callback.

test/realm_test.dart
desistefanova: I added also a check if the data was copied correctly.

lib/src/realm_class.dart
```diff
@@ -530,6 +530,23 @@ class Realm implements Finalizable {
...
+  ///
+  /// 1. The destination file should not already exist.
+  /// 2. Copying realm is not allowed within a write transaction as well as during migration.
+  /// 3. When using synced Realm, it is required that all local changes are synchronized with the server before the copy can be written.
+  ///    This is to be sure that the file can be used as a starting point for a newly installed application.
+  ///    The function will throw if there are pending uploads.
+  /// 4. Copying a local `Realm` to a synced `Realm` is not supported.
+  void writeCopy(Configuration config) {
+    if (isInTransaction || _isInMigration) {
+      throw RealmError("Copying realm is not allowed within a write transaction as well as during migration.");
```
blagoev, line 545:
```suggestion
      throw RealmError("Copying a Realm is not allowed within a write transaction or during migration.");
```

test/realm_test.dart
blagoev: > This test is testing whether the new realm is really read-only.

There is no such thing as read-only realm. there are realms opened with read-only configuration. And as such this test is testing if readonly config is working. This is why I think these checks are not needed here at all

P.S. Please don't resolve comments that are not handled as the suggested.

test/realm_test.dart
blagoev: ok got it

blagoev (changes requested):

test/realm_test.dart
blagoev: > I usually add this to all the tests, which have to prove that the realm can be opened.
This should not be needed in any test other than the `isClosed` test that we have already

test/realm_test.dart
desistefanova: The comment was handled, since the check for copied data was added. But if we don't demonstrate that a copied realm open with read-only config is not writable, then what will be the difference with the copy local to local test. I wanted to give the users an idea that they are able to use this and to protect their data in some use cases. If you insist to remove the check, then it is better I to remove the whole test.

desistefanova: I removed the write checks. Let me know what you think about deleting the test.

blagoev: we have a test for readonly realms. I imagine a test where we open a readonly realm and copy it. Then we check only if the data was copied correctly and we don't need to do anything related to readonly on the copied realm.

blagoev (approved):

test/realm_test.dart
blagoev: the test is fine now

PR #1118, opened 2023-01-23
Update mapTo API doc

desistefanova:
Extends the description that `mapTo` can be user on class level.

### DIFF
--- a/common/lib/src/realm_common_base.dart
+++ b/common/lib/src/realm_common_base.dart
@@ -59,8 +59,7 @@ class RealmModel {
 
 /// MapTo annotation for class level and class member level.
 ///
-/// MapTo on class level indicates that a class should be persisted under a different collection name.
-/// MapTo on class member level indicates that a property should be persisted under a different name.
+/// Indicates that the class or the property should be persisted under a different name.
 /// This is useful when opening a Realm across different bindings where code style conventions might differ
 /// or when migrating models.
 ///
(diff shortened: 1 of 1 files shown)


common/lib/src/realm_common_base.dart
```diff
@@ -57,10 +57,12 @@ class RealmModel {
   const RealmModel([this.type = ObjectType.realmObject]);
 }
 
-/// MapTo annotation for class member level.
+/// MapTo annotation for class level and class member level.
 ///
-/// Indicates that a property should be persisted under a different name.
-/// This is useful when opening a Realm across different bindings where code style conventions might differ.
+/// MapTo on class level indicates that a class should be persisted under a different collection name.
```
blagoev, line 62:
```suggestion
/// MapTo on class level indicates that a class should be persisted under a different name.
```
The word `collection` here is not clear. I suggest removing it

common/lib/src/realm_common_base.dart
```diff
@@ -57,10 +57,12 @@ class RealmModel {
...
+/// MapTo on class member level indicates that a property should be persisted under a different name.
```
blagoev, line 63: The text now is too verbose imo. I would suggest keeping the previous text.
So 
```
/// Indicates that the class or the property should be persisted under a different name

```

blagoev (changes requested):
I have some suggestions to improve the proposed changes.

common/lib/src/realm_common_base.dart
desistefanova: Ok. Done.

common/lib/src/realm_common_base.dart
desistefanova: I applied your next suggestion.

blagoev (approved):

PR #1122, opened 2023-01-26
Flutter integration test

desistefanova:


desistefanova:
There is no benefit for our tests to be migrated to integration_test, since we don't have widgets.

PR #1123, opened 2023-01-27
add App.reconnect

nielsenko:
Expose the ability to reconnect all syncSessions in one go with `App.reconnect`.

Typical use would be to reconnect when a NIC change state, fx using the `connectivity_plus` package:
```dart
Connectivity().onConnectivityChanged.listen((event) {
  if (event != ConnectivityResult.none) {
    app.reconnect();
  }
});
```

Fixes: #383

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,14 @@
 **This project is in Release Candidate stage.**
 
 ### Enhancements
-* None
+* Add `App.reconnect` to reconnect ahead of schedule. Example of use in concert with `connectivity_plus` package:
+```dart
+Connectivity().onConnectivityChanged.listen((event) {
+  if (event != ConnectivityResult.none) {
+    app.reconnect();
+  }
+});
+```
 
 ### Fixed
 * None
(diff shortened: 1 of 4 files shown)


nielsenko:
Note that to have a good test we need a way to "disconnect" the network. We should revisit once we implement `GenericNetworkTransport` and all network traffic pass through the SDK code..

test/app_test.dart
```diff
@@ -292,6 +292,32 @@ Future<void> main([List<String>? args]) async {
...
+
+  baasTest('App.reconnect', (appConfiguration) async {
+    final app = App(appConfiguration);
+
+    final user = await app.logIn(Credentials.anonymous());
+    final configuration = Configuration.flexibleSync(user, [Task.schema]);
+    final realm = getRealm(configuration);
+    final session = realm.syncSession;
+
+    // TODO: We miss a way to force a disconnect. Once we implement GenericNetworkTransport
```
blagoev, line 304: can we use `HttpClient.close()`?

nielsenko: Problem is that core does not use the http transport abstraction during sync, only during authentication and such.. so we need a few more bits in place first

nielsenko: I believe we can (and should) customize this by creating our own network transport, but it is perhaps a bit beyond this particular task.

CHANGELOG.md
```diff
@@ -3,7 +3,14 @@
 **This project is in Release Candidate stage.**
 
 ### Enhancements
-* None
+* Add `App.reconnect()` providing a hint to Realm to reconnect all sync sessions.
+```dart
+Connectivity().onConnectivityChanged.listen((event) {
```
blagoev, line 8: I suggest we remove this sample here

blagoev (approved):
I suggest to remove the sample from the changelog.

CHANGELOG.md
nielsenko: Why? If I may ask?

desistefanova (approved):

PR #1128, opened 2023-01-30
CollectionChanges cleared notification

### LINKED ISSUE #795: CollectionChanges.isCleared
ChangeSet.IsCleared

[C-API] **CollectionChangeSet.collection_was_cleared** is not returned by **realm_collection_changes_get_num_changes** as an output parameter.

desistefanova:
Fixes #795

### DIFF
--- a/lib/src/collections.dart
+++ b/lib/src/collections.dart
@@ -45,7 +45,7 @@ class CollectionChanges {
   final List<Move> moves;
   final bool isCleared;
 
-  const CollectionChanges(this.deletions, this.insertions, this.modifications, this.modificationsAfter, this.moves, this.cleared);
+  const CollectionChanges(this.deletions, this.insertions, this.modifications, this.modificationsAfter, this.moves, this.isCleared);
 }
 
 /// Describes the changes in a Realm collection since the last time the notification callback was invoked.
@@ -74,7 +74,7 @@ class RealmCollectionChanges implements Finalizable {
   List<int> get newModified => _changes.modificationsAfter;
 
   /// `true` if the collection was cleared.
-  bool get isCeared => _changes.isCleared;
+  bool get isCleared => _changes.isCleared;
 }
 
 extension RealmCollectionChangesInternal on RealmCollectionChanges {
--- a/lib/src/collections.dart
+++ b/lib/src/collections.dart
@@ -43,8 +43,9 @@ class CollectionChanges {
   final List<int> modifications;
   final List<int> modificationsAfter;
   final List<Move> moves;
+  final bool cleared;
 
-  const CollectionChanges(this.deletions, this.insertions, this.modifications, this.modificationsAfter, this.moves);
+  const CollectionChanges(this.deletions, this.insertions, this.modifications, this.modificationsAfter, this.moves, this.cleared);
 }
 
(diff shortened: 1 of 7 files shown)


test/realm_set_test.dart
```diff
@@ -523,6 +523,45 @@ Future<void> main([List<String>? args]) async {
       await Future<void>.delayed(Duration(milliseconds: 20));
     });
 
+    test('RealmSet<$type>.cleared notifications', () async {
```
nielsenko, line 526: Why not structure this test like for lists and use `expectLater` + `emitsInOrder`? I think it reads better and avoids the brittle delays.

desistefanova: I was wondering the same. But decided to keep the tests in the files similar.
OK, I will change it.

nielsenko (approved):
LGTM

lib/src/collections.dart
```diff
@@ -43,8 +43,9 @@ class CollectionChanges {
   final List<int> modifications;
   final List<int> modificationsAfter;
   final List<Move> moves;
+  final bool cleared;
```
blagoev, line 46:
```suggestion
  final bool isCleared;
```

lib/src/collections.dart
```diff
@@ -71,6 +72,9 @@ class RealmCollectionChanges implements Finalizable {
   /// The indexes in the new version of the collection which were modified. Conceptually, it contains the same entries as [modified] but after the
   /// insertions and deletions have been accounted for.
   List<int> get newModified => _changes.modificationsAfter;
+
+  /// Gets a value indicating whether the collection has been cleared.
+  bool get cleared => _changes.cleared;
```
blagoev, line 77:
```suggestion
  bool get isCleared => _changes.cleared;
```

test/list_test.dart
```diff
@@ -1176,4 +1176,19 @@ Future<void> main([List<String>? args]) async {
     realm.write(() => team.players.move(2, 3));
     expect(team.players, [bob, alice, dan, carol]);
   });
+
+  test('RealmList.cleared notifications', () async {
```
blagoev, line 1180:
```suggestion
  test('RealmList.isCleared notifications', () async {
```

test/realm_set_test.dart
```diff
@@ -523,6 +523,28 @@ Future<void> main([List<String>? args]) async {
...
```
blagoev, line 526:
```suggestion
    test('RealmSet<$type>.isCleared notifications', () async {
```

blagoev (changes requested):

lib/src/collections.dart
nielsenko: Or should it be `wasCleared` then, as it has already happened. I'm okay with any of those, but probably prefer just `cleared`.

desistefanova: I will rename it to `isCleared`, to be known that it is a bool value. Otherwise, `cleared' could be thought that it would return the indexes of the cleared items.

nielsenko: Okay, also matches dotnet https://www.mongodb.com/docs/realm-sdks/dotnet/latest/reference/Realms.ChangeSet.html

blagoev (approved):

PR #1132, opened 2023-01-30
normalize sync connection params

blagoev:


lib/src/native/realm_core.dart
```diff
@@ -1413,36 +1429,37 @@ class _RealmCore {
...
-
       _realmLib.realm_app_config_set_base_url(handle._pointer, configuration.baseUrl.toString().toCharPtr(arena));
 
       _realmLib.realm_app_config_set_default_request_timeout(handle._pointer, configuration.defaultRequestTimeout.inMilliseconds);
 
       if (configuration.localAppName != null) {
         _realmLib.realm_app_config_set_local_app_name(handle._pointer, configuration.localAppName!.toCharPtr(arena));
       }
+      else {
+        _realmLib.realm_app_config_set_local_app_name(handle._pointer, ''.toCharPtr(arena));
```
nielsenko, line 1454: Should we use https://pub.dev/packages/package_info_plus to get this info?

lib/src/native/realm_core.dart
```diff
@@ -1413,36 +1429,37 @@ class _RealmCore {
...
+      }
 
       if (configuration.localAppVersion != null) {
         _realmLib.realm_app_config_set_local_app_version(handle._pointer, configuration.localAppVersion!.toCharPtr(arena));
+      } else {
+        _realmLib.realm_app_config_set_local_app_version(handle._pointer, ''.toCharPtr(arena));
```
nielsenko, line 1460: Should we use https://pub.dev/packages/package_info_plus

nielsenko (approved):
LGTM. We could consider using https://pub.dev/packages/package_info_plus to get the app name and version (see comments), but only for flutter.

lib/src/native/realm_core.dart
blagoev: Wouldn't bring another dependency for this. If needed we can add our own implementation.

blagoev: currently this is not needed

lib/src/native/realm_core.dart
```diff
@@ -581,8 +581,8 @@ class _RealmCore {
     });
   }
 
-  void realmSetAutoRefresh(Realm realm, bool enable) {
-    _realmLib.realm_set_auto_refresh(realm.handle._pointer, enable);
+  void realmDisableAutoRefreshForTesting(Realm realm) {
```
desistefanova, line 584: I changed this method name in `session` class and removed the parameter. I think that the parameter `enabled` should be available here in realm_core.dart. We just expose the C-API in this class, we don't have to support smart logic on this level in my oppinion.

lib/src/native/realm_core.dart
desistefanova: I wonder why to call the method `realm_app_config_set_local_app_name` if we don't need this.

lib/src/native/realm_core.dart
desistefanova: Same. Do we need to call this method for setting nothing?

test/realm_test.dart
```diff
@@ -28,8 +28,8 @@ import 'package:timezone/data/latest.dart' as tz;
 import 'package:path/path.dart' as p;
 import 'package:cancellation_token/cancellation_token.dart';
 import '../lib/realm.dart';
-import '../lib/src/realm_class.dart' as realmInternal;
 import 'test.dart';
+import '../lib/src/native/realm_core.dart';
```
desistefanova, line 32: I prefer `import '../lib/src/realm_class.dart' show RealmInternal;` instead of everything from realm_core.dart.

test/test.dart
```diff
@@ -340,7 +340,7 @@ const int jsMaxInt = 9007199254740991;
 const int jsMinInt = -9007199254740991;
 
 //Overrides test method so we can filter tests
-void test(String name, dynamic Function() testFunction, {dynamic skip}) {
+void test(String name, dynamic Function() testFunction, {dynamic skip, Map<String, dynamic>? onPlatform}) {
```
desistefanova, line 343: We don't use this parameter anywhere `onPlatform`. Why do we need it?

test/realm_test.dart
```diff
@@ -1843,6 +1843,18 @@ Future<void> main([List<String>? args]) async {
     expect(results.length, 1);
     receivePort.close();
   });
+
+  test('Device info', () {
```
desistefanova, line 1847: I think this test is better to be located at `app_test.dart`.  And probably `deviceName` and `deviceVersion` could be properties of AppConfigurationInternal.

lib/src/native/realm_core.dart
```diff
@@ -1413,36 +1429,37 @@ class _RealmCore {
...
+      //TODO: set cpu architecture when Core merges https://github.com/realm/realm-core/pull/6256#pullrequestreview-1275624284
       _realmLib.realm_app_config_set_cpu_arch(handle._pointer, ''.toCharPtr(arena));
 
-      _realmLib.realm_app_config_set_device_name(handle._pointer, ''.toCharPtr(arena));
-      _realmLib.realm_app_config_set_device_version(handle._pointer, ''.toCharPtr(arena));
+      final deviceName = getDeviceName();
+      _realmLib.realm_app_config_set_device_name(handle._pointer, deviceName.toCharPtr(arena));
+      
+      final deviceVersion = getDeviceVersion();
+      _realmLib.realm_app_config_set_device_version(handle._pointer, deviceVersion.toCharPtr(arena));
```
desistefanova, line 1441: According to me deviceName and deviceVerison could be members of AppConfigurationInternal.

test/realm_test.dart
blagoev: We are not doing that anywhere. The fact that we need to import realm_class explicitly shows we should just import realm_core. that is the reason I have changed this and removed the internal implementation

blagoev: An internal implementation is needed only if called from multiple places from within the SDK libraries.

test/test.dart
blagoev: that is just making the interface compatible with the testing framework API. In-case we need it.

test/realm_test.dart
blagoev: This is not related to app at all. The native implementation is exposed as realm_dart_xxx as well.

lib/src/native/realm_core.dart
blagoev: Internal members are needed only if accessed on multiple places within the sdk libraries ie (configuration.dart, set.dart, realm_object.dart etc)

lib/src/native/realm_core.dart
blagoev: please, see the next comment

lib/src/native/realm_core.dart
blagoev: as above

test/realm_test.dart
blagoev: To clarify in other places in the tests we are importing realm_core.dart when we want to work with core internals for some tests. Hence we do it here as well

PR #1140, opened 2023-02-05
Upgrade to core13.4

desistefanova:


CHANGELOG.md
```diff
@@ -3,16 +3,17 @@
 **This project is in Release Candidate stage.**
 
 ### Enhancements
-* None
+* Improve performance of interprocess mutexes on iOS which don’t need to support reader-writer locking. The primary beneficiary of this is beginning and ending read transactions, which is now almost as fast as pre-v13.0.0 (Core upgrade).
```
nielsenko, line 6: From previous experience I believe this should go under fixed. We are not enhancing the dart sdk. See discussion for upgrade to v13.3.0

desistefanova: Comparing to the [core changelog](https://github.com/realm/realm-core/blob/master/CHANGELOG.md) it is enhancement. Are you sure we have to move it under Fixes?

blagoev: The difference is very subtle but having this in Fixes is a tid better imo. Previously we decided to move it to fixes

CHANGELOG.md
```diff
@@ -3,16 +3,17 @@
...
 
 ### Fixed
 * None
 
 ### Compatibility
 * Realm Studio: 13.0.0 or later.
+* Fileformat: Generates files with format v23. Reads and automatically upgrade from fileformat v5 (Core upgrade).
```
blagoev, line 13: I think we should remove this line? We mentioned it on the Core 13 upgrade changelog,  but now we are just bumping a minor version of Core.

blagoev (approved):
+1 on @nielsenko's comment

nielsenko (approved):

Issue #1017, opened 2022-11-07
[Bug]: [LINUX] Failed to load dynamic library 'librealm_dart.so': librealm_dart.so: cannot open shared object file: No such file or directory


c0c4i:
### What happened?

Hello, I'm using realm 0.7.0+rc on linux arm64 for Flutter.
In this initial code
```
var config = Configuration.local(
      [
        RealmBoard.schema,
        // other schema
      ],
      schemaVersion: 12,
    );
    realm = Realm(config);
```
I have this exception
```
"Failed to load dynamic library 'librealm_dart.so': librealm_dart.so: cannot open shared object file: No such file or directory"
```
I don't know what to do.
Thank you for help.

### Repro steps

1. Run the app
2. Crash on line `realm = Realm(config);`

### Version

Flutter 3.5.0-12.0.pre.145 • channel master • https://github.com/flutter/flutter.git Framework • revision a1289a4135 (5 hours ago) • 2022-11-07 06:46:25 -0500 Engine • revision 891d4a3577 Tools • Dart 2.19.0 (build 2.19.0-374.0.dev) • DevTools 2.19.0

### What Realm SDK flavor are you using?

Local Database only

### What type of application is this?

Flutter Application

### Client OS and version

Linux Ubuntu 20.04 LTS

### Code snippets

```
var config = Configuration.local(
      [
        RealmBoard.schema,
        // other schema
      ],
      schemaVersion: 12,
    );
    realm = Realm(config);
```

### Stacktrace of the exception/crash you're getting

```shell
#0      _open (dart:ffi-patch/ffi_dynamic_library_patch.dart:11:43)
#1      new DynamicLibrary.open (dart:ffi-patch/ffi_dynamic_library_patch.dart:22:12)
#2      _open (package:realm/src/init.dart:84:52)
#3      _openRealmLib.<anonymous closure> (package:realm/src/init.dart:70:11)
#4      _openRealmLib (package:realm/src/init.dart:76:18)
#5      initRealm (package:realm/src/init.dart:105:24)
#6      new _RealmCore._ (package:realm/src/native/realm_core.dart:72:17)
#7      new _RealmCore (package:realm/src/native/realm_core.dart:81:37)
#8      realmCore (package:realm/src/native/realm_core.dart:51:30)
#9      realmCore (package:realm/src/native/realm_core.dart)
#10     Realm._openRealm (package:realm/src/realm_class.dart:180:12)
[...]

PR #1144, opened 2023-02-06
Reproduce: loading library failure

desistefanova:
This PR reproduces the issue: #1141, #1017


desistefanova:
This PR will not be merged. It was created to prove that loading library fails for ubuntu-20 if it is build on 22.

PR #1147, opened 2023-02-07
fix test expectation

blagoev:


test/subscription_test.dart
```diff
@@ -519,47 +519,48 @@ Future<void> main([List<String>? args]) async {
...
-    } catch (e) {
-      expect(e is RealmException, isTrue);
-      expect((e as RealmException).message, expectedErrorMessage);
-      expect(realm.subscriptions.state, SubscriptionSetState.error);
-      expect(realm.subscriptions.error, isNotNull);
-      expect(realm.subscriptions.error is RealmException, isTrue);
-      expect((realm.subscriptions.error as RealmException).message, expectedErrorMessage);
-    }
-  });
+  //TODO: remove after App Services support for all queryable fields has landed
```
nielsenko, line 522: I think I would prefer skip over uncommented, but just a nit.

nielsenko (approved):

test/subscription_test.dart
blagoev: me too, but `skip` was harder to add :)

blagoev: actually, I will just remove the whole thing

desistefanova (approved):

PR #1167, opened 2023-02-17
Use system temporary directory during test

### LINKED ISSUE #1166: Named pipes are breaking flutter in MacOS
### What happened?

After installing realm and running some unit tests, `flutter run` and `flutter build` stopped working.

This is caused by the named pipes that are created by the test implementation of realm. Reported in this flutter issue: https://github.com/flutter/flutter/issues/108805

This could be fixed by changing the default directory for files created by realm to `/tmp`.

### Repro steps

1. Create a project with realm used in a widget test
2. Run the widget tests
3. Observe that some named pipes are left behind by realm
    * default.realm.note
    * default.realm.management/access_control.new_commit.cv
    * default.realm.management/access_control.pick_writer.cv
4. Run `flutter run --verbose`
5. Observe that it is stuck at `xattr -r -d com.apple.FinderInfo /path/to/project`

### Version

Flutter 3.7.1 • channel stable • https://github.com/flutter/flutter.git Framework • revision 7048ed95a5 (2 weeks ago) • 2023-02-01 09:07:31 -0800 Engine • revision 800594f1f4 Tools • Dart 2.19.1 • DevTools 2.20.1

Local Database only

Flutter Application

### Client OS and version

MacOS 13.1 (22C65)

### Code snippets

our realm configuration:

```
    final config =
        Configuration.local([DatabaseSystem.schema, DatabaseDevice.schema]);
    _realmDatabase = Realm(config);
```

_No response_

### Relevant log output

_No response_

maBarabas:
Fixes #1166

### DIFF
--- a/lib/src/native/realm_core.dart
+++ b/lib/src/native/realm_core.dart
@@ -2219,7 +2219,7 @@ class _RealmCore {
   String getAppDirectory() {
     try {
       if (!isFlutterPlatform || Platform.environment.containsKey('FLUTTER_TEST')) {
-        return Directory.current.absolute.path; // dart or flutter test
+        return Directory.systemTemp; // dart or flutter test
       }
 
       // Flutter from here on..


maBarabas:
@cla-bot check

lib/src/native/realm_core.dart
```diff
@@ -2219,7 +2219,7 @@ class _RealmCore {
   String getAppDirectory() {
     try {
       if (!isFlutterPlatform || Platform.environment.containsKey('FLUTTER_TEST')) {
-        return Directory.current.absolute.path; // dart or flutter test
+        return Directory.systemTemp; // dart or flutter test
```
nirinchev, line 2222: This is a breaking change since it moves database files in dart apps (e.g. console apps) to systemTemp, which is not what we want.

nirinchev (changes requested):

lib/src/native/realm_core.dart
maBarabas: From what I can see, `getAppDirectory` is a private API (not documented in dartdoc) and it's only caller is `Configuration.defaultSotragePath` when `isFlutterPlatform` is true. Maybe this could be refactored to be more obvious.

Since this path is not documented anywhere, I think it should be OK to change it for Dart unit tests and Flutter tests?

desistefanova:
We don't need to merge this PR. A sample is given in https://github.com/realm/realm-dart/issues/1166

PR #1168, opened 2023-02-17
Upgrade core to 13.4.2

desistefanova:


nielsenko (approved):

blagoev (changes requested):
Suggest we list this fix as well

> Sync client may have failed an assertion during shutdown if all sessions hadn't been ready to finalize by the time the Client destructor ran. (Core upgrade)

blagoev (approved):

PR #1171, opened 2023-02-20
Bump github-actions commits sha

desistefanova:
Related to #971
To avoid CI warnings we are update actions to use Node 16 and action-core@1.10.0
Only `coverallsapp` action is not updated and causes two warnings.


nielsenko (approved):
LGTM

blagoev (approved):

PR #1174, opened 2023-02-21
Strip android symbols

### LINKED ISSUE #1172: Why is the realm lib in the release build so big?
### Problem

Why is the realm lib in the release build so big?
realm

My project is hitting 123mb after adding the realm
build

### Solution

Decrease lib size in realm build

### Alternatives

_No response_

Would be a major improvement

Local Database only

blagoev:
This greatly reduces the Android binaries sizes in `Release` builds. 
We are now stripping binaries on CI as we should. The issue was compounded by the upgrade to OpenSSL 3.

fixes #1172

### DIFF
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -33,14 +33,14 @@ jobs:
 
       - name: Check cache
         id: check-cache
-        if: ${{ !contains(github.head_ref, 'release/') }}
+        # if: ${{ !contains(github.head_ref, 'release/') }}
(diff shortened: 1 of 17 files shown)


nirinchev (commented):
Can you run this build with caching disabled (e.g. change the cache key to something like `binaries1-...` to make sure the artifacts size is reduced?

desistefanova (approved):

blagoev:
@nirinchev Already did that

PR #1186, opened 2023-02-28
Test unicode symbols

### LINKED ISSUE #1191: Test unicode symbols
Add test for opening a realm with unicode symbols in the path.

Add test for queries and subscriptions with unicode parameters.

PR - [https://github.com/realm/realm-dart/pull/1186](https://github.com/realm/realm-dart/pull/1186)

desistefanova:
Add new tests with unicode symbols.
Generated new strings for test emails using the allowed symbols for email.
Fixes #1191
Depends on core release:
https://github.com/realm/realm-core/issues/6350
https://github.com/realm/realm-core/issues/6336

### DIFF
--- a/test/realm_test.dart
+++ b/test/realm_test.dart
@@ -308,7 +308,7 @@ Future<void> main([List<String>? args]) async {
     realm.write(() => realm
       ..add(Car("Audi"))
       ..add(Car("Tesla")));
-    final cars = realm.query<Car>(r'make == $0', ['Tesla']);
+    final cars = realm.query<Car>(r'make == "$0"', ['Tesla']);
     expect(cars.length, 1);
     expect(cars[0].make, "Tesla");
   });
@@ -331,7 +331,7 @@ Future<void> main([List<String>? args]) async {
     expect(t1.players, [p1]);
     expect(t2.players, [p2]);
     expect(t3.players, [p1, p2]);
-    final filteredTeams = realm.query<Team>(r'$0 IN players AND name BEGINSWITH $1', [p1, 'A']);
+    final filteredTeams = realm.query<Team>(r'$0 IN players AND name BEGINSWITH "$1"', [p1, 'A']);
     expect(filteredTeams.length, 1);
     expect(filteredTeams[0].name, "A1");
   });
@@ -1348,7 +1348,7 @@ Future<void> main([List<String>? args]) async {
     final config = await _subscribeForAtlasAddedData(app, queryDifferentiator: queryDifferentiator, itemsCount: itemsCount);
     var syncedRealm = await getRealmAsync(config);
     expect(syncedRealm.isClosed, false);
-    final data = syncedRealm.query<Product>(r'name BEGINSWITH $0', [queryDifferentiator]);
+    final data = syncedRealm.query<Product>(r'name BEGINSWITH "$0"', [queryDifferentiator]);
     expect(data.length, itemsCount);
   });
 
@@ -1828,7 +1828,7 @@ Future<void> main([List<String>? args]) async {
     final realm = getRealm(Configuration.local([Person.schema]));
     String personName = generateRandomString(5);
     final path = realm.config.path;
-    final results = realm.query<Person>(r"name == $0", [personName]);
+    final results = realm.query<Person>(r"name == '$0'", [personName]);
 
     expect(realm.refresh(), false);
     realmCore.realmDisableAutoRefreshForTesting(realm);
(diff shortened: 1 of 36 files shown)


test/test.dart
```diff
@@ -399,9 +399,18 @@ String generateRandomRealmPath() {
 }
 
 final random = Random();
+String _getRandom(int len, String chars) => List.generate(len, (index) => chars[random.nextInt(chars.length)]).join();
+
 String generateRandomString(int len) {
-  const chars = 'abcdefghjklmnopqrstuvwxuz';
-  return List.generate(len, (index) => chars[random.nextInt(chars.length)]).join();
+  return _getRandom(len, 'abcdefghjklmnopqrstuvwxuz');
```
blagoev, line 405: Not sure I understand the reasoning behind this change.

test/test.dart
```diff
@@ -399,9 +399,18 @@ String generateRandomRealmPath() {
 }
 
 final random = Random();
+String _getRandom(int len, String chars) => List.generate(len, (index) => chars[random.nextInt(chars.length)]).join();
```
blagoev, line 402: _getRandom as a method name that returns a `String` is not great imo. I suggest we revert this change

test/test.dart
```diff
@@ -399,9 +399,18 @@ String generateRandomRealmPath() {
...
-  return List.generate(len, (index) => chars[random.nextInt(chars.length)]).join();
+  return _getRandom(len, 'abcdefghjklmnopqrstuvwxuz');
+}
+
+String generateRandomUnicodeString(int len) {
+  return _getRandom(len, 'uvwxuzфоо-барΛορεμლორემ植物החללجمعتsøren');
+}
+
+String generateRandomEmail(int len) {
+  return _getRandom(len, r"abcdefghjklmnopqrstuvwxuz!#$%&*+-'/=?^_`{|}~0123456789");
```
blagoev, line 413: This does not seem to generate an random email. It just generates a random String from what it seems allowed characters presumably for valid email address. 
I suggest we refactor this code to really return a random email address in full including the domain name and make this one invoke the generateRandomString

test/realm_test.dart
```diff
@@ -1859,6 +1859,46 @@ Future<void> main([List<String>? args]) async {
     expect(realmCore.getDeviceName(), matcher);
     expect(realmCore.getDeviceVersion(), matcher);
   });
+
+  test('Realm path with unicode symbols', () {
+    var config = Configuration.local([Car.schema], path: "${generateRandomUnicodeString(10)}.realm");
```
blagoev, line 1864: This should just call generateRandomEmail passing an optional argument for including unicode symbols or not

blagoev (commented):
the tests seem fine but the `generate` method changes needs refactoring

test/test.dart
blagoev: also I don't think we need all the allowed symbols in an email. We just need to make sure we are using a different email in our tests.

desistefanova: I refactored the methods as you suggested. I also added an options for including the allowed characters for email, because I think that it is good to have them in case some email validation issue appear in future.

test/test.dart
desistefanova: I removed this method.

test/test.dart
desistefanova: I needed to generate strings containing unicode and not containing unicode symbols. That's why this method was refactored.

desistefanova:
> One thing. There is not guarantee that the `generateRandomString` function actually return a string with non-ascii characters, when the `unicode` parameter is set to `true`. It is likely, but not guaranteed. Also, ascii characters are part of unicode, so perhaps the flag could have a better name.

Ok, I changed the name to `allowUnicode`, which means not only ASCI characters could be expected.

test/test.dart
```diff
@@ -399,9 +399,14 @@ String generateRandomRealmPath() {
 }
 
 final random = Random();
-String generateRandomString(int len) {
-  const chars = 'abcdefghjklmnopqrstuvwxuz';
-  return List.generate(len, (index) => chars[random.nextInt(chars.length)]).join();
+String generateRandomString(int len, {bool allowUnicode = false, String includingSymbols = ""}) {
```
blagoev, line 402: This will work but I find the API little weird. 
I would prefer if the symbols passed are replacing the symbols used in the method. So the caller is either using the default implementation of `generateRandomString` or passing a character set specific at the call site. 
Otherwise one needs to really go and check which symbols are the default ones and also most importantly one can not exclude symbols from the default ones. Say for example I want to generate a random string which does not include `abc` in it.

Lets rename `includingSymbols` to characterSet and make it optional and null. If the caller passes it then we use only this characterSet to generate the random string otherwise we use the default one

blagoev (changes requested):

test/test.dart
blagoev: or better yet we initialize the `characterSet` parameter with the default character set right in the method definition

desistefanova: If we replace the whole charset with `characterSet `parameter's value then the parameter `allowUnicode` is useless. I will have to add again the method `generateRandomUnicodeString`. I refactored the methods according to your suggestions.

blagoev (approved):

PR #1192, opened 2023-03-02
Add Decimal128

nielsenko:
Add support for `Decimal128`. This behaves as you would expect from IEEE754, including behaviour of NaN comparison, ie. `Nan == NaN` is `false`. Note that you can still query for a `NaN` value in the DB as realm-core treats `NaN` as any other (single) value, to have a total order.

Be aware that the Dart `compareTo` method is implemented to honor the total ordering of IEEE754-2008 paragraf 5.10, which mimic the behaviour of Dart `double`. Note that this means that `compareTo` will order any values, including `+/- NaN`, `+/- Infinity` as well as `+/- 0`. This implies the `compareTo` operation is different from the partial ordering given by `operator==`, `operator<`, and `operator>`. For example, IEEE doubles impose that `Decimal128.zero == -Decimal128.zero` and all comparison operations on NaN return false.

This function imposes a complete ordering for doubles. When using `compareTo`, the following properties hold:

- All NaN values are considered equal, and greater than any numeric value.
- `-Decimal128.zero` is less than `Decimal128.zero` (and the integer 0), but greater than any non-zero negative value.
- Negative infinity is less than all other values and positive infinity is greater than all non-NaN values.
- All other values are compared using their numeric value.


dotjon0:
so pleased this is in motion @nielsenko , thank you soooo much!

nielsenko:
@dotjon0 I will need to postpone this work about a week, but then I'll return to it. I hope it is not inconvenient.

Note that it is a fairly rudimentary implementation supporting just basic the arithmetic operations (`+`,`-`,`*`,`/`, and `abs()`)and the ability to convert to from `int`, `double`, and `String`.

dotjon0:
> @dotjon0 I will need to postpone this work about a week, but then I'll return to it. I hope it is not inconvenient.
> 
> Note that it is a fairly rudimentary implementation supporting just basic the arithmetic operations (`+`,`-`,`*`,`/`, and `abs()`)and the ability to convert to from `int`, `double`, and `String`. 
> 
> 

@nielsenko no need to apologise, we are just very grateful for this feature going in! Functionality sounds great. Thank you so much

desistefanova (commented):
I think it will be good to have tests about performing math operation between different types. For example:
decimal128 * int, decimal128 + double, decimal128 * long ...

lib/src/native/realm_core.dart
```diff
@@ -47,9 +47,26 @@ import '../user.dart';
 import '../set.dart';
 import 'realm_bindings.dart';
 
-late RealmLibrary _realmLib;
+const bugInTheSdkMessage = "This is likely a bug in the Realm SDK - please file an issue at https://github.com/realm/realm-dart/issues";
 
-final _RealmCore realmCore = _RealmCore();
+final _realmLib = () {
```
desistefanova, line 52: Is this change related to decimal128?

lib/src/native/realm_core.dart
```diff
@@ -47,9 +47,26 @@ import '../user.dart';
...
+    final additionMessage =
+        isFlutterPlatform ? bugInTheSdkMessage : "Did you forget to run `dart run realm_dart install` after upgrading the realm_dart package?";
+    throw RealmException('Realm SDK package version does not match the native library version ($libraryVersion != $nativeLibraryVersion). $additionMessage');
+  }
+  return result;
+}();
+
+// stamped into the library by the build system (see prepare-release.yml)
+const libraryVersion = '1.0.3';
+late String nativeLibraryVersion;
```
desistefanova, line 65: Where the `nativeLibraryVersion` is set?

nielsenko: line 54

lib/src/native/realm_core.dart
nielsenko: Yes. Allows `lib` to be used before `_RealmCore` is instantiated.

lib/src/decimal128.dart
```diff
@@ -0,0 +1,149 @@
...
+import 'dart:convert';
+import 'dart:ffi';
+
+import 'package:ffi/ffi.dart';
+
+import 'native/realm_bindings.dart';
+import 'native/realm_core.dart';
+
+/// A 128-bit decimal floating point number.
+class Decimal128 extends Comparable<Decimal128> {
```
desistefanova, line 28: Should be exported in realm_class.dart. It is not visible in the API documentation.

lib/src/decimal128.dart
```diff
@@ -0,0 +1,149 @@
...
+  /// The value 0.
+  static final zero = Decimal128.fromInt(0);
+
+  /// The value 1.
+  static final one = Decimal128.fromInt(1);
+
+  /// The value 10.
+  static final ten = Decimal128.fromInt(10);
```
desistefanova, line 36: Wondering, why 1 and 10 are special and have them as properties.

lib/src/decimal128.dart
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  /// Converts a `int` into a [Decimal128].
+  factory Decimal128.fromInt(int value) {
+    return Decimal128._(lib.realm_dart_decimal128_from_int64(value));
+  }
+
+  /// Converts a `double` into a [Decimal128].
+  factory Decimal128.fromDouble(double value) {
+    return Decimal128.parse(value.toString()); // TODO(kn): Find a way to optimize this
```
desistefanova, line 74: One TODO comment has left here.

test/decimal128_test.dart
```diff
@@ -0,0 +1,305 @@
...
+
+    expect(Decimal128.one.compareTo(Decimal128.nan), -1);
+    expect(Decimal128.nan.compareTo(Decimal128.nan), 0);
+    expect(Decimal128.nan.compareTo(Decimal128.one), 1);
+  });
+
+  repeatTest('Decimal128.compareTo + <, <=, ==, !=, >=, >', (x, xInt, y, yInt) {
+    expect(x.compareTo(y), -(y.compareTo(x)));
+    expect(x.compareTo(y), xInt.compareTo(yInt));
+    expect(x == y, y == x);
```
desistefanova, line 243: expect(x != y, xInt != yInt);

test/decimal128_test.dart
```diff
@@ -0,0 +1,305 @@
...
+    expect(x < y, xInt < yInt);
+    expect(x <= y, y >= x);
+    expect(x <= y, xInt <= yInt);
+    expect(x > y, y < x);
+    expect(x > y, xInt > yInt);
+    expect(x >= y, y <= x);
+    expect(x >= y, xInt >= yInt);
+
+    expect(x.compareTo(x), 0);
+    expect(x == x, isTrue);
```
desistefanova, line 255:
```suggestion
     expect(x != x, isFalse);
     expect(x == x, isTrue);
```

test/decimal128_test.dart
desistefanova, line 243:
```suggestion
    expect(x != y, xInt != yInt);
    expect(x == y, y == x);
```

test/decimal128_test.dart
```diff
@@ -0,0 +1,305 @@
...
+
+        body(x, xInt, y, yInt);
+      },
+      times,
+    );
+  });
+}
+
+Future<void> main([List<String>? args]) async {
+  test('Decimal128.nan', () {
```
desistefanova, line 53: What about adding some test for query realm by Decimal128?

lib/src/decimal128.dart
nielsenko: I expect 0,1 and 10 to be the most used constants, so I const construct them. Obviously the user is free not to use them, or define const constructed values themself. Problem is the Decimal.fromInt cannot be a const constructor, so it is more memory efficient to use `Decimal128.one` vs `Decimal.fromInt(1)` repeatedly.. and also ever so slightly faster

lib/src/decimal128.dart
nielsenko: Yeah.. it is a bit of a bummer.. Let me look it over again

lib/src/native/realm_core.dart
blagoev: I would suggest we move decimal128.dart into the native directory and keeping the initialization of realmCore as is. 
_realmLib is intentionally hidden from the rest of the SDK. We could declare deciam128 as a partial file to realm_core.dart if we want to keep it separate from the rest of the native invocations. Which I agree is better.

lib/src/native/realm_core.dart
```diff
@@ -47,9 +47,26 @@ import '../user.dart';
...
+  if (libraryVersion != nativeLibraryVersion) {
+    final additionMessage =
+        isFlutterPlatform ? bugInTheSdkMessage : "Did you forget to run `dart run realm_dart install` after upgrading the realm_dart package?";
+    throw RealmException('Realm SDK package version does not match the native library version ($libraryVersion != $nativeLibraryVersion). $additionMessage');
+  }
+  return result;
+}();
+
+// stamped into the library by the build system (see prepare-release.yml)
+const libraryVersion = '1.0.3';
```
blagoev, line 64: This won't work with our prepare-release.yaml. We are looking for a specific string to replace there.

blagoev (changes requested):
I don't see the support for `Decimal128` in Realm models. I think we should expand this PR and add it here.

lib/src/decimal128.dart
blagoev: Why do we expect these to be frequently used?

lib/src/native/realm_core.dart
blagoev, line 52: I would suggest we move decimal128.dart into the native directory and keeping the initialization of realmCore as is.
_realmLib is intentionally hidden from the rest of the SDK. We could declare deciam128 as a partial file to realm_core.dart if we want to keep it separate from the rest of the native invocations. Which I agree is better.

nielsenko:
> I don't see the support for `Decimal128` in Realm models. I think we should expand this PR and add it here.

I have extended the generator to allow `Decimal128` props, and added a simple test

lib/src/native/realm_core.dart
nielsenko: Hmm.. I thought I had updated `prepare-release.yml` let me check..

test/dynamic_realm_test.dart
```diff
@@ -87,16 +87,17 @@ Future<void> main([List<String>? args]) async {
...
       nullableDateProp: date,
       nullableDoubleProp: -123.456,
       nullableObjectIdProp: objectId,
       nullableUuidProp: uuid,
-      nullableIntProp: 123);
+      nullableIntProp: 123,
+      nullableDecimalProp: Decimal128.fromDouble(4242));
 
-  AllTypes _getEmptyAllTypes() => AllTypes('', false, DateTime(0).toUtc(), 0, objectId, uuid, 0);
+  AllTypes _getEmptyAllTypes() => AllTypes('', false, DateTime(0).toUtc(), 0, objectId, uuid, 0, Decimal128.zero);
```
desistefanova, line 100: Decimal128 has to be added also in `_validateDynamic` and `_validateDynamicLists` below.

desistefanova (commented):
Will be good to add Decimal128 also to `realm_set_test.dart` and `realm_value_test.dart` tests.

lib/src/decimal128.dart
nielsenko: Being the additive identity, multiplicative identity, and the base (0, 1, and 10). It also matches https://pub.dev/documentation/decimal/latest/decimal/Decimal-class.html

nielsenko:
> I don't see the support for `Decimal128` in Realm models. I think we should expand this PR and add it here.

I have included that in this PR as well.

nielsenko:
> Will be good to add Decimal128 also to `realm_set_test.dart` and `realm_value_test.dart` tests.

Done. This actually exposed the fact that I forgot to enable `Decimal128` in `RealmValue`

nielsenko:
> I don't see the support for `Decimal128` in Realm models. I think we should expand this PR and add it here.

We now have full support, including usage in mixed.

lib/src/native/realm_core.dart
nielsenko: I tried doing this, but it gets ugly quickly. Fx. we then need to `show` a class from `native/realm_core.dart`. I'm not sure we want to go that way. What do you think?

nielsenko: You can see the changes in the [Move decimal128.dart to native as parts file of realm_core.dart](https://github.com/realm/realm-dart/pull/1192/commits/c9433333669cbf74c8298df66b136c2c9be2e2c0) commit

lib/src/native/realm_core.dart
nielsenko: Fixed

lib/src/decimal128.dart
nielsenko: We got to live with this for now :-/

lib/src/native/realm_core.dart
blagoev: I think its better. I like that we are not importing ffi and bindings stuff into non native part of the SDK.

The only "downside" is we need to do 
> export 'native/realm_core.dart' show Decimal128;

but that's not an issue since we are in control there.

test/decimal128_test.dart
```diff
@@ -0,0 +1,305 @@
...
+        final y = Decimal128.fromInt(yInt);
+
+        body(x, xInt, y, yInt);
+      },
+      times,
+    );
+  });
+}
+
+Future<void> main([List<String>? args]) async {
```
blagoev, line 52: we should import test.dart and `await setupTests` in all tests files. Thus at least keep consistency on init between the test files

test/decimal128_test.dart
```diff
@@ -0,0 +1,305 @@
...
+const int defaultTimes = 100000;
+
+void repeat(dynamic Function() body, [int times = defaultTimes]) {
+  for (var i = 0; i < times; ++i) {
+    body();
+  }
+}
+
+@isTest
+void repeatTest(String description, dynamic Function(Decimal128 x, int xInt, Decimal128 y, int yInt) body, [int times = defaultTimes]) {
```
blagoev, line 34: I can't invoke single test from command line or vs code. Could we make this work

lib/src/native/realm_core.dart
```diff
@@ -47,9 +48,26 @@ import '../user.dart';
...
+late String nativeLibraryVersion;
```
blagoev, line 68: It seems we don't need this field. So 
>nativeLibraryVersion = result.realm_dart_library_version().cast<Utf8>().toDartString();
could be a local variable.

blagoev (changes requested):
All in all its good. I have some suggestions for improvements.

test/decimal128_test.dart
nielsenko: We don't use realms at all in `decimal128_test.dart` so I really have no reason for that. The integration of `Decimal128` with realm is tested elsewhere, like in `realm_object_test.dart`, `realm_value_test.dart`, etc.
But sure..

test/decimal128_test.dart
blagoev: I mean I tried that in VS code. Is it working for you?

test/decimal128_test.dart
blagoev: I see your point, but I am afraid we might miss some initialization we rely on if we diverge here. For example the individual test timeout gets set up,  `test` method gets replace etc. For example I was surprised with my tests timing out while debugging these tests.

test/decimal128_test.dart
blagoev: I just tried and I get the same error 
> No tests match regular expression "^Decimal128 divide( \(variant: .*\))?$".
and I think I have the latest versions of the VSCode extensions

nielsenko: Yes. How do you invoke it?

blagoev: Also if I do 
dart test -n "device" -r expanded in the repo the `decimal` tests are not executed

lib/src/native/realm_core.dart
blagoev: Also If I do `dart test -n "devide" -r expanded` in the repo the `decimal` tests are not included in the run at all.

test/decimal128_test.dart
```diff
@@ -0,0 +1,305 @@
...
+  final r = Random(42); // use a fixed seed to make tests deterministic
```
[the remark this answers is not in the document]
nielsenko: Fixed, but for the record...

`r` is declared and used twice over just 7 lines.. this is **not** a readability issue. I will fix this just because, but please don't vilify all short variable names. They put emphasis on the _operations_ as opposed to the _variables_.

Or in the words of Bob Nyström (one of the Dart language designers): https://journal.stuffwithstuff.com/2016/06/16/long-names-are-long/

test/decimal128_test.dart
nielsenko: Fixed

test/decimal128_test.dart
```diff
@@ -0,0 +1,305 @@
...
+
+    expect(x.compareTo(x), 0);
+    expect(x == x, isTrue);
+    expect(x < x, isFalse);
+    expect(x <= x, isTrue);
+    expect(x > x, isFalse);
+    expect(x >= x, isTrue);
+  });
+
+  repeatTest('Decimal128.toInt/fromInt roundtrip', (x, xInt, y, yInt) {
```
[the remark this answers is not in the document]
nielsenko: Okay. This the reg exp that VSCode builds to select the test when you press the little `run` button:
image
I have fixed the test naming

src/realm_dart.cpp
```diff
@@ -53,9 +55,12 @@ RLM_API void realm_dart_initializeDartApiDL(void* data) {
     Dart_InitializeApiDL(data);
 }
 
+using namespace realm;
```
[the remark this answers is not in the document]
nielsenko: Done

lib/src/native/realm_core.dart
nielsenko: Try `dart test -n 'divide' -r expanded` instead.

src/realm_dart.cpp
```diff
@@ -53,9 +53,12 @@ RLM_API void realm_dart_initializeDartApiDL(void* data) {
...
+using namespace realm::c_api;
```
blagoev, line 57: Do we need these statements now in this fail?

blagoev (approved):

src/realm_dart.cpp
nielsenko: No, will remove

test/decimal128_test.dart
nielsenko: Done

desistefanova (approved):

test/decimal128_test.dart
nielsenko: It is tested in line 261

PR #1193, opened 2023-03-02
Cancel in-flight workflows when a commit is pushed

nirinchev:
This makes it so existing workflows runs are canceled when a new workflow starts for the same branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,8 +12,11 @@ env:
   BAAS_PROJECT_ID: ${{ secrets.ATLAS_QA_PROJECT_ID}}
   REALM_CI: true
 
-jobs:
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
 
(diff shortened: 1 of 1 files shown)


nielsenko (approved):
👍

.github/workflows/ci.yml
```diff
@@ -12,8 +12,11 @@ env:
   BAAS_PROJECT_ID: ${{ secrets.ATLAS_QA_PROJECT_ID}}
   REALM_CI: true
 
-jobs:
+concurrency:
+  group: ${{ github.head_ref || github.run_id }}
```
desistefanova, line 16: Is this mean, that if we push to the `main` branch the old workflows on this branch will not be canceled, because the group will be by `run_id`? If yes, that's great.

nirinchev: Yes - `head_ref` on `main` will be empty, so it will fall back to the `run_id`.

desistefanova (approved):

PR #1194, opened 2023-03-03
Backward compatibility with Xcode 13 for IOS

### LINKED ISSUE #1197: Allow user to use Xcode 13
As Realm JS does we should handle this gracefully. See https://github.com/realm/realm-js/pull/5531

desistefanova:
In relation to https://developer.apple.com/forums/thread/716965
Added flag `-fno-objc-msgsend-selector-stubs` for IOS
Fixes #1197


src/CMakeLists.txt
```diff
@@ -37,16 +38,19 @@ else()
...
 elseif(CMAKE_SYSTEM_NAME STREQUAL "Linux")
     string(APPEND OUTPUT_DIR "/linux")
 elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
     string(APPEND OUTPUT_DIR "/macos")
+    add_cxx_flag_if_supported(-fno-objc-msgsend-selector-stubs)
 elseif(CMAKE_SYSTEM_NAME STREQUAL "Android")
     string(APPEND OUTPUT_DIR "/android/${CMAKE_ANDROID_ARCH_ABI}")
 elseif(CMAKE_SYSTEM_NAME STREQUAL "iOS")
     string(APPEND OUTPUT_DIR "/ios")
+    add_cxx_flag_if_supported(-fno-objc-msgsend-selector-stubs)
```
blagoev, line 53: I would add these later in the 
`elseif(CMAKE_SYSTEM_NAME STREQUAL iOS)`
and would add another branch for macOS as 
`elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")`

blagoev (changes requested):

src/CMakeLists.txt
desistefanova: Ok, I will return it backwards.

blagoev (approved):
looks ok. 
lets do the testing before merging

desistefanova:
We don't need this change, since the framework is dynamic, not static.

desistefanova:
We don't need this change, since the framework is dynamic, not static.

PR #1219, opened 2023-03-17
Update ci-actions to 043cc5a8446748b823957c6ea4d08e6e495ea996

nielsenko:
Bump to latest green build of realm/ci-actions

### DIFF
--- a/.github/workflows/Issue-Needs-Attention.yml
+++ b/.github/workflows/Issue-Needs-Attention.yml
@@ -9,4 +9,4 @@ on:
 
 jobs:
   applyNeedsAttentionLabel:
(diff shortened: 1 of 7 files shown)


blagoev (approved):
looks ok, I guess we are updating to a newer version all these actions. Not sure what's in the new version. Could we add some info in the PR description just so we remember why we did this.

nielsenko:
> looks ok, I guess we are updating to a newer version all these actions. Not sure what's in the new version. Could we add some info in the PR description just so we remember why we did this.

It is just the latest green build of `ci-actions`. Wanted to bump so we don't fall to much behind, but there i nothing important for us as I can tell.

PR #1226, opened 2023-03-24
Core upgrade 13.8.0 - Core Logging

### LINKED ISSUE #859: Support Core Logging

### LINKED ISSUE #1234: Log messages with timestamps by default
Currently we are not logging with timestamps and it is useful to include that information.

desistefanova:
Support Core Logging
Applied changes from https://github.com/realm/realm-core/pull/6339
The default logger is wrapped into `_RealmLogger`. To manage the log level at runtime use `Realm.logger.level`.
The default log level in 'Info'. Default print is available only in the first isolate working with realm.
Setting an instance of `Logger` is still supported, but it will be wrapped into an instance of `_RealmLogger` and the default printing is no more available.
The default printing could be stopped also with `Realm.logger.clearListeners()`.

Fixes #859
Fixes #1234


lib/src/native/realm_core.dart
```diff
@@ -80,7 +80,12 @@ class _RealmCore {
   }
 
   factory _RealmCore() {
-    return _instance ??= _RealmCore._();
+    _instance ??= _RealmCore._();
+    final logCallback = Pointer.fromFunction<Void Function(Handle, Int32, Pointer<Int8>)>(_logCallback);
+    _realmLib.realm_set_log_callback(
+        logCallback.cast(), Realm.logger.level.toInt(), noopUserdata.toWeakHandle(), _realmLib.addresses.realm_dart_delete_weak_handle);
```
desistefanova, line 86: We set the logger here in advance in order to have it before the scheduler is initialized.
`realm_dart_delete_weak_handle` is a dummy function, because the userdata_free function is required.

blagoev:
We should handle #1234 as part of the core logging changes

lib/src/realm_class.dart
```diff
@@ -486,10 +487,20 @@ class Realm implements Finalizable {
     return realmCore.realmEquals(this, other);
   }
 
-  /// The logger to use for logging
-  static Logger logger = Logger.detached('Realm')
-    ..level = RealmLogLevel.info
-    ..onRecord.listen((event) => print(event));
+  /// The logger to use for logging.
+  /// The default logger is [RealmLogger] and the default log level is [RealmLogLevel.info].
```
nielsenko, line 491: I don't think we should mention `RealmLogger`.

lib/src/realm_class.dart
```diff
@@ -935,3 +892,154 @@ class MigrationRealm extends DynamicRealm {
...
+}
+
+/// Represents a logger that manages the realm log level at runtime.
+///
+/// {@category Realm}
+class RealmLogger implements Logger {
+  final Logger _logger;
+
+  /// Creates an instance of [RealmLogger].
+  RealmLogger({Level level = RealmLogLevel.info, void Function(LogRecord)? onRecord})
```
nielsenko, line 958: I think we should get rid of this ctor, and set the default logger on the static `Realm.logger` property instead, and have it wrapped by a `_RealmLogger` just as if the user had set it themself.

lib/src/realm_class.dart
```diff
@@ -935,3 +892,154 @@ class MigrationRealm extends DynamicRealm {
...
+  @override
+  String get fullName => _logger.fullName;
+
+  @override
+  void info(Object? message, [Object? error, StackTrace? stackTrace]) {
+    _logger.info(message, error, stackTrace);
+  }
+
+  @override
+  bool isLoggable(Level value) {
```
nielsenko, line 1013: Consider using `=>`

test/realm_test.dart
```diff
@@ -1898,6 +1899,29 @@ Future<void> main([List<String>? args]) async {
     expect(query.length, 1);
     expect(query[0].name, productName);
   });
+
+  test('Realm logger change level to error', () {
+    final oldLogger = Realm.logger;
+    try {
+      int count = 0;
```
nielsenko, line 1906: Do you want to actually test something regarding count at the end? As is it could be the callback was never invoked.

test/app_test.dart
```diff
@@ -217,6 +217,35 @@ Future<void> main([List<String>? args]) async {
     }
   });
 
+  baasTest('Change Realm.logger level at runtime', (configuration) async {
```
nielsenko, line 220: It seems the two added tests basically test the same..

nielsenko (changes requested):

lib/src/realm_class.dart
desistefanova: @nielsenko thank you about your comments. All your suggestions are handled in https://github.com/realm/realm-dart/pull/1226/commits/5da5b375f813574d63eaaae13709cae28a847e7b commit.

test/test.dart
```diff
@@ -743,3 +743,10 @@ extension StreamEx<T> on Stream<Stream<T>> {
     await inner?.cancel();
   }
 }
+
+Future<void> waitFutureWithTimeout(Future<void> future, {String? timeoutError, int seconds = 300}) async {
```
nielsenko, line 747: There exists a `.timeout` on `Future` already?

lib/src/scheduler.dart
```diff
@@ -22,21 +22,27 @@ import 'dart:isolate';
 import 'native/realm_core.dart';
 
 final _receivePortFinalizer = Finalizer<RawReceivePort>((p) => p.close());
-final Scheduler scheduler = Scheduler._();
+late final Scheduler scheduler;
```
blagoev, line 25: I think we should not do changes in this file except the `logMessage` change

lib/src/scheduler.dart
```diff
@@ -49,3 +55,10 @@ class Scheduler {
     handle.release();
   }
 }
+
+///@nodoc
+abstract class RealmCoreScheduler {
```
blagoev, line 60: This is not needed

src/realm_dart_logger.cpp
```diff
@@ -0,0 +1,113 @@
...
+// limitations under the License.
+//
+////////////////////////////////////////////////////////////////////////////////
+
+#include <sstream>
+#include <set>
+#include <mutex>
+#include <thread>
+#include <map>
+#include <chrono>
```
blagoev, line 24:
```suggestion
#include <chrono>
#include <algorithm>
```

src/realm_dart_logger.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+bool send_message_to_scheduler(Dart_Port port, realm_log_level_e level, const char* message)
+{
+    Dart_CObject c_level;
+    c_level.type = Dart_CObject_kInt32;
+    c_level.value.as_int32 = level;
+
+    Dart_CObject c_message;
+    c_message.type = Dart_CObject_kString;
+    c_message.value.as_string = (char*)message;
```
blagoev, line 72: We should use `const_cast` here

blagoev: Also this seems very unsafe are we sure we are copying the memory from this pointer?

src/realm_dart_logger.cpp
```diff
@@ -0,0 +1,113 @@
...
+    }
+    realm_set_log_callback(realm_dart_logger_callback, default_log_level, nullptr, nullptr);
+    is_core_logger_callback_set = true;
+    return true;
+}
+
+RLM_API void realm_dart_set_log_level(realm_log_level_e level, Dart_Port port)
+{
+    std::lock_guard lock(dart_logger_mutex);
+    if (dart_send_ports[port] != level)
```
blagoev, line 106: If the port does not exists?

test/app_test.dart
```diff
@@ -185,35 +186,60 @@ Future<void> main([List<String>? args]) async {
   });
 
   baasTest('Realm.logger', (configuration) async {
-    final oldLogger = Realm.logger;
+    Realm.logger = Logger.detached(generateRandomString(10))..level = RealmLogLevel.all;
```
blagoev, line 189: Why not move these tests in `realm_logger_test.dart`

test/realm_test.dart
```diff
@@ -22,6 +22,7 @@ import 'dart:async';
 import 'dart:convert';
 import 'dart:io';
 import 'dart:isolate';
+import 'package:logging/logging.dart';
```
blagoev, line 25: why this change?

test/test.dart
```diff
@@ -375,22 +375,26 @@ void xtest(String? name, dynamic Function() testFunction, {dynamic skip, Map<Str
...
   setUpAll(() async => await (baasSetup ??= setupBaas()));
 
   setUp(() {
-    Realm.logger = Logger.detached('test run')
-      ..level = Level.ALL
-      ..onRecord.listen((record) {
-        testing.printOnFailure('${record.time} ${record.level.name}: ${record.message}');
-      });
-
+    final defaultLogger = Realm.logger;
```
blagoev, line 384: Could we isolate these changes only for `realm_logger_test.dart`. it seems to excessive we need to do these globally

test/realm_logger_test.dart
```diff
@@ -0,0 +1,142 @@
...
+import 'dart:async';
+import 'dart:isolate';
+import 'package:logging/logging.dart';
+import 'package:test/test.dart' hide test, throws;
+import '../lib/realm.dart';
+import 'test.dart';
+
+Future<void> main([List<String>? args]) async {
+  await setupTests(args, loggerTest: true);
+  baasTest('Realm loggers log messages in all the isolates', (appConfiguration) async {
```
blagoev, line 30: Can we do a test for checking an expected message from Core is really logged by the logger. Also for different expected levels

blagoev (changes requested):

CHANGELOG.md
```diff
@@ -1,16 +1,30 @@
...
 
 ### Fixed
-* None
+* Installed one default logger to be used all over Core (Core upgrade).
+* Fixed a bug that may have resulted in arrays being in different orders on different devices (Core upgrade).
+* Fixed a crash when querying a mixed property with a string operator (contains/like/beginswith/endswith) or with case insensitivity (Core upgrade).
+* Querying for equality of a string on an indexed mixed property was returning case insensitive matches. For example querying for `myIndexedMixed == "Foo"` would incorrectly match on values of "foo" or "FOO" etc (Core upgrade).
+* Adding an index to a Mixed property on a non-empty table would crash with an assertion (Core upgrade).
+* `SyncSession.pause()` could hold a reference to the database open after shutting down the sync session, preventing users from being able to delete the realm (Core upgrade).
+* `Realm.logger` did not perform any locking, resulting in data races if it was called while the default logger was being read on another isolate (Core upgrade).
```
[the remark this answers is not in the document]
desistefanova: Ok

CHANGELOG.md
```diff
@@ -1,16 +1,30 @@
...
+* Performance improvement for the following queries (Core upgrade):
+    * Significant (~75%) improvement when counting (query count) the number of exact matches (with no other query conditions) on a String/int/Uuid/ObjectId property that has an index. This improvement will be especially noticiable if there are a large number of results returned (duplicate values).
+    * Significant (~99%) improvement when querying for an exact match on a Timestamp property that has an index.
+    * Significant (~99%) improvement when querying for a case insensitive match on a Mixed property that has an index.
+    * Moderate (~25%) improvement when querying for an exact match on a Boolean property that has an index.
+    * Small (~5%) improvement when querying for a case insensitive match on a Mixed property that does not have an index.
 
 ### Fixed
-* None
+* Installed one default logger to be used all over Core (Core upgrade).
```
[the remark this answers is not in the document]
desistefanova: Ok

CHANGELOG.md
```diff
@@ -1,16 +1,30 @@
 ## vNext (TBD)
 
 ### Enhancements
-* None
+* `Realm.logger.level` allows changing the log level at runtime per isolate. To stop the logging use level `RealmLogLevel.off`([#1226](https://github.com/realm/realm-dart/pull/1226)).
+* Added a default print of the logs only for the first isolate working with the realm. Calling `Realm.logger.clearListeners()` will stop the printing by default ([#1226](https://github.com/realm/realm-dart/pull/1226)).
```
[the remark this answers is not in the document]
desistefanova: Ok

lib/src/native/realm_core.dart
```diff
@@ -67,15 +67,36 @@ final _realmLib = () {
...
   static const int RLM_INVALID_CLASS_KEY = 0x7FFFFFFF;
   // ignore: unused_field
   static const int RLM_INVALID_PROPERTY_KEY = -1;
   // ignore: unused_field
   static const int RLM_INVALID_OBJECT_KEY = -1;
 
-  final encryptionKeySize = 64;
+  final int encryptionKeySize = 64;
+
+  static Object noopUserdata = Object();
```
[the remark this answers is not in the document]
desistefanova: It is updated, just this line was left by mistake.

lib/src/native/realm_core.dart
```diff
@@ -80,7 +80,10 @@ class _RealmCore {
   }
 
   factory _RealmCore() {
-    return _instance ??= _RealmCore._();
+    _instance ??= _RealmCore._();
+    scheduler = Scheduler.init(_instance!);
```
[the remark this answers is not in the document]
desistefanova: They both depend on each other the `Scheduler` and the `_RealmCore`.

lib/src/scheduler.dart
desistefanova: Let's discuss this.

src/realm_dart_logger.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include <map>
+#include <chrono>
+
+#include "realm_dart_logger.h"
+#include <realm/object-store/c_api/types.hpp>
+
+auto& dart_logger_mutex = *new std::mutex;
+bool is_core_logger_callback_set = false;
+std::map<Dart_Port, realm_log_level_e> dart_send_ports;
+realm_log_level_e default_log_level = RLM_LOG_LEVEL_INFO;
```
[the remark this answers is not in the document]
desistefanova: I use default_log_level as default on two places.

src/realm_dart_logger.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+
+bool isPortRegistered(Dart_Port key)
```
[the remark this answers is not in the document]
desistefanova: Ok

src/realm_dart_logger.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include <set>
+#include <mutex>
+#include <thread>
+#include <map>
+#include <chrono>
+
+#include "realm_dart_logger.h"
+#include <realm/object-store/c_api/types.hpp>
+
+auto& dart_logger_mutex = *new std::mutex;
```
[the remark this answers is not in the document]
desistefanova: It's a global variable shared between threads. It is initiated once for the whole process.

test/test.dart
desistefanova: This change comes from another PR. They are done in order to print logs in the CI only when an error occurs.

test/realm_test.dart
desistefanova: removed

test/realm_logger_test.dart
desistefanova: Let's discuss this.

test/app_test.dart
desistefanova: Having logger test in different files is good because the tests are running in different isolates. Since these tests don't require to be marked as loggerTest `setupTests(args, loggerTest: true)` because they replace the logger themselves, it is not a problem they to be here.

src/realm_dart_logger.cpp
```diff
@@ -0,0 +1,113 @@
...
+    realm_set_log_callback(realm_dart_logger_callback, default_log_level, nullptr, nullptr);
+    is_core_logger_callback_set = true;
+    return true;
+}
+
+RLM_API void realm_dart_set_log_level(realm_log_level_e level, Dart_Port port)
+{
+    std::lock_guard lock(dart_logger_mutex);
+    if (dart_send_ports[port] != level)
+    {
```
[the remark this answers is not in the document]
desistefanova: looks ok to me

src/realm_dart_logger.cpp
desistefanova: If there is no item for this port in the map, an item will be automatically added even if you try to get it `dart_send_ports[port]`.

src/realm_dart_logger.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+
+bool isPortRegistered(Dart_Port key)
+{
+    return (dart_send_ports.find(key) != dart_send_ports.end());
+}
+
+realm_log_level_e calucale_minimum_log_level()
+{
+    auto min_log_level = default_log_level;
```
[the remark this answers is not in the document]
desistefanova: calucale_minimum_log_level is an internal function that is called by other functions, which are already guarded.

src/realm_dart_logger.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+#include <sstream>
+#include <set>
+#include <mutex>
+#include <thread>
+#include <map>
+#include <chrono>
+
+#include "realm_dart_logger.h"
+#include <realm/object-store/c_api/types.hpp>
```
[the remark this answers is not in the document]
desistefanova: This is old. I used something from utils which were not exposed. I will remove it.

lib/src/native/realm_core.dart
```diff
@@ -51,7 +51,7 @@ late RealmLibrary _realmLib;
 
 final _RealmCore realmCore = _RealmCore();
 
-class _RealmCore {
+class _RealmCore implements RealmCoreScheduler {
```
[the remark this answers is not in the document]
desistefanova: I need `RealmCoreScheduler` in order to define this constructor ` Scheduler.init(RealmCoreScheduler realmCore)` in `scheduler.dart`, since `_RealmCore` is not accessible outside the file `realm_core.dart`.

src/realm_dart_logger.cpp
desistefanova: I changed it to `const_cast`, but I don't understand what you mean with the question.

src/realm_dart_logger.cpp
desistefanova: Why do we need this?

lib/src/scheduler.dart
desistefanova: Commented below.

src/realm_dart_logger.cpp
```diff
@@ -0,0 +1,113 @@
...
+{
+    return (dart_send_ports.find(key) != dart_send_ports.end());
+}
+
+realm_log_level_e calucale_minimum_log_level()
+{
+    auto min_log_level = default_log_level;
+    if (dart_send_ports.size() > 0)
+    {
+        auto min_element = std::min_element(dart_send_ports.begin(), dart_send_ports.end(),
```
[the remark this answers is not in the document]
desistefanova: Let's discuss it.

test/app_test.dart
blagoev: We shouldn't group the tests based on the implementation. We are grouping them based on feature. I don't see the reason to have logger tests all over the place. They should really be done in one file regardless if they are "logger" tests or not or the behavior they are testing. This makes it easy to maintain.

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
blagoev: I don't think we need to do this. We are introducing an interface that only one class will be ever implementing. Also RealmCore is not part of the SDK as a hierarchy class. Its purpose is to initialize and abstract Core native access. Lets discuss this

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
blagoev: What I am saying is these should be initialized only once in the _RealmCore ctor is the better place to do it

src/realm_dart_logger.cpp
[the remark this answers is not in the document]
blagoev: What do you mean. Is the file formatted or not?

src/realm_dart_logger.cpp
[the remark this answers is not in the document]
blagoev: There is no reason introduce a variable which is initialized to an already existing constant and never changed.

src/realm_dart_logger.cpp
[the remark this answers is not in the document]
blagoev: yes, indeed. but why is it created on the heap?

src/realm_dart_logger.cpp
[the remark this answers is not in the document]
blagoev: This could pose a problem if someone refactors the code later. Either we guard it as well or we should have a comment that it needs to be invoked with the lock held. The later option is not ideal but would work

src/realm_dart_logger.cpp
blagoev: std::min_element is in algorithm header

test/test.dart
blagoev: hm strange it shows as changes in this PR.

src/realm_dart_logger.cpp
```diff
@@ -53,7 +47,7 @@ realm_log_level_e calucale_minimum_log_level()
 
 RLM_API void realm_dart_release_logger(Dart_Port port) {
     std::lock_guard lock(dart_logger_mutex);
-    if (isPortRegistered(port))
+    if ((dart_send_ports.find(port) != dart_send_ports.end()))
```
blagoev, line 50: seems too many brackets

blagoev:
```dart
if (dart_send_ports.find(port) != dart_send_ports.end())
```

src/realm_dart_logger.cpp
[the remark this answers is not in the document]
blagoev: // This function should be invoked while holding the lock `dart_logger_mutex`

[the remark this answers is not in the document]
blagoev: we can also try std::recursive_mutex

src/realm_dart_logger.cpp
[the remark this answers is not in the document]
desistefanova: replaced with `std::recursive_mutex dart_logger_mutex;`

src/realm_dart_logger.cpp
desistefanova: Ok, checked with `dart_send_ports.find(port)`

lib/src/realm_class.dart
```diff
@@ -487,10 +487,29 @@ class Realm implements Finalizable {
...
-  static Logger logger = Logger.detached('Realm')
-    ..level = RealmLogLevel.info
-    ..onRecord.listen((event) => print(event));
+  static Logger? _logger;
+
+  /// The logger to use for logging.
+  /// The default log level is [RealmLogLevel.info].
+  /// The level could be changed per isolate.
+  /// To manage the log level at runtime use `Realm.logger.level` setter.
+  /// To listen to the log event use `Realm.logger.onRecord.listen`.
```
blagoev, line 496:
```suggestion
```

lib/src/realm_class.dart
```diff
@@ -487,10 +487,29 @@ class Realm implements Finalizable {
...
 
-  /// The logger to use for logging
-  static Logger logger = Logger.detached('Realm')
-    ..level = RealmLogLevel.info
-    ..onRecord.listen((event) => print(event));
+  static Logger? _logger;
+
+  /// The logger to use for logging.
+  /// The default log level is [RealmLogLevel.info].
+  /// The level could be changed per isolate.
```
blagoev, line 494:
```suggestion
```

lib/src/realm_class.dart
```diff
@@ -487,10 +487,29 @@ class Realm implements Finalizable {
     return realmCore.realmEquals(this, other);
   }
 
-  /// The logger to use for logging
-  static Logger logger = Logger.detached('Realm')
-    ..level = RealmLogLevel.info
-    ..onRecord.listen((event) => print(event));
+  static Logger? _logger;
+
+  /// The logger to use for logging.
```
blagoev, line 492:
```suggestion
  /// The logger to use for Realm logging in this `Isolate`
```

lib/src/realm_class.dart
```diff
@@ -487,10 +487,29 @@ class Realm implements Finalizable {
...
+  /// The default log level is [RealmLogLevel.info].
+  /// The level could be changed per isolate.
+  /// To manage the log level at runtime use `Realm.logger.level` setter.
```
blagoev, line 495:
```suggestion
```

blagoev: I think we shouldn't put `logging` package documentation here. Users should be aware of how `Logging` API is used.

lib/src/realm_class.dart
```diff
@@ -936,3 +907,145 @@ class MigrationRealm extends DynamicRealm {
...
+  ///
+  /// Same as [Level.OFF];
+  static const off = Level.OFF;
+}
+
+/// Represents a logger that manages the realm log level at runtime.
+/// @nodoc
+class _RealmLogger implements Logger {
+  final Logger _logger;
+  bool isUserLogger = false;
```
blagoev, line 970: What is this field for. It seems unused?

lib/src/native/realm_core.dart
```diff
@@ -1665,35 +1669,33 @@ class _RealmCore {
...
+    final message_level = LevelExt.fromInt(level);
+    Realm.logger.log(message_level, message);
+  }
 
-    // Don't do expensive utf8 to utf16 conversion unless needed.
-    if (logger.isLoggable(level)) {
-      logger.log(level, message.cast<Utf8>().toDartString());
+  void _initDefaultLogger() {
+    bool isDefaultLogger = _realmLib.realm_dart_init_default_logger(RealmInternal.defaultLogger.level.toInt());
+    if (isDefaultLogger && !RealmInternal.isUserLogger) {
```
blagoev, line 1680: isUserLogger is seems not needed here. 
It is either isDefaultLogger true or not.

lib/src/realm_class.dart
```diff
@@ -936,3 +907,145 @@ class MigrationRealm extends DynamicRealm {
...
+  _RealmLogger._(Logger logger, {bool userDefinedLogger = false})
```
blagoev, line 971: Private members don't need private ctors

lib/src/realm_class.dart
```diff
@@ -487,10 +487,29 @@ class Realm implements Finalizable {
     return realmCore.realmEquals(this, other);
   }
 
-  /// The logger to use for logging
-  static Logger logger = Logger.detached('Realm')
-    ..level = RealmLogLevel.info
-    ..onRecord.listen((event) => print(event));
+  static Logger? _logger;
```
blagoev, line 490:
```suggestion
  static Logger _logger = _RealmLogger._(Logger.detached('Realm')..level = RealmLogLevel.info);
```

lib/src/realm_class.dart
```diff
@@ -487,10 +487,29 @@ class Realm implements Finalizable {
...
+  /// The level could be changed per isolate.
+  /// To manage the log level at runtime use `Realm.logger.level` setter.
+  /// To listen to the log event use `Realm.logger.onRecord.listen`.
+  /// It is also possible to set a new instance of custom [Logger].
+  static Logger get logger {
+    if (_logger == null) {
+      _logger = RealmInternal.defaultLogger;
+      // In case Realm.logger.onRecord is used in an isolate with no other realm implementations
+      // we must guarantee that the logger is initialized and its level is properly set to realm core.
+      realmCore.setLogLevel(_logger!.level);
```
blagoev, line 503: Do we need this any more. It seems we don't need this call

lib/src/realm_class.dart
```diff
@@ -487,10 +487,29 @@ class Realm implements Finalizable {
...
+    }
+    return _logger!;
+  }
+
+  static set logger(Logger value) {
+    _logger?.clearListeners();
```
blagoev, line 509: No need for _logger to be nullable

lib/src/realm_class.dart
```diff
@@ -660,6 +679,12 @@ extension RealmInternal on Realm {
     return _handle;
   }
 
+  /// Initializes the logger if null. Used in RealmCore.
+  static Logger get defaultLogger => Realm._logger ??= _RealmLogger._(Logger.detached('Realm')..level = RealmLogLevel.info);
+
+  // Returns true if the users have defined their own logger
+  static bool get isUserLogger => Realm._logger != null && (Realm._logger as _RealmLogger).isUserLogger;
```
blagoev, line 686:
```suggestion
  
```

blagoev: This should not be needed

lib/src/realm_class.dart
```diff
@@ -660,6 +679,12 @@ extension RealmInternal on Realm {
     return _handle;
   }
 
+  /// Initializes the logger if null. Used in RealmCore.
+  static Logger get defaultLogger => Realm._logger ??= _RealmLogger._(Logger.detached('Realm')..level = RealmLogLevel.info);
```
blagoev, line 683:
```suggestion
```

blagoev (changes requested):

lib/src/native/realm_core.dart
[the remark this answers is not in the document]
desistefanova: I returned this class back `RealmCoreScheduler` and the Scheduler initialization on `RealmCore`, in order to fix a bug for the following case:
If the users call `Realm(config)` only they should see the default messages printed. The event was not emitted unless they call `Realm.logger` getter.

test/realm_logger_test.dart
desistefanova: It appeared that the tests can not be isolated if we run a new isolate per test, because the first logger is alway replaced in the main isolate by caling `setupTests`.

lib/src/native/realm_core.dart
```diff
@@ -1665,14 +1692,13 @@ class _RealmCore {
     });
   }
 
-  static void _logCallback(Object userdata, int levelAsInt, Pointer<Int8> message) {
-    final logger = Realm.logger;
-    final level = LevelExt.fromInt(levelAsInt);
+  void loggerLogMessage(int level, String message) {
```
nielsenko, line 1695: Not fan of the `logger` prefix

nielsenko, line 1695:
```suggestion
  void logMessage(int level, String message) {
```

lib/src/native/realm_core.dart
```diff
@@ -1665,14 +1692,13 @@ class _RealmCore {
...
+    final message_level = LevelExt.fromInt(level);
+    Realm.logger.log(message_level, message);
+  }
 
-    // Don't do expensive utf8 to utf16 conversion unless needed.
-    if (logger.isLoggable(level)) {
-      logger.log(level, message.cast<Utf8>().toDartString());
-    }
+  void loggerSetLogLevel(Level logLevel) {
```
nielsenko, line 1700:
```suggestion
  void setLogLevel(Level logLevel) {
```

lib/src/realm_class.dart
```diff
@@ -487,10 +487,18 @@ class Realm implements Finalizable {
...
+
+  /// The logger to use for Realm logging in this `Isolate`
+  /// The default log level is [RealmLogLevel.info].
+  static Logger get logger {
+    return _logger;
+  }
+
+  static set logger(Logger value) {
+    _logger.clearListeners();
+    _logger = _RealmLogger(value);
```
nielsenko, line 500: What if `value is _RealmLogger`?

nielsenko (commented):
I few questions and suggestions.

One thing I would like you to think about is:
```dart
Realm.logger = Realm.logger; // double wrap
```

lib/src/native/realm_core.dart
desistefanova: This name `loggerSetLogLevel` was suggested by @blagoev above.

lib/src/realm_class.dart
```diff
@@ -776,60 +784,6 @@ abstract class NotificationsController implements Finalizable {
   }
 }
 
-/// Specifies the criticality level above which messages will be logged
```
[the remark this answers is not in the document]
desistefanova: Ok, moved back to the old place.

lib/src/realm_class.dart
desistefanova: Fixed.

test/realm_logger_test.dart
```diff
@@ -0,0 +1,340 @@
...
+    Map<Level, List<String>> logMessages = await Isolate.run(() async {
+      Realm.logger = Logger.detached(generateRandomString(10))..level = RealmLogLevel.all;
+
+      configuration = AppConfiguration(
+        configuration.appId,
+        baseFilePath: configuration.baseFilePath,
+        baseUrl: configuration.baseUrl,
+      );
+
+      clearCachedApps();
```
blagoev, line 41: This could interfere with other tests.

test/realm_logger_test.dart
```diff
@@ -0,0 +1,340 @@
...
+      final app = App(configuration);
+      final realm = await getIntegrationRealm(app: app);
```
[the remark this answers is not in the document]
blagoev: I suggest we don't use any logic in these test except opening local realm. and closing it in this test isolate explicitly.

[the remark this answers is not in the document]
blagoev: Also using an inMemoryRealm is preferable for these tests

blagoev (changes requested):

blagoev (changes requested):
Our tests rely on Apps which are setup in another Isolate. I think we should not be doing this and rely only on local and perhaps inMemory realms clearing them explicitly in these tests from the separate Isolates

test/realm_logger_test.dart
```diff
@@ -0,0 +1,340 @@
...
+      await realm.syncSession.waitForDownload();
+      return await logMessages;
+    });
+
+    expectLogMessages(logMessages,
+        notExpectedMessagesFromLevels: [RealmLogLevel.fatal, RealmLogLevel.error, RealmLogLevel.warn],
+        expectedMessagesFromLevels: [RealmLogLevel.trace, RealmLogLevel.debug, RealmLogLevel.detail, RealmLogLevel.info]);
+  });
+
+  baasTest('Realm.logger level changed at runtime', (configuration) async {
```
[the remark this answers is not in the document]
desistefanova: It was not working till now. It didn't affect the core logger. Now it is possible to change it up and down.

test/realm_logger_test.dart
```diff
@@ -0,0 +1,340 @@
...
+    Map<String, int> logsCount = await Isolate.run(() async {
+      Map<String, int> results = {};
+      Realm.logger.level = RealmLogLevel.error;
+
+      for (var i = 0; i < 2; i++) {
+        String attemptNum = "Attempt $i";
+        String generatedName = generateRandomString(15);
+        action() async => simulateError(configuration, generatedName);
```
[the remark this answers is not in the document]
desistefanova: `simulateError` this is the method, I will just make it private.

test/realm_logger_test.dart
```diff
@@ -0,0 +1,340 @@
...
+    Map<String, int> results = await Isolate.run(() async {
+      Map<String, int> results = {};
+      action() async => simulateError(configuration, generatedName);
+
+      // Root Isolate  with level Error. Attach to a new logger.
+      final rootIsolate = attachToLoggerBeforeAction(rootIsolateName, action, setNewLogger: true, logLevel: RealmLogLevel.error);
+
+      // Isolate 1 with level Error. Attach to the existing logger.
+      ReceivePort isolate1ReceivePort = ReceivePort();
+      final isolate1 = await Isolate.spawn((SendPort sendPort) async {
```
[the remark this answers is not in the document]
blagoev: I am not sure what you mean, can you link to that code. Also if you use `Isolate.run` you can exit at any point

src/realm_dart_scheduler.cpp
```diff
@@ -40,7 +40,7 @@ struct SchedulerData {
 //This can be invoked on any thread
 void realm_dart_scheduler_free_userData(void* userData) {
     SchedulerData* schedulerData = static_cast<SchedulerData*>(userData);
-
+    realm_dart_release_logger(schedulerData->port);
```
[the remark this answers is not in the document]
blagoev: yes, indeed. resolving

test/realm_logger_test.dart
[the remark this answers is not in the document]
blagoev: What I mean is the test name is not correct. What does it mean to be `at runtime`. Isn't it changed always `at runtime`?

test/realm_logger_test.dart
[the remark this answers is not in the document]
blagoev:
```dart
 action() async => simulateError(configuration, generatedName);
```
this whole line repeats multiple times. Why no use a private method instead?

test/realm_logger_test.dart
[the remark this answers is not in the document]
desistefanova: Ok, I removed ` at runtime`

src/realm_dart.cpp
```diff
@@ -148,3 +148,8 @@ RLM_API void realm_set_auto_refresh(realm_t* realm, bool enable) {
 RLM_API const char* realm_get_library_cpu_arch() {
     return cpuArch.c_str();
 }
+
+RLM_API void realm_dart_log(realm_log_level_e level, const char* message) {
+    auto logger = realm::util::Logger::get_default_logger();
```
blagoev, line 153: We should not be doing this. Why not call the `realm_dart_logger_callback` directly with a message.

blagoev (changes requested):

lib/src/scheduler.dart
```diff
@@ -15,36 +14,39 @@
...
+  int get nativePort => _receivePort.sendPort.nativePort;
 
-    receivePort.handler = (dynamic message) {
-      realmCore.invokeScheduler(handle);
+  Scheduler._() {
+    _receivePortFinalizer.attach(this, _receivePort, detach: this);
+
+    _receivePort.handler = (dynamic message) {
+      if (message is List) {
+        realmCore.loggerLogMessage(message[0] as int, message[1] as String);
```
blagoev, line 35:
```suggestion
       final message_level = LevelExt.fromInt(level);
       Realm.logger.log(message_level, message);
```
I thought that we don't have access to `LevelExt` here, but since we don I suggest we remove the `realmCore.loggerLogMessage` function and do the logging right here. Since `realmCore.loggerLogMessage`  look like we are actually calling Core.

lib/src/native/realm_core.dart
```diff
@@ -1665,14 +1692,19 @@ class _RealmCore {
     });
   }
 
-  static void _logCallback(Object userdata, int levelAsInt, Pointer<Int8> message) {
-    final logger = Realm.logger;
-    final level = LevelExt.fromInt(levelAsInt);
+  void loggerLogMessage(int level, String message) {
```
blagoev, line 1695: I suggest we remove this function. Sorry for missing this previously.

src/realm_dart.cpp
blagoev: I mean expose `realm_dart_log_message_for_testing` and from that call `realm_dart_logger_callback`. Also these should be in `realm_dart_logger.cpp`

test/realm_logger_test.dart
```diff
@@ -0,0 +1,337 @@
...
+import 'test.dart';
+
+Future<void> main([List<String>? args]) async {
+  await setupTests(args);
+
+  test('Realm.logger expected messages for log levels', () async {
+    String generatedMessage = generateRandomString(10);
+    Map<Level, int> logMessages = await Isolate.run(() async {
+      Realm.logger = Logger.detached(generateRandomString(5))..level = RealmLogLevel.all;
+      final allMessages = await _attachToLoggerBeforeAction("root isolate", () => _simulateLogs(generatedMessage));
```
blagoev, line 36:
```suggestion
      final allMessages = await _attachToLoggerBeforeAction("root isolate", _simulateLogs(generatedMessage));
```
My suggestion was to do this instead. Could this work?

test/realm_logger_test.dart
```diff
@@ -0,0 +1,337 @@
...
+      Realm.logger = Logger.detached(generateRandomString(5))..level = RealmLogLevel.all;
+      final allMessages = await _attachToLoggerAfterAction("root isolate", () => _simulateLogs(generatedMessage));
+      return _matchedMessagesCountPerLevel(allMessages, generatedMessage);
+    });
+    _expectAllLevelLogs(logMessages);
+  });
+
+  baasTest('Realm.logger level changed', (configuration) async {
+    Map<String, Map<Level, int>> logMessages = await Isolate.run(() async {
+      Map<String, Map<Level, int>> results = {};
```
blagoev, line 54:
```suggestion
      final results = <String, Map<Level, int>>{};
```

I see we are using explicit types everywhere could we use `final` or `var` instead

test/realm_logger_test.dart
```diff
@@ -0,0 +1,337 @@
...
+      Realm.logger.level = RealmLogLevel.error;
+
+      for (var i = 0; i < 2; i++) {
+        String attemptNum = "Attempt $i";
```
blagoev, line 58:
```suggestion
        final attemptNum = "Attempt $i";
```

test/realm_logger_test.dart
blagoev: The test are harder to read cause it is too verbose on many places

test/realm_logger_test.dart
```diff
@@ -0,0 +1,337 @@
...
+        String generatedMessage = generateRandomString(15);
+        final allMessages = await _attachToLoggerBeforeAction("root isolate", () => _simulateLogs(generatedMessage));
+        final matchedMessages = _matchedMessagesCountPerLevel(allMessages, generatedMessage);
+        results.addEntries([MapEntry(attemptNum, matchedMessages)]);
```
blagoev, line 62:
```suggestion
        results[attemptNum] = matchedMessages;
```

test/realm_logger_test.dart
```diff
@@ -0,0 +1,337 @@
...
+  for (Level level in Level.LEVELS) {
+    final matches = logRecords?.where((element) => element.level == level && element.message.contains(specificMessage));
+    matchedMessages.addEntries([MapEntry(level, matches?.length ?? 0)]);
+  }
+  return matchedMessages;
+}
+
+void _expectLogMessages(
+  Map<Level, int> actualMessages, {
+  List<Level> notExpectedMessagesFromLevels = const [],
```
blagoev, line 270: This API is not great and hard to read. `expectLogMessages` has an argument of `notExpectedMessages`.
Why we need to expect counts when there is not even an argument for count in this function.
This is really making me stop and think how we are validating and what we are validating.

test/realm_logger_test.dart
```diff
@@ -0,0 +1,337 @@
...
+
+      final rootMessages = await rootIsolate;
+      final rootMatches = _matchedMessagesCountPerLevel(rootMessages, generatedName);
+      results.addEntries([MapEntry(rootIsolateName, rootMatches[RealmLogLevel.error]!)]);
+
+      results.addEntries([MapEntry(isolate1Name, await isolate1ReceivePort.first as int)]);
+      results.addEntries([MapEntry(isolate2Name, await isolate2ReceivePort.first as int)]);
+
+      isolate1ReceivePort.close();
+      (await isolate1).kill(priority: Isolate.immediate);
```
blagoev, line 108: Why do we need to do this?

test/realm_logger_test.dart
```diff
@@ -0,0 +1,337 @@
...
+  RealmInternal.log(RealmLogLevel.fatal, message);
+  RealmInternal.log(RealmLogLevel.error, message);
+  RealmInternal.log(RealmLogLevel.warn, message);
+}
+
+Map<Level, int> _matchedMessagesCountPerLevel(List<LogRecord>? logRecords, String specificMessage) {
+  Map<Level, int> matchedMessages = {};
+  for (Level level in Level.LEVELS) {
+    final matches = logRecords?.where((element) => element.level == level && element.message.contains(specificMessage));
+    matchedMessages.addEntries([MapEntry(level, matches?.length ?? 0)]);
```
blagoev, line 263:
```suggestion
    matchedMessages[level] = matches?.length ?? 0;
```

test/realm_logger_test.dart
blagoev: why don't we use Isolate.exit() in each isolate. Also why don't we use `Isolate.run` instead?

test/realm_logger_test.dart
```diff
@@ -0,0 +1,337 @@
...
+  } catch (error) {
+    completer.complete();
+  }
+
+  return events;
+}
+
+Future<List<LogRecord>> _attachToLoggerAfterAction(String isolateName, FutureOr<void> Function() action, {Level? logLevel, bool setNewLogger = false}) async {
+  final completer = Completer<void>();
+  await Future<void>.delayed(Duration(milliseconds: 200));
```
blagoev, line 223: Why do we need this. Seems like a hack.

test/realm_logger_test.dart
blagoev: It is really not clear to me how I will be able to write a new test using this common functions

test/realm_logger_test.dart
```diff
@@ -0,0 +1,337 @@
...
+void _expectLogMessages(
+  Map<Level, int> actualMessages, {
+  List<Level> notExpectedMessagesFromLevels = const [],
+  List<Level> expectedMessagesFromLevels = const [],
+}) {
+  // Check count of various levels
+
+  for (final level in expectedMessagesFromLevels) {
+    final int count = actualMessages[level] ?? 0;
+    expect(count, greaterThan(0), reason: 'There is no messages from $level');
```
blagoev, line 277:
```suggestion
    expect(count, greaterThan(0), reason: 'There are no messages from $level');
```

blagoev (changes requested):
Its very hard to read and understand what each test is trying to test, especially the common private functions.
I have added some suggestions but in general the tests seems not easy to maintain or write new tests

test/realm_logger_test.dart
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  for (final level in notExpectedMessagesFromLevels) {
+    final count = actualMessages[level] ?? 0;
+    expect(count, 0, reason: 'Messages from level $level are not expected:\n  found ${actualMessages[level] ?? 0}');
+  }
+}
+
+void _expectErrorLogs(Map<String, Map<Level, int>> logMessages, String expectedName) {
```
blagoev, line 286: Could we remove the underscore in the common functions in each tests library. It will make it a bit more readable.

test/realm_logger_test.dart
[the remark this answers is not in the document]
blagoev: I have other comments below. resolving this one.

test/realm_logger_test.dart
blagoev: we are doing it on many places

lib/src/realm_class.dart
```diff
@@ -747,6 +755,10 @@ extension RealmInternal on Realm {
       addUnmanagedRealmObjectFromValue(value.value, update);
     }
   }
+
+  static void log(Level level, String message) {
```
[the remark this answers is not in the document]
desistefanova: Why only for testing? You can log message and it will work.

lib/src/scheduler.dart
desistefanova: You could just change the name of the function, instead of adding package dependency.

test/realm_logger_test.dart
desistefanova: It would be nice to have one comon function for validation for all the levels

test/realm_logger_test.dart
desistefanova: I kill them after both isolates complete, because on this way I have 2 errors logged in both isolates, even though they log by one error.

lib/src/realm_class.dart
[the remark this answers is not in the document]
blagoev: you should use `Realm.logger.log`

lib/src/scheduler.dart
blagoev: I am not sure why do you think we will add a package dependency

src/realm_dart.cpp
blagoev: `calucale_minimum_log_level` is not called when logging

desistefanova: Yes, but it is called when setting the level. If we set a wrong level to core you will never catch it, since you will skip the core logger.

nielsenko:
Suggested edit:
```diff
diff --git a/CHANGELOG.md b/CHANGELOG.md
index 21f338dd..24a41e29 100644
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
   `Realm.logger.level` allows changing the log level per isolate.
 * Add logging at the Storage level (Core upgrade).
 * Performance improvement for the following queries (Core upgrade):
-    * Significant (~75%) improvement when counting (query count) the number of exact matches (with no other query conditions) on a String/int/Uuid/ObjectId property that has an index. This improvement will be especially noticiable if there are a large number of results returned (duplicate values).
+    * Significant (~75%) improvement when counting (query count) the number of exact matches (with no other query conditions) on a String/int/Uuid/ObjectId property that has an index. This improvement will be especially noticeable if there are a large number of results returned (duplicate values).
     * Significant (~99%) improvement when querying for an exact match on a Timestamp property that has an index.
     * Significant (~99%) improvement when querying for a case insensitive match on a Mixed property that has an index.
     * Moderate (~25%) improvement when querying for an exact match on a Boolean property that has an index.

```

blagoev (approved):
The tests were rewritten and are passing so this is ready

PR #1228, opened 2023-03-24
Fix test packages ambiguous `throws`

desistefanova:
Hides throws from 'package:test/expect.dart'.
Critical for main branch.

Disable test "Realm path with unicode symbols" for Android and IOS.
Fix job status in case of error while creating the App services

### DIFF
--- a/lib/src/cli/atlas_apps/baas_client.dart
+++ b/lib/src/cli/atlas_apps/baas_client.dart
@@ -16,10 +16,14 @@
 //
 ////////////////////////////////////////////////////////////////////////////////
 
+import 'dart:async';
+
 import 'package:http/http.dart' as http;
 import 'dart:convert';
 
 class BaasClient {
+  static Object? initializationError;
+  static Completer<void>? initializing;
   static const String _confirmFuncSource = '''exports = async ({ token, tokenId, username }) => {
     // process the confirm token, tokenId and username
     if (username.includes("realm_tests_do_autoverify")) {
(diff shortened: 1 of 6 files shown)


test/realm_test.dart
```diff
@@ -1864,7 +1864,7 @@ Future<void> main([List<String>? args]) async {
     var config = Configuration.local([Car.schema], path: "${generateRandomUnicodeString()}.realm");
     var realm = getRealm(config);
     expect(realm.isClosed, false);
-  });
+  }, skip: Platform.isAndroid || Platform.isIOS);
```
nirinchev, line 1867: Why is this needed?

nielsenko: [test_api-0.5.0 ](https://pub.dev/packages/test_api) was released 3 days ago.

blagoev: Shold we create a task to investigate and remove this later?

blagoev: Also if it fails on Android why disable for iOS?

desistefanova: The tests fail for both [Android](https://github.com/realm/realm-dart/actions/runs/4531527657/jobs/7982684657#step:9:1215) and [IOS ](https://github.com/realm/realm-dart/actions/runs/4531527657/jobs/7982684404#step:9:2484) when having unicode symbols in the realm path. I will disable these tests in order to be able to fix the main branch. I [created a new issue for the paths.](https://github.com/realm/realm-dart/issues/1230)

test/test.dart
```diff
@@ -532,26 +532,30 @@ extension on Map<String, String?> {
...
+    client.publicRSAKey = publicRSAKeyForJWTValidation;
 
-  var apps = await client.getOrCreateApps();
-  baasApps.addAll(apps);
-  _baasClient = client;
+    var apps = await client.getOrCreateApps();
+    baasApps.addAll(apps);
+    _baasClient = client;
+  } catch (e) {
+    baasApps.clear();
```
blagoev, line 557: we are silencing all exceptions here. Not sure why we are doing that. Then it seems we allow the run to continue. What is the intention here?

desistefanova: Right, I would've printed the exception. Yes, The tests continue to be run and all the local tests would pass and the sync tests will fail. This will cause the job to fail based on the test report. Otherwise, the tests are not ran and the job doesn't register any errors.

lib/src/cli/atlas_apps/baas_client.dart
```diff
@@ -449,14 +449,14 @@ class BaasClient {
...
     while (true) {
       try {
-        await _patch('groups/$_groupId/apps/$app/services/$mongoServiceId/config', syncConfig);
         await _post('groups/$_groupId/apps/$app/services/$mongoServiceId/default_rule', rules);
+        await _patch('groups/$_groupId/apps/$app/services/$mongoServiceId/config', syncConfig);
         break;
       } catch (err) {
-        if (attempt++ < 24) {
-          print('Failed to update service after ${attempt * 5} seconds. Will keep retrying ...');
+        if (attempt++ < 30) {
```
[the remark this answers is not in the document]
desistefanova: This change, didn't help us. We will investigate the issue with creating the service.

lib/src/cli/atlas_apps/baas_client.dart
```diff
@@ -357,12 +358,17 @@ class BaasClient {
...
-    final dynamic createUserResult = await _post('groups/$_groupId/apps/$appId/users', '{"email": "realm-test@realm.io", "password":"123456"}');
-    print("Create user result: $createUserResult");
+      );
+      await _put('groups/$_groupId/apps/$appId/sync/config', '{ "development_mode_enabled": true }');
+
+      //create email/password user for tests
+      final dynamic createUserResult = await _post('groups/$_groupId/apps/$appId/users', '{"email": "realm-test@realm.io", "password":"123456"}');
+      print("Create user result: $createUserResult");
+    } catch (error) {
+      print(error);
```
blagoev, line 368: Do we need to print the error here?

test/test.dart
```diff
@@ -532,26 +532,31 @@ extension on Map<String, String?> {
...
+        ? BaasClient.docker(baasUrl, differentiator)
+        : BaasClient.atlas(baasUrl, cluster, apiKey!, privateApiKey!, projectId!, differentiator));
+
+    client.publicRSAKey = publicRSAKeyForJWTValidation;
+
+    var apps = await client.getOrCreateApps();
+    baasApps.addAll(apps);
+    _baasClient = client;
+  } catch (error) {
+    print(error);
```
blagoev, line 557: same as above

test/test.dart
```diff
@@ -561,6 +566,7 @@ Future<void> baasTest(
   AppNames appName = AppNames.flexible,
   dynamic skip,
 }) async {
+  if (BaasClient.initializationError != null) throw BaasClient.initializationError!;
```
blagoev, line 569: small one but... can we wrap this in brackets?
```suggestion
  if (BaasClient.initializationError != null){
   throw BaasClient.initializationError!;
   }
```

test/test.dart
```diff
@@ -581,6 +587,7 @@ Future<AppConfiguration> getAppConfig({AppNames appName = AppNames.flexible}) as
 
   final app = baasApps[appName.name] ??
       baasApps.values.firstWhere((element) => element.name == BaasClient.defaultAppName, orElse: () => throw RealmError("No BAAS apps"));
+  if (app.error != null) throw app.error!;
```
blagoev, line 590: same as above nit

test/test.dart
```diff
@@ -532,26 +532,31 @@ extension on Map<String, String?> {
 
 BaasClient? _baasClient;
 Future<void> setupBaas() async {
-  final baasUrl = arguments[argBaasUrl];
-  if (baasUrl == null) {
-    return;
+  if (BaasClient.initializationError != null) return;
```
blagoev, line 535: why do we need this as static on the `BaasClient`? Can we move it to module level variable?

blagoev (changes requested):

lib/src/cli/atlas_apps/baas_client.dart
desistefanova: Yes, I would like to see which of the actions of app services creation are passed successfully before the error.

test/test.dart
desistefanova: Done.

lib/src/cli/atlas_apps/baas_client.dart
blagoev: I agree we can leave it for the test.dart but here I think we should not `print`.
If you want to see which apps succeeded you can print them in the tests.dart

blagoev (approved):

nielsenko (approved):

PR #1231, opened 2023-03-28
Add RealmResults.isValid

nielsenko:
- Set new logger
- Pass dummy handle to for userfdata_ree
- 'throws' is depricated in test/expect.dart
- Added scheduler
- Test added
- Api doc updated
- Api doc updated
- Update changelog
- API doc updated
- Return the old logger at the end of the test
- Fix changeRealmLogLevel
- Add RealmResults.isValid
- Add tests
- Update CHANGELOG


test/results_test.dart
```diff
@@ -662,4 +662,34 @@ Future<void> main([List<String>? args]) async {
...
+    expect(playersAsResults, [alice, bob, carol, dan]);
+
+    realm.write(() => realm.delete(team));
+
+    expect(team.isValid, isFalse); // dead object
+    expect(players.isValid, isFalse); // parent is dead
+    expect(() => players.isEmpty, throwsException);
+    expect(playersAsResults.isValid, isFalse);
+    expect(() => playersAsResults, throwsException);
+    expect(() => playersAsResults.freeze(), throwsException);
```
blagoev, line 693: Should we mark this with a TODO that we will be fixing it not to crash but return an empty list?

blagoev (approved):

test/results_test.dart
nielsenko: Core reversed their thinking on this and decided it should throw, but it is consistent with `isValid` returning `false`

desistefanova (approved):

test/results_test.dart
nielsenko: And now they reversed their thinking again..

PR #1232, opened 2023-03-28
Send Slack message if the CI fails on main branch

desistefanova:
Related to #940 and #971
Slack a message to internal channel in case of CI `failure` in `main` branch.
image


.github/workflows/ci.yml
```diff
@@ -553,3 +553,18 @@ jobs:
...
+    needs:
+      - cleanup-matrix
+    runs-on: ubuntu-latest
+    steps:
+       - name: Slack on fail
+         id: send-slack-message
+         if: false #failure() && ${{ github.head_ref == "" }}
+         uses: act10ns/slack@v2.0.0
+         with:
+              webhook-url: ${{ secrets.SLACK_WEBHOOK_URL }}
```
desistefanova, line 567: @nirinchev can you help us with setting `secrets.SLACK_WEBHOOK_URL` to github secrets?

.github/workflows/ci.yml
```diff
@@ -660,10 +657,34 @@ jobs:
...
+        uses: coverallsapp/github-action@v2
         with:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           parallel-finished: true
 
       - name: Output Coveralls response
         run: echo ${{ steps.publish-coverage.outputs.coveralls-api-result }}
+
+  slack:
+    name: Slack
```
blagoev, line 669: should we use a more meaningful name or add a description.

.github/workflows/ci.yml
```diff
@@ -120,7 +119,6 @@ jobs:
     secrets: inherit
     needs:
       - deploy-flutter-cluster
-      - create-flutter-shared-apps
```
blagoev, line 123: why did we remove these steps. Don't we need it here. If not it is not obvious why?

.github/workflows/ci.yml
```diff
@@ -660,10 +657,34 @@ jobs:
...
       - name: Output Coveralls response
         run: echo ${{ steps.publish-coverage.outputs.coveralls-api-result }}
+
+  slack:
+    name: Slack
+    needs: 
+      - cleanup-dart-matrix
+      - cleanup-flutter-matrix
+    runs-on: ubuntu-latest
+    if: always()
```
blagoev, line 674: why don't we skip here if this is not main branch CI

blagoev (commented):
added some questions mostly.

.github/workflows/ci.yml
desistefanova: Creating apps may fails at the middle of creating. We have to clean at end in any case. Otherwise, `cleanup-flutter-shared-apps` doesn't happen if  `create-flutter-shared-apps` has failed.

.github/workflows/ci.yml
desistefanova: Fixed

.github/workflows/ci.yml
desistefanova: Let's leave it like this, On this way we make it sure that this step will run even though the previous `jobs` and `needs` fail.

.github/workflows/dart-desktop-tests.yml
```diff
@@ -96,7 +96,7 @@ jobs:
       - name: Publish realm_dart coverage
         if: inputs.os == 'linux'
         id: publish-coverage
-        uses: coverallsapp/github-action@a1a1a8a300a7e89df3630639df8fb23de5cc6368
+        uses: coverallsapp/github-action@v2
```
nielsenko, line 99: Why not use a specific hash? We usually prefer specific hashes to avoid a possible supply chain attack, though I agree we rarely do any a thorough validation when upgrading.

.github/workflows/ci.yml
```diff
@@ -674,9 +674,9 @@ jobs:
     if: always()
     steps:
         # Run this action to set env.WORKFLOW_CONCLUSION
-      - uses: technote-space/workflow-conclusion-action@v3
+      - uses: technote-space/workflow-conclusion-action@45ce8e0eb155657ab8ccf346ade734257fd196a5@v3
```
nielsenko:
```suggestion
      - uses: technote-space/workflow-conclusion-action@45ce8e0eb155657ab8ccf346ade734257fd196a5
```

.github/workflows/ci.yml
```diff
@@ -674,9 +674,9 @@ jobs:
...
 
-      - uses: act10ns/slack@v2
+      - uses: act10ns/slack@ed1309ab9862e57e9e583e51c7889486b9a00b0f@v2
```
nielsenko:
```suggestion
      - uses: act10ns/slack@ed1309ab9862e57e9e583e51c7889486b9a00b0f
```

nielsenko (changes requested):

nielsenko (approved):

blagoev (approved):

PR #1233, opened 2023-03-28
Fix crashes while creating CI App Services

desistefanova:
Shares the apps for testing the email credentials - `autoConfirm-shared` and `emailConfirm-shared` between all the environments for a test run (attempt).
Generate splunk query for log visualisation in case of test errors. Using this link we can audit app services' logs even though the app service is deleted.


test/test.dart
```diff
@@ -579,26 +578,33 @@ Future<void> baasTest(
...
   test(name, () async {
-    final config = await getAppConfig(appName: appName);
-    await testFunction(config);
+    final app = baasApps[appName.name] ??
+        baasApps.values.firstWhere((element) => element.name == BaasClient.defaultAppName, orElse: () => throw RealmError("No BAAS apps"));
+    try {
+      if (app.error != null) {
+        throw app.error!;
+      }
+      final config = await getAppConfig(baasUri, app.clientAppId);
```
blagoev, line 590: This is introducing too much work on the caller. Why did we change the signature of `getAppConfig`. It will be much nicer if we keep the getAppConfig by name and do this filtering inside it.

test/test.dart
```diff
@@ -579,26 +578,33 @@ Future<void> baasTest(
...
-Future<AppConfiguration> getAppConfig({AppNames appName = AppNames.flexible}) async {
-  final baasUrl = arguments[argBaasUrl];
-
-  final app = baasApps[appName.name] ??
-      baasApps.values.firstWhere((element) => element.name == BaasClient.defaultAppName, orElse: () => throw RealmError("No BAAS apps"));
-  if (app.error != null) {
-    throw app.error!;
-  }
-
+Future<AppConfiguration> getAppConfig(Uri baasUri, String clientAppId) async {
```
blagoev, line 603: we should keep the getAppConfig signature as described above

test/test.dart
```diff
@@ -623,8 +629,8 @@ Future<String> createServerApiKey(App app, String name, {bool enabled = true}) a
   return await client.createApiKey(baasApp, name, enabled);
 }
 
-Future<Realm> getIntegrationRealm({App? app, ObjectId? differentiator}) async {
-  app ??= App(await getAppConfig());
+Future<Realm> getIntegrationRealm(AppConfiguration appConfiguration, {App? app, ObjectId? differentiator}) async {
```
blagoev, line 632: Its weird that we need to pass appConfig as required but also we can pass the full app as optional. From the caller Its not clear enough what this method does with the arguments supplied.

test/session_test.dart
```diff
@@ -120,7 +120,7 @@ Future<void> main([List<String>? args]) async {
   });
 
   baasTest('SyncSession.waitForUpload with no changes', (configuration) async {
-    final realm = await getIntegrationRealm();
+    final realm = await getIntegrationRealm(configuration);
```
blagoev, line 123: Why do we need to pass the app config as required now.

blagoev (changes requested):

test/test.dart
desistefanova: I agree that this refactoring is not needed anymore, since I changed the idea that I had.

test/test.dart
desistefanova: Done

test/session_test.dart
desistefanova: Not needed.

blagoev (approved):

PR #1235, opened 2023-03-29
overhaul metrics

blagoev:


lib/src/cli/metrics/metrics_command.dart
```diff
@@ -58,10 +62,11 @@ Future<void> uploadMetrics(Options options) async {
   hierarchicalLoggingEnabled = true;
   log.level = options.verbose ? Level.INFO : Level.WARNING;
 
-  final skipUpload = isRealmCI || Platform.environment['CI'] != null || Platform.environment['REALM_DISABLE_ANALYTICS'] != null;
-  if (skipUpload && !isRealmCI) {
-    // skip early
-    log.info('Skipping metrics upload');
+  var skipUpload = (isRealmCI || Platform.environment['CI'] != null || Platform.environment['REALM_DISABLE_ANALYTICS'] != null) &&
+      Platform.environment['REALM_DEBUG_ANALYTICS'] == null;
```
desistefanova, line 66: Could you check this logic? It looks strange. If it is not the CI then 'REALM_DISABLE_ANALYTICS' should be equal to null and not equal to null.

blagoev: the only change here is I introduced REALM_DEBUG_ANALYTICS which invokes the analytics logic but does not send it

desistefanova: Oh, now I see that `REALM_DISABLE_ANALYTICS `<> `REALM_DEBUG_ANALYTICS`. The names are almost equal.

nielsenko (approved):
LGTM. There may be some changes coming to the spec, but let us take that if it happens.

PR #1240, opened 2023-04-01
Added support for binary data type

### LINKED ISSUE #1185: Support Uint8List datatype
### Problem

Using a model containing a Uint8List datatype throws the error `Property type Uint8List not supported`, however, the tests imply this is supported (https://github.com/realm/realm-dart/blob/main/generator/test/good_test_data/all_types.dart).

### Solution

_No response_

### Alternatives

_No response_

Dealbreaker

Local Database only

Jon-Salmon:
Fixes #1185 

I hope this is the right way to go about doing this, I couldn't find any documentation as to how it works under the hood. I also veered on the side of caution concerning when a copy operation was required, which may not be necessary so there could be some optimization potential here.

### DIFF
--- a/lib/src/native/realm_core.dart
+++ b/lib/src/native/realm_core.dart
@@ -2860,6 +2860,14 @@ void _intoRealmValue(Object? value, Pointer<realm_value_t> realm_value, Allocato
         realm_value.ref.values.uuid.bytes[i] = bytes[i];
       }
       realm_value.ref.type = realm_value_type.RLM_TYPE_UUID;
+    } else if (value is Uint8List) {
+      Uint8List binary = value;
+      final result = allocator<Uint8>(binary.length);
+      final Uint8List nativeBinary = result.asTypedList(binary.length);
+      nativeBinary.setAll(0, binary);
+      realm_value.ref.values.binary.data = result.cast();
+      realm_value.ref.values.binary.size = binary.length;
+      realm_value.ref.type = realm_value_type.RLM_TYPE_BINARY;
     } else if (value is DateTime) {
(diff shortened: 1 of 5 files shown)


Jon-Salmon:
I haven't been able to check the test's work due to `No BAAS client` errors.

desistefanova:
Thank you for contributing! 
We will take a look at the PR soon.
In order to run the sync tests you will have to add the following environments variables to your setting.json for VS code.
    "dart.env": {    
        "BAAS_URL": "https://realm.mongodb.com",
        "BAAS_CLUSTER": "<cluster_name>",
        "BAAS_API_KEY": "<public_key>",
        "BAAS_PRIVATE_API_KEY": "<private_key>",
        "BAAS_PROJECT_ID": "<project_id>"
    },
You can run the tests with the local realm without these configurations.

desistefanova (commented):
@Jon-Salmon thanks for the contribution. The PR looks good, but we need to add this type to all the tests. And also the realm model generator should be tested and updated if it is needed. We will probably create another PR from yours in order to be able to complete these tasks.

Jon-Salmon:
Thanks, looking forward to support being added

Jon-Salmon:
@desistefanova Are there any approximate timescales as to when this might be merged?

desistefanova:
> @desistefanova Are there any approximate timescales as to when this might be merged?

Hi @Jon-Salmon! We appreciate your contribution. This PR is in our list, but we have some plan that follow and it takes time. Unfortunately, I'm not able to give you approximate timescales.

blagoev:
I am going to close this PR since a lot more changes are needed to support this. We will be working on providing binary type support in the near future.

PR #1242, opened 2023-04-05
Destructive schema changes prevent deleting realm

desistefanova:
Runs destructive schema change tests after all the other tests have completed.
This test causes terminating flexible sync and could break the app service used by the other tests.
The test is added only to dart tests.
Depends on PR https://github.com/realm/realm-core/pull/6478 and issue https://github.com/realm/realm-core/issues/6482

### DIFF
--- a/.github/workflows/dart-desktop-tests.yml
+++ b/.github/workflows/dart-desktop-tests.yml
@@ -77,7 +77,9 @@ jobs:
           clusterName: ${{ env.BAAS_CLUSTER }}
 
       - name: Run tests
-        run: ${{ inputs.architecture == 'arm' && 'arch -arm64 ' || '' }}dart test -r expanded --coverage ./coverage/ -j 1 --test-randomize-ordering-seed random
+        run: |
+          ${{ inputs.architecture == 'arm' && 'arch -arm64 ' || '' }}dart test -x destructive -r expanded --coverage ./coverage/ -j 1 --test-randomize-ordering-seed random
+          ${{ inputs.architecture == 'arm' && 'arch -arm64 ' || '' }}dart test  test/destructive_schema_test.dart -r expanded
 
       # we're pruning generated files, the cli folder, as well as realm_bindings.dart from our coverage reports
       - name: Generate realm_dart coverage report
--- a/.github/workflows/dart-desktop-tests.yml
+++ b/.github/workflows/dart-desktop-tests.yml
@@ -77,6 +77,8 @@ jobs:
           clusterName: ${{ env.BAAS_CLUSTER }}
 
       - name: Run tests
+       # The first command excludes tests with tag 'destructive'. Te second runs only 'destructive' schema change tests.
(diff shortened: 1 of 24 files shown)


.github/workflows/dart-desktop-tests.yml
```diff
@@ -77,7 +77,9 @@ jobs:
           clusterName: ${{ env.BAAS_CLUSTER }}
 
       - name: Run tests
-        run: ${{ inputs.architecture == 'arm' && 'arch -arm64 ' || '' }}dart test -r expanded --coverage ./coverage/ -j 1 --test-randomize-ordering-seed random
+        run: |
+          ${{ inputs.architecture == 'arm' && 'arch -arm64 ' || '' }}dart test -x destructive -r expanded --coverage ./coverage/ -j 1 --test-randomize-ordering-seed random
+          ${{ inputs.architecture == 'arm' && 'arch -arm64 ' || '' }}dart test  test/destructive_schema_test.dart -r expanded
```
blagoev, line 82: Why do we invoke it two times. Isn't the above enough?

test/destructive_schema_test.dart
```diff
@@ -0,0 +1,86 @@
...
+
+@RealmModel()
+@MapTo("Card")
+class _CardV2 {
+  @PrimaryKey()
+  @MapTo('_id')
+  late Uuid id;
+}
+
+// Destructive schema change causes the server to force the translator to restart.
```
blagoev, line 43: Can this fail negatively impact the other workflows running in parallel and reusing the apps?

.github/workflows/dart-desktop-tests.yml
desistefanova: The first command excludes tests with tag 'destructive'. After all the other tests pass then we run another command for testing only the destructive tests. Otherwise, 'destructive' schema change tests terminates the flexible sync and this breaks the rest of the tests. I will put this as a comment in the yml file.

test/destructive_schema_test.dart
desistefanova: Yes, see above.

.github/workflows/dart-desktop-tests.yml
desistefanova: I run the 'destructive' tests only for Dart. I think it is enough.

test/destructive_schema_test.dart
blagoev: How are we making sure we don't impact other workflows here?

.github/workflows/dart-desktop-tests.yml
blagoev: My question is aren't the test with `destructive` tag the same as the tests in `test/destructive_schema_test.dart`?

blagoev: ok this actually excluded in the first run. resolving

blagoev (approved):

PR #1251, opened 2023-04-19
Add a force flag to install

nielsenko:
This allows us to force install the shared library even if we would
normally skip it. This is useful if you wan't to run tests against a
released version without actually building the realm again.


blagoev:
This will work as long as the binary version is the same. I don't see why not have it

blagoev (approved):

PR #1267, opened 2023-05-08
Override first, last and single on RealmResults

nielsenko:
Avoid instantiating a _RealmResultsIterator for first and
single calls, and avoid iterating the entire collection for last.


lib/src/results.dart
```diff
@@ -68,10 +65,23 @@ class RealmResults<T extends Object?> extends collection.IterableBase<T> with Re
...
 
-  /// The number of values in this `Results` collection.
   @override
   int get length => realmCore.getResultsCount(this);
 
+  @override
+  T get first => this[0];
+
+  @override
+  T get last => this[length - 1];
```
desistefanova, line 75: Is this tested with `length = 0`?

nielsenko: It will throw `RealmException`

nielsenko: Same with `first`

lib/src/results.dart
```diff
@@ -53,11 +52,9 @@ class RealmResults<T extends Object?> extends collection.IterableBase<T> with Re
     }
   }
 
-  /// `true` if the `Results` collection is empty.
   @override
   bool get isEmpty => length == 0;
 
-  /// Returns a new `Iterator` that allows iterating the elements in this `RealmResults`.
```
blagoev, line 60: While `RealmResults` is extending `Iterable` it should not try hiding fact that it is a `RealmResults` object implementation. In this line of thought why not leave the comments mentioning it.

test/results_test.dart
```diff
@@ -662,4 +662,42 @@ Future<void> main([List<String>? args]) async {
...
     }
   });
+
+  test('RealmResults.first throws on empty', () async {
+    final config = Configuration.local([Dog.schema, Person.schema]);
+    final realm = getRealm(config);
+
+    final empty = Iterable<int>.empty();
+    expect(() => empty.first, throwsA(isA<StateError>()));
+    expect(() => realm.all<Dog>().first, throwsA(isA<RealmException>())); // TODO: Should this be StateError?
```
blagoev, line 672: I think we should leave it as `RealmException` since it is a Realm class

test/results_test.dart
```diff
@@ -662,4 +662,42 @@ Future<void> main([List<String>? args]) async {
...
+    expect(() => realm.all<Dog>().last, throwsA(isA<RealmException>())); // TODO: Should this be StateError?
+  });
+
+  test('RealmResult.single throws on empty', () {
+    final config = Configuration.local([Dog.schema, Person.schema]);
+    final realm = getRealm(config);
+
+    final empty = Iterable<int>.empty();
+    expect(() => empty.single, throwsA(isA<StateError>()));
+    expect(() => realm.all<Dog>().single, throwsA(isA<StateError>()));
```
blagoev, line 690: I think this should throw `RealmException`

blagoev (changes requested):
added some suggestions on the exceptions mostly. I think we should be doing these optimizations.

desistefanova (approved):

test/results_test.dart
nielsenko: throws `RealmStateError`

lib/src/results.dart
```diff
@@ -68,10 +65,23 @@ class RealmResults<T extends Object?> extends collection.IterableBase<T> with Re
...
+  @override
+  T get first => this[0];
+
+  @override
+  T get last => this[length - 1];
+
+  @override
+  T get single {
+    if (length != 1) {
+      throw StateError('Expected exactly one element but was $length');
```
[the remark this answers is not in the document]
nielsenko: throw `RealmStateError`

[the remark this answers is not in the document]
blagoev: ok. I see we want to mimic the state's Collection.single. Then should we change the text to be 

`Bad state: Too many elements` as the state one does.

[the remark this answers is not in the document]
blagoev: and `Bad state: No element`

test/results_test.dart
```diff
@@ -741,4 +741,42 @@ Future<void> main([List<String>? args]) async {
     final items = realm.query<Product>(ids);
     expect(items.length, 0);
   });
+
+  test('RealmResults.first throws on empty', () async {
+    final config = Configuration.local([Dog.schema, Person.schema]);
+    final realm = getRealm(config);
+
+    final empty = Iterable<int>.empty();
+    expect(() => empty.first, throwsA(isA<StateError>()));
```
blagoev, line 750: why do we test the state's Inerrable behavior? I suggest we test only the Realm implementation

test/results_test.dart
```diff
@@ -741,4 +741,42 @@ Future<void> main([List<String>? args]) async {
...
+
+  test('RealmResult.single throws on two items', () {
+    final config = Configuration.local([Dog.schema, Person.schema]);
+    final realm = getRealm(config);
+
+    final twoItems = [1, 2];
+    expect(() => twoItems.single, throwsA(isA<StateError>()));
+
+    realm.write(() => realm.addAll([Dog('Fido'), Dog('Spot')]));
+    expect(() => realm.all<Dog>().single, throwsA(isA<StateError>()));
```
blagoev, line 780:
```suggestion
    expect(() => realm.all<Dog>().single, throwsA<RealmStateError>());
```
we have a utility method for exceptions also I suggest we test the exact error we are throwing.

test/results_test.dart
```diff
@@ -741,4 +741,42 @@ Future<void> main([List<String>? args]) async {
...
+    expect(() => empty.single, throwsA(isA<StateError>()));
+    expect(() => realm.all<Dog>().single, throwsA(isA<RealmStateError>()));
+  });
+
+  test('RealmResult.single throws on two items', () {
+    final config = Configuration.local([Dog.schema, Person.schema]);
+    final realm = getRealm(config);
+
+    final twoItems = [1, 2];
+    expect(() => twoItems.single, throwsA(isA<StateError>()));
```
blagoev, line 777: I suggest we remove these tests of the state's behavior

test/results_test.dart
```diff
@@ -741,4 +741,42 @@ Future<void> main([List<String>? args]) async {
...
+    expect(() => realm.all<Dog>().first, throwsA(isA<RealmException>()));
+  });
+
+  test('RealmResults.last throws on empty', () async {
+    final config = Configuration.local([Dog.schema, Person.schema]);
+    final realm = getRealm(config);
+
+    final empty = Iterable<int>.empty();
+    expect(() => empty.last, throwsA(isA<StateError>()));
+    expect(() => realm.all<Dog>().last, throwsA(isA<RealmException>()));
```
blagoev, line 760: These two lines show the discrepancy. If we try to mimic the state Iterable then `last` should throw 
`Bad state: No element` for Realm as well. At least it should be `RealmStateError` when empty

blagoev: I believe `first` as well

blagoev (changes requested):

desistefanova (approved):

test/results_test.dart
nielsenko: A reason not to do that is that it will require an extra call to native to check the `length`, and another error is already raised if not. But if we deem the extra call acceptable, then I can easily add it like in `single` (where it is needed to raise some error at all)

nielsenko: Latest commit test raises `RealmStateError` on `first` and `last` as well, if `RealmResults` is empty

test/results_test.dart
nielsenko: Just to illustrate behaviour for regular `Iterable`.. will add a comment

lib/src/results.dart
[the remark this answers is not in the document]
nielsenko: Done

test/results_test.dart
nielsenko: Changed

blagoev (approved):

PR #1275, opened 2023-05-12
Upper sdk constraint now <4.0.0

nielsenko:


desistefanova (approved):

desistefanova (commented):
Will be nice to have some description in the PR.
Are we going to mention this in the changelog?

blagoev (approved):
Perhaps we could mention in changelog that we are compatible with Dart 3.0 and Flutter 3.10. Just to make it more visible and explicit

PR #1290, opened 2023-05-24
remove _RealmLogger wrapper

blagoev:
Removes `_RealmLogger` wrapper. 

Reminder to all: We should really be careful not to implement classes from dependencies unless we are sure we can't be broken by a minor dependency update.

### DIFF
--- a/lib/src/realm_class.dart
+++ b/lib/src/realm_class.dart
@@ -501,8 +501,11 @@ class Realm implements Finalizable {
     }
 
     _logger.clearListeners();
+    realmCore.realmLoggerLevelChangedSubscipiton.cancel();
     _logger = value;
-    _logger.onLevelChanged.listen((logLevel) => realmCore.loggerSetLogLevel(logLevel ?? RealmLogLevel.off, scheduler.nativePort));
+    realmCore.loggerSetLogLevel(_logger.level, scheduler.nativePort);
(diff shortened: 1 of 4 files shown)


lib/src/realm_class.dart
```diff
@@ -496,8 +496,13 @@ class Realm implements Finalizable {
...
 
   static set logger(Logger value) {
+    if (_logger == value) {
+      return;
+    }
+
     _logger.clearListeners();
-    _logger = (value is _RealmLogger) ? value : _RealmLogger(value);
+    _logger = value;
+    _logger.onLevelChanged.listen((logLevel) => realmCore.loggerSetLogLevel(logLevel ?? RealmLogLevel.off, scheduler.nativePort));
```
nielsenko, line 505: We should track this subscription and cancel it when a new logger is set, so that we don't have another logger which log level can change the one in core?

Ie. to avoid 
```dart
final old = Realm.logger;
Realm.logger = CustomLogger();
old.logLevel = Level.ALL;
```

desistefanova, line 505: Call `realmCore.loggerSetLogLevel` before attaching to the listener for the level.

blagoev: good catch. fixing

blagoev: fixing

PR #1293, opened 2023-05-29
Backlinks getters throws for unmanaged objects

### LINKED ISSUE #1106: Error occurs when a backlink iterable is empty
### What happened?

For example:

```
@Backlink(#items)
late Iterable<$ExampleList> exampleLists;
```

When an item has no relationship with `ExampleList`, it seems not possible to access `exampleLists`.

`print(item.exampleLists.isEmpty);
`

results in:

```
Exception has occurred.
_CastError (type 'Null' is not a subtype of type 'RealmResults<ExampleList>' in type cast)
```

### Repro steps

See above

### Version

2.19.0-444.3.beta / 3.7.0-1.3.pre

Local Database only

Flutter Application

### Client OS and version

0.9.0+rc

### Code snippets

_No response_

_No response_

### Relevant log output

_No response_

desistefanova:
Using backlinks is only possible for managed(persisted) objects.
Fixes #1106
Fix test: "Realm.open (flexibleSync) - listen for download progress on an empty realm"


nielsenko (changes requested):
We should not change the output of the generator for this in my opinion. This should be handled in the accessor.

desistefanova:
> We should not change the output of the generator for this in my opinion. This should be handled in the accessor.

How could we know in the accessor that this is a backlink?

generator/lib/src/realm_field_info.dart
```diff
@@ -80,7 +80,12 @@ class RealmFieldInfo {
   Iterable<String> toCode() sync* {
     final getTypeName = type.isRealmCollection ? basicMappedTypeName : basicNonNullableMappedTypeName;
     yield '@override';
-    yield "$mappedTypeName get $name => RealmObjectBase.get<$getTypeName>(this, '$realmName') as $mappedTypeName;";
+    if (isRealmBacklink) {
+      final backlinkGetter = "RealmObjectBase.get<$getTypeName>(this, '$realmName')";
+      yield "$mappedTypeName get $name => ($backlinkGetter as $mappedTypeName?)?? $mappedTypeName();";
+    } else {
+      yield "$mappedTypeName get $name => RealmObjectBase.get<$getTypeName>(this, '$realmName') as $mappedTypeName;";
```
nielsenko, line 87:
```suggestion
      yield "$mappedTypeName get $name => $backlinkGetter as $mappedTypeName;";
```

nielsenko (approved):
You are right about not being able to do this in the accessor with the current design.

blagoev (changes requested):
.NET throws an exception. Why are we diverging on this?

desistefanova:
> #1106

We have an user request with a reasonable argument: https://github.com/realm/realm-dart/issues/1106#issuecomment-1387179707

desistefanova:
> .NET throws an exception. Why are we diverging on this?

We have an user request with a reasonable argument: https://github.com/realm/realm-dart/issues/1106#issuecomment-1387179707

nielsenko (changes requested):
Missing the changes to `RealmGenerator`

nielsenko (approved):

generator/lib/src/realm_field_info.dart
```diff
@@ -80,7 +80,8 @@ class RealmFieldInfo {
   Iterable<String> toCode() sync* {
     final getTypeName = type.isRealmCollection ? basicMappedTypeName : basicNonNullableMappedTypeName;
     yield '@override';
-    yield "$mappedTypeName get $name => RealmObjectBase.get<$getTypeName>(this, '$realmName') as $mappedTypeName;";
+    final validateBacklink = isRealmBacklink ? ", validateBacklink: true" : "";
+    yield "$mappedTypeName get $name => RealmObjectBase.get<$getTypeName>(this, '$realmName'$validateBacklink) as $mappedTypeName;";
```
blagoev, line 84: This could check if the object isManaged and throw here.

test/realm_test.dart
```diff
@@ -1345,11 +1345,14 @@ Future<void> main([List<String>? args]) async {
...
-
+    final completer = Completer<void>();
     var syncedRealm = await getRealmAsync(configuration, onProgressCallback: (syncProgress) {
       transferredBytes = syncProgress.transferredBytes;
+      if (syncProgress.transferredBytes == syncProgress.transferableBytes) {
+        completer.complete();
+      }
     });
-
+    completer.future.timeout(Duration(milliseconds: 300), onTimeout: () => throw Exception("onProgressCallback does not happen."));
```
blagoev, line 1355:
```suggestion
    completer.future.timeout(Duration(milliseconds: 300), onTimeout: () => throw Exception("onProgressCallback did not happen."));
```

blagoev (changes requested):

generator/lib/src/realm_field_info.dart
desistefanova: The purpose was the check and the exception to be in one place. Otherwise this will be repeated for all the backlink properties:
```dart
RealmResults<Source> get oneToMany {
    if (!isManaged) {
      throw RealmException("Using backlinks is only possible for managed(persisted) objects.");
    }
    return RealmObjectBase.get<Source>(this, 'oneToMany') as RealmResults<Source>;
  }
```

lib/src/realm_object.dart
```diff
@@ -313,7 +313,10 @@ mixin RealmObjectBase on RealmEntity implements RealmObjectBaseMarker, Finalizab
   };
 
   /// @nodoc
-  static Object? get<T extends Object?>(RealmObjectBase object, String name) {
+  static Object? get<T extends Object?>(RealmObjectBase object, String name, {bool validateBacklink = false}) {
```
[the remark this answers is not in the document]
desistefanova: I agree, that it is not good we to change the interface, that's why it was added as a discussion point. 
An optional parameter wouldn't be a problem, but still this parameter is too specific for such a general method.
Unfortunately, we don't have any information about `RealmPropertyType` and this is the reason I have added this parameter.

generator/lib/src/realm_field_info.dart
blagoev: why would be a problem to be repeated? This is generated code.

lib/src/realm_object.dart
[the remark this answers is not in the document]
blagoev: it is not easy to generalize it and if we do it it will negatively impact the benefit of doing it. Hence I suggested to check in-place in the generated code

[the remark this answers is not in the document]
desistefanova: Ok. This is done.

generator/lib/src/realm_field_info.dart
desistefanova: Ok, Done.

generator/lib/src/realm_field_info.dart
```diff
@@ -80,8 +80,13 @@ class RealmFieldInfo {
   Iterable<String> toCode() sync* {
     final getTypeName = type.isRealmCollection ? basicMappedTypeName : basicNonNullableMappedTypeName;
     yield '@override';
-    final validateBacklink = isRealmBacklink ? ", validateBacklink: true" : "";
-    yield "$mappedTypeName get $name => RealmObjectBase.get<$getTypeName>(this, '$realmName'$validateBacklink) as $mappedTypeName;";
+    if (isRealmBacklink) {
+      yield "$mappedTypeName get $name {";
+      yield "if(!isManaged){throw RealmException('Using backlinks is only possible for managed(persisted) objects.');}";
```
blagoev, line 85: I would suggest formatting this a bit here. The generator code should be readable by us

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@
 * None
 
 ### Fixed
-* None
+* Getting `Backlink` properties of not persisted objects will throw exception: "Using backlinks is only possible for managed(persisted) objects." ([#1293](https://github.com/realm/realm-dart/pull/1293)).
```
```suggestion
* Getting `Backlink` properties of unmanaged Realm objects will throw exception: "Using backlinks is only possible for managed objects." ([#1293](https://github.com/realm/realm-dart/pull/1293)).
```

generator/lib/src/realm_field_info.dart
blagoev: no need to use persisted when we say managed and this is the word we are using across the docs and api docs imo
```suggestion
      yield "if(!isManaged){throw RealmException('Using backlinks is only possible for managed objects.');}";
```

generator/lib/src/realm_field_info.dart
```diff
@@ -80,7 +80,13 @@ class RealmFieldInfo {
   Iterable<String> toCode() sync* {
     final getTypeName = type.isRealmCollection ? basicMappedTypeName : basicNonNullableMappedTypeName;
     yield '@override';
-    yield "$mappedTypeName get $name => RealmObjectBase.get<$getTypeName>(this, '$realmName') as $mappedTypeName;";
+    if (isRealmBacklink) {
+      yield "$mappedTypeName get $name {";
+      yield "if(!isManaged){throw RealmException('Using backlinks is only possible for managed(persisted) objects.');}";
```
blagoev, line 85: Should this be a `RealmError`. I think we are pretty incosistent with these validations but the question here is still valid I think?

blagoev (changes requested):

blagoev (approved):

nielsenko (approved):
Okay

PR #1300, opened 2023-05-30
Add support for FTS

nirinchev:
Adds an optional positional argument to the `@Indexed` annotation that enables running `TEXT` queries:

```dart
@RealmModel()
class _Book {
  late String title;

  @Indexed(RealmIndexType.fullText)
  late String summary;
}

final books = realm.query<Book>("summary TEXT \$0", "fantasy novel")
```

Fixes https://github.com/realm/realm-dart/issues/1248.


generator/test/error_test_data/not_an_indexable_type.expected
```diff
@@ -1,4 +1,4 @@
-Realm only supports the @Indexed() annotation on fields of type
+Realm only supports the @Indexed annotation on fields of type
```
nielsenko, line 1: You actually have to write the `()` in dart as you const constructing an object, but I guess it is okay for error message

nielsenko (approved):
Assuming you are right about `DateTime` not being a valid primary key, which is odd given that it is indexable.

generator/test/error_test_data/not_an_indexable_type.expected
nirinchev: Yeah, I removed it from the error because now you can also do `@Indexed(RealmIndexType.fullText)` and figured it may be more confusing if the error message removed the argument than if it didn't include the parentheses at all.

CHANGELOG.md
```diff
@@ -1,10 +1,12 @@
...
 
 ### Enhancements
-* None
+* Added support for Full-Text search (simple term) queries. ([#1248](https://github.com/realm/realm-dart/issues/1248))
+  * To enable FTS queries on string properties, add the `@Indexed(RealmIndexType.fullText)` annotation.
+  * To run queries, use the `TEXT` operator: `realm.all<Book>().query("description TEXT \$0", "fantasy novel")`.
 
 ### Fixed
-* None
+* Fixed an issue where the generator would incorrectly consider a `DateTime` field a valid primary key.
```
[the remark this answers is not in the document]
nirinchev: It has never been supported as far as I know. No idea if there are technical reasons for it or if people never wanted it.

generator/lib/src/field_element_ex.dart
```diff
@@ -137,27 +137,58 @@ extension FieldElementEx on FieldElement {
...
-        final annotation = (primaryKey ?? indexed)!.annotation;
-        final listOfValidTypes = RealmPropertyType.values //
-            .map((t) => t.mapping)
-            .where((m) => m.indexable && ((m.type != bool && m.type != RealmValue) || primaryKey == null))
-            .map((m) => m.type);
+        if (indexType == RealmIndexType.none) {
+          throw RealmInvalidGenerationSourceError("Don't use RealmIndexType.none for @Indexed annotations",
+              element: this,
+              primarySpan: typeSpan(file),
+              primaryLabel: '@Indexed(RealmIndexType.none) is not valid annotation here',
```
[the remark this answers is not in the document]
nirinchev: It's possible if we introduce two types for that, but it seemed a bit of an overkill. The reason why I think it makes sense to have `.none` there is because the schema can also be introspected from the Realm file. So in case you're writing an app that can open a Realm file dynamically, we should be able to communicate the state of the indexes on each column.

CHANGELOG.md
```diff
@@ -1,10 +1,12 @@
 ## vNext (TBD)
 
 ### Enhancements
-* None
+* Added support for Full-Text search (simple term) queries. ([#1248](https://github.com/realm/realm-dart/issues/1248))
```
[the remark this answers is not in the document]
nirinchev: The practice in other SDKs is to link to the issue being fixed for a couple of reasons:
1. A fix may be split over multiple PRs.
2. Issues created by users typically have more context about the user-facing impacts of the bug.
3. Typically you can more easily get to the PR by doing a git blame than to the issue.

nirinchev:
The macOS test failure is a timeout of the tests after 45 minutes, while the Ubuntu 20 failure is the `Realm.logger supports logging from multiple isolates` test and it seems unrelated to these changes.

generator/lib/src/field_element_ex.dart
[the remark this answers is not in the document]
desistefanova: If I get it correctly, `@Indexed(RealmIndexType.none)` is equivalent to not having `@Indexed` annotation at all, right? Are we going to require  from our users to place this annotation on every field that is not indexed?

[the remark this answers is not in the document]
nielsenko: No, and this PR does not imply that.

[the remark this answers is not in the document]
nirinchev: @nielsenko is correct - `@Indexed(RealmIndexType.none)` is incorrect usage of the API which will result in a generator error. But the `.none` option on `RealmIndexType` exists so that you can do something like:

```dart
var property = realm.schema[0].properties[0];
if (property.indexType == RealmIndexType.none) {
  console.log("Property ${property.name} is not indexed");
}
```

I.e. if you're introspecting the schema, you may want to know the index status of the properties and `.none` is valid status then.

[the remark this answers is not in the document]
blagoev: I totally agree with @desistefanova here. 
`@Indexed(RealmIndexType.none)` is very weird API and we should disallow writing it. This should be a static restriction instead of runtime one. Btw I am really surprised @nielsenko is agreeing with this as a strong proponent of static compile time APIs. This one is really no brainer to change to a compile time one just remove the `RealmIndexType.none` from the enum.

As for the case @nirinchev mentions,  if one wants to inspect a property at runtime we will return `null` for its `index` flag. 
```dart
var property = realm.schema[0].properties[0]; 
if (property.indexType == null) { //or if (property.index == null) { or something similarly named.
console.log("Property ${property.name} is not indexed"); 
}
```

common/lib/src/realm_types.dart
```diff
@@ -54,15 +55,43 @@ enum RealmPropertyType {
...
 
   const RealmPropertyType([this.mapping = const Mapping<Never>()]);
 
   final Mapping<dynamic> mapping;
 }
 
+/// Describes the indexing mode for properties annotated with the @Indexed annotation.
+enum RealmIndexType {
+  /// Indicates that the property is not indexed
+  none,
```
blagoev, line 70: I agree with @desistefanova. This should be not be introduced.

generator/lib/src/field_element_ex.dart
```diff
@@ -137,27 +137,58 @@ extension FieldElementEx on FieldElement {
...
+        if (indexType == RealmIndexType.fullText && type.realmType != RealmPropertyType.string) {
+          throw RealmInvalidGenerationSourceError('Cannot add full-text index on a non-string property',
+              element: this,
+              primarySpan: typeSpan(file),
+              primaryLabel: 'Cannot use RealmIndexType.fullText for property of type $modelTypeName',
+              todo: 'Change the index type to general or change the property type to string');
+        }
+
+        if (type.realmType?.mapping.indexable != true) {
+          final listOfValidTypes = RealmPropertyType.values //
```
blagoev, line 179:
```suggestion
          final listOfValidTypes = RealmPropertyType.values
```
for me the formatter does not require these `//` at the end

generator/lib/src/realm_field_info.dart
```diff
@@ -27,15 +27,15 @@ class RealmFieldInfo {
   final FieldElement fieldElement;
   final String? mapTo;
   final bool isPrimaryKey;
-  final bool indexed;
+  final RealmIndexType indexType;
```
blagoev, line 30: we could preserve the full `Indexed` annotation here
```suggestion
  final Indexed? index;
```

generator/test/error_test_data/bool_not_allowed_as_primary_key.expected
```diff
@@ -12,5 +12,5 @@ in: asset:pkg/test/error_test_data/bool_not_allowed_as_primary_key.dart:9:8
 9 │   late bool bad;
   │        ^^^^ bool is not a valid type here
   ╵
-Change the type of 'bad', or remove the @PrimaryKey() annotation
+Change the type of 'bad' or remove the @PrimaryKey annotation
```
blagoev, line 15: good improvement

lib/src/realm_property.dart
```diff
@@ -35,7 +35,11 @@ class SchemaProperty {
   final bool primaryKey;
 
   /// `true` if the property is indexed
-  final bool indexed;
+  @Deprecated("Use indexType instead")
+  bool get indexed => indexType == RealmIndexType.general;
+
+  /// Returns the index type for this property
+  final RealmIndexType indexType;
```
blagoev, line 42:
```suggestion
  final Indexed? index;
```

blagoev: we could expose the full annotation now that it has more data

test/indexed_test.dart
```diff
@@ -177,4 +201,59 @@ void main([List<String>? args]) {
...
+      final value = realm.add(ObjectWithFTSIndex(title, summary));
+      value.nullableSummary = summary;
+      return value;
+    }
+
+    realm.write(() {
+      final objectWithDefaults = addObject('N/A', '');
+      objectWithDefaults.nullableSummary = null;
+
+      addObject(animalFarm,
```
blagoev, line 227: it is educational, but do we really need so much text to test FTS

blagoev (changes requested):
Added some suggestions. And I think we should remove `RealmIndexType.none`. It adds very little value and makes the API weird and prone to runtime errors.

generator/lib/src/field_element_ex.dart
```diff
@@ -137,27 +137,58 @@ extension FieldElementEx on FieldElement {
           );
         }
         */
+
+        if (type.realmType?.mapping.canBePrimaryKey != true) {
+          final file = span!.file;
+          final listOfValidTypes = RealmPropertyType.values //
```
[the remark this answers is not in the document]
nirinchev: I just moved this code and it had the slashes at the end. I have no strong opinions about the formatting of it, but if I remove them, the linter will move the entire statement on a single line.

generator/lib/src/field_element_ex.dart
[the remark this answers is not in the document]
nirinchev: I generally dislike exposing nullable enums as `null` can just be represented as a value within the enum. That being said, it's not a hill I'm willing to die on, so will change it.

generator/lib/src/field_element_ex.dart
nirinchev: Same comment as above - it's moved code and removing the slashes moves everything on a single line.

generator/lib/src/realm_field_info.dart
nirinchev: I don't think we should. We don't preserve the annotations for other things, such as `mapTo` or `isPrimaryKey`, so I'd rather extract the meaningful information.

test/indexed_test.dart
nirinchev: I was too lazy to write new tests and copied the ones from the .NET SDK. They did reveal several bugs with the original implementation, so I'd say they were useful at the time. I'm not sure if they continue to be useful and probably a more compact search text would be just fine. That being said, it'll be a little bit extra effort to come up with equivalent tests and I'm not sure the benefit is there.

generator/lib/src/field_element_ex.dart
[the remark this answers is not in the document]
blagoev: for me it does not do that. at least the VS code formatter. not sure what you mean by `linter`. Lets leave as is

generator/lib/src/field_element_ex.dart
blagoev: lets leave it. for me it does not move.

generator/lib/src/realm_field_info.dart
blagoev: the annotation `Indexed` is just a data bag class in here

blagoev: instead of having a nullable enum we could just have an annotation instance here

test/indexed_test.dart
blagoev: ok lets copy another malpractice from the .NET SDK :) At least it will make everyone know what `silmarillion` is.

generator/lib/src/field_element_ex.dart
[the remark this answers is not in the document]
nielsenko: For the record I still am a strong proponent for static compile time APIs 😄 .. I usually don't use nullable enums either, preferring to have `.none` like value instead, but perhaps just having a `RealmIndexType? indexType` on `Property` is better in this case?

Wrt. to misuse, I just want to point out that the generator actually did catch it, if anyone used `RealmIndexType.none`.

generator/lib/src/field_element_ex.dart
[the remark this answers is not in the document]
nirinchev: Sorry, meant "formatter" - this seems to be caused by the dart extension for VS Code on mac. I haven't done any special configuration for it, so I guess it picked up some rules from the project itself. Not sure why there would be a difference between macOS and Windows - probably something to investigate within the team at some point 🤷‍♂️

generator/lib/src/realm_field_info.dart
nirinchev: Right, but the only meaningful thing it contains is the `indexType`. You could make the same argument for the `MapTo` annotation - it is similarly a data bag with a single field. In any case, I would prefer that we extract the annotation data since it matches better the expectations I have of `RealmFieldInfo` - it's just a container for all the information that describes a field in Realm. The fact certain information is obtained through annotations is kind of an implementation detail and not something we need here.

test/indexed_test.dart
nirinchev: The .NET SDK: educating developers since 2015 😜

generator/lib/src/realm_field_info.dart
blagoev: Ok lets keep it

blagoev (approved):

PR #1310, opened 2023-06-07
Use PartBuilder instead of SharedPartBuilder

nielsenko:
This will allow other builders to consume output of the realm generator.
Generated files are now suffixed with `.realm.dart` instead of `.g.dart`.

Due to the behaviour of `PartBuilder` source files needs to have a `part` declaration (`SharedPartBuilder` is more lenient for some reason). This means that the trick we used in the tests of not specifying the part no longer works. Instead a lot mock files has been added, of the form `<x>.realm.dart` containing:
```dart
// MOCK FILE!
part of '<x>.dart';
```
mostly to avoid a lot of red squiggles in the project due to parse errors.

`test_utils.dart` has been reworked quite a bit in the process, and now contains a `testCompile` method for easy testing of the builder.

Related to: #1254


.vscode/settings.json
```diff
@@ -147,5 +147,6 @@
     "rope": "cpp",
     "slist": "cpp",
     "__mutex_base": "cpp"
-  }
+  },
+  "dart.flutterSdkPath": "/Users/kasper/Projects/flutter/stable"
```
desistefanova, line 151: I suppose this path shouldn't be committed.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## vNext (TBD)
 
+### Breaking changes
+* The generated parts are now named `.realm.dart` instead of `.g.dart`. This is because the builder is now a `PartBuilder`, instead of a `SharedPartBuilder`. This implies that other builders can consume the output.
+
```
desistefanova, line 5: That would be a great improvement. 
The thing that is not clear for me is how the users will use the other builders. Do they have to add their attributes each time the `RealmObject` is generated? Can they put the attributes on the private class `_Person` ? Will we copy all of them to the output class `Person` during the generating of the model? How we will keep their attributes for the other builders?

nielsenko: That is separate to this PR. One way I have been thinking about is to lift annotations from the model class to the generated realm object class.

.github/workflows/ci.yml
```diff
@@ -139,7 +139,7 @@ jobs:
     secrets: inherit
     with:
       os: macos
-      runner: macos-arm
+      runner: macos-latest-xlarge # per https://github.blog/2023-10-02-introducing-the-new-apple-silicon-powered-m1-macos-larger-runner-for-github-actions/
```
nirinchev, line 142: This needs to be macos-14 to make sure we're not paying for the extra large runners.

.github/workflows/ci.yml
```diff
@@ -369,7 +369,8 @@ jobs:
       BAAS_BAASAAS_API_KEY: ${{ secrets.BAASAAS_API_KEY}}
     defaults:
       run:
-        working-directory: packages/realm/tests  
+        shell: bash
```
nirinchev, line 372: Why is this necessary? It doesn't hurt, but bash is already the default shell on linux, isn't it?

.github/workflows/ci.yml
```diff
@@ -484,9 +489,9 @@ jobs:
         with:
           channel: 'stable'
 
-      - name: Delete generated files
-        run: find . -name "*.g.dart" -delete
-        shell: bash
+      - name: Delete generated files in realm_dart
+        run: find . -name "*.realm.dart"
```
nirinchev, line 493: This doesn't delete the files, does it?

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## vNext (TBD)
 
+### Breaking changes
+* The generated parts are now named `.realm.dart` instead of `.g.dart`. This is because the builder is now a `PartBuilder`, instead of a `SharedPartBuilder`. This implies that other builders can consume the output.
```
nirinchev, line 4: Is there any special setup users need to do to take advantage of this? Might be a good idea to add a minimal example of passing the Realm generator output to another builder.

nirinchev: Also I think we need to be explicit about what people need to do to get their project up-and-running again - it may be kind of obvious, but it's usually better to have too much info than too little.

packages/realm_generator/test/common.realm.dart
```diff
@@ -0,0 +1,2 @@
+// MOCK FILE!
```
nirinchev, line 1: I think this needs more detailed code comment. I know there's a comment in the PR, but in 2 years, it will take quite a bit of detective work to uncover the reason behind these mocks.

packages/realm_generator/test/error_test_data/invalid_model_name_mapping.dart
```diff
@@ -1,10 +0,0 @@
-import 'package:realm_common/realm_common.dart';
-//part 'invalid_model_name_mapping.g.dart';
-
-const one = '1';
-
-@RealmModel()
-@MapTo(one) // <- invalid
-// prefix is not important, as we explicitly define name with @MapTo,
-// but obviously 1 is not a valid class name
```
nirinchev, line 9: Is this removed because `'1'` is now a valid class name?

nirinchev (approved):
Looks good - just small nits on my part.

nirinchev:
Also, this is a breaking change, so we should have probably targeted ni/core-vnext instead of main.

.github/workflows/ci.yml
nielsenko: This is separate commit. I did the change because our dedicated `macos-arm` runner was down, and I couldn't remember whom to bother about it. I'll roll it back, but perhaps we should consider using the free-for-open-source M1 runners on github?

.github/workflows/ci.yml
nielsenko: Yes, it is only needed for the generator tests, which also runs on windows. I'll revert here.

packages/realm_generator/test/error_test_data/invalid_model_name_mapping.dart
nielsenko: No. But this test is not actually run on main either. Apparently it broke about the time when the generator tests was refactored to use `.expected` files.

The reason is this check:
```dart
    if (_path.extension(file.path) == '.dart' && !file.path.endsWith('g.dart')) {
```
should probably have been:
```dart
    if (_path.extension(file.path) == '.dart' && !file.path.endsWith('.g.dart')) {
```
or just
```dart
    if (_path.extension(file.path) == '.dart') {
```
since there are no `.g.dart` file in those directories.

So, I'm just making it explicit that the test is not being run. It would fail, if we fix the mistake. 

The test originates back to a time when `@MapTo` on classes had a different meaning than today. Is `1` actually an illegal class name in realm-core?

packages/realm_generator/test/common.realm.dart
nielsenko: Actually, your IDE will light up like a Christmas tree, if you try to remove one. 

In a way it was odd that the shared-part builder we used before did not require the part declaration. Now that it is required, the LSP will flag the part directive if the file is missing. 

So in a sense, I would need to explain why they weren't needed before, which seems less useful to the future developer (/me).

.github/workflows/ci.yml
nirinchev: Yes, we saw that the arm runner is broken. Since there are free runners on GitHub we should use those.

packages/realm_generator/test/common.realm.dart
nirinchev: Right, my point is that for someone who's not especially proficient with generators, it may not be obvious how the tests are setup and what role these mock files play. We don't have be overly descriptive, but a short comment like "These are here to avoid the IDE reporting errors during development. The actual file will be generated during the test run and compared to the .expected file" would to a long way to clarifying things to people new to the codebase.

packages/realm_generator/test/error_test_data/invalid_model_name_mapping.dart
nirinchev: I think it should be - at least haven't seen any restrictions on numbers being used as table names.

packages/realm_generator/test/common.realm.dart
nielsenko: Would you prefer to repeat the description in each mock file, or add it as a common comment somewhere?

nirinchev: I think we can just find-replace the `// MOCK FILE!` comment with a more detailed description.

CHANGELOG.md
nielsenko: Added suggestions

PR #1316, opened 2023-06-10
bundle id sync conn

blagoev:
Fixes android build groovy scripts
Adds bundle id sync conn


nielsenko (approved):
There are a lot of extra cleanup here as well, but the stuff about stamping the pubspec.yaml name attribute into the exe looks good to me, even if it seems odd we cannot read this from some standard place.

flutter/realm_flutter/example/android/build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
     }
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:4.1.0'
+        classpath 'com.android.tools.build:gradle:7.3.0'
```
desistefanova, line 8: After increasing the version, there is a compile message in Android studio
```
The option 'android.enableR8' is deprecated.
It was removed in version 7.0 of the Android Gradle plugin.
Please remove it from `gradle.properties`.
```

blagoev: that's a warning coming from our flutter\example. I don't think it is crucial to anything

PR #1320, opened 2023-06-13
binary type support

blagoev:
Adds Binary type support

fixes https://github.com/realm/realm-dart/issues/1185


CHANGELOG.md
```diff
@@ -16,7 +16,8 @@
 ## 1.2.0 (2023-06-08)
 
 ### Enhancements
-* Added support for Full-Text search (simple term) queries. ([#1300](https://github.com/realm/realm-dart/pull/1300))
+  * Added support binary data type. ([#1320](https://github.com/realm/realm-dart/pull/1320))
+  * Added support for Full-Text search (simple term) queries. ([#1300](https://github.com/realm/realm-dart/pull/1300))
```
blagoev, line 20: this was just formatted

generator/lib/src/realm_model_info.dart
```diff
@@ -34,41 +36,51 @@ class RealmModelInfo {
   Iterable<String> toCode() sync* {
     yield 'class $name extends $modelName with RealmEntity, RealmObjectBase, ${baseType.className} {';
     {
-      final allSettable = fields.where((f) => !f.type.isRealmCollection && !f.isRealmBacklink).toList();
+      final allSettableFields = fields.where((f) => !f.type.isRealmCollection && !f.isRealmBacklink).toList();
```
nielsenko, line 39: Not a fan of this name change - repeating the type in the name

nielsenko (approved):
Already discussed on round-table yesterday. To summarize:
- I like that we use `Uint8LIst` on the interface.
- I don't off the top of my mind have a better way to handle default values for binary.
- We could consider creating a `Matcher` that knows how to compare `Uint8List`s to avoid special cases in the tests, but I don't think it important. 
- Hopefully we can improve performance once we use the new binding generator.

generator/lib/src/realm_model_info.dart
blagoev: will revert

generator/lib/src/realm_model_info.dart
```diff
@@ -16,6 +16,8 @@
 //
 ////////////////////////////////////////////////////////////////////////////////
 
+import 'dart:typed_data';
+
```
desistefanova, line 20:
```suggestion
```

desistefanova: This is marked as unused.

desistefanova (approved):
It looks good to me.

Issue #1326, opened 2023-06-21
Unable to add realm as a dependency due to http dependency conflict


botraunak:
### What happened?

While adding realm as a dependency to the project, pub complains due to conflict with the http dependency

### Repro steps

Add google fonts package as a dependency or any other package that depends on http 
Add the realm package
Error

### Version

Dart SDK version: 3.0.5, Flutter 3.10.5 • channel stable

### What Atlas Services are you using?

Local Database only

### What type of application is this?

Flutter Application

### Client OS and version

Both

### Relevant log output

```shell
flutter pub get
Resolving dependencies...
Because no versions of realm match >1.2.0 <2.0.0 and realm 1.2.0 depends on http ^0.13.4, realm ^1.2.0 requires http ^0.13.4.
And because google_fonts >=4.0.5 depends on http ^1.0.0, realm ^1.2.0 is incompatible with google_fonts >=4.0.5.
So, because realm_test depends on both google_fonts ^5.1.0 and realm ^1.2.0, version solving failed.
exit code 1
```
```

**Comment by botraunak** [comment]
fixed it locally for now by overriding the http depedency in my project
```
dependency_overrides:
  http: ^1.0.0
```

desistefanova:
Hi @botraunak, 
we really appreciate that you have reported this issue!
We will fix it. Thank you!

PR #1329, opened 2023-06-22
Bump `http` package version ^1.0.0

desistefanova:
Fixes: #1326


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 * Added support binary data type. ([#1320](https://github.com/realm/realm-dart/pull/1320))
 * Extended `ClientResetError` to return the `backupFilePath` where the backup copy of the realm will be placed once the client reset process has completed. ([#1291](https://github.com/realm/realm-dart/pull/1291))
 * Added `CompensatingWriteError` containing detailed error information about the writes that have been reverted by the server due to permissions or subscription view restrictions. The `Configuration.flexibleSync.syncErrorHandler` will be invoked with this error type when this error occurs ([#1291](https://github.com/realm/realm-dart/pull/1291)).
+* Bumped `http` package version to ^1.0.0 ([#1329](https://github.com/realm/realm-dart/pull/1329)).
```
nielsenko, line 7: Do we document package updates like this? If we do, then it should go in the internal section, no?

nielsenko: Otherwise this would qualify as minor version bump, which it definitely shouldn't

blagoev: no I don't think we need to do this

blagoev (approved):
It is ok,  after we remove the changelog entry

PR #1342, opened 2023-07-03
Support RealmSet.freeze

### LINKED ISSUE #1341: Support RealmSet.freeze
Missing freeze on RealmSet

nielsenko:
Fixes #1341

### DIFF
--- a/lib/src/set.dart
+++ b/lib/src/set.dart
@@ -99,6 +99,9 @@ abstract class RealmSet<T extends Object?> extends SetBase<T> with RealmEntity i
 
   /// Converts this [Set] to a [RealmResults].
   RealmResults<T> asResults();
+
+  /// Creates a frozen snapshot of this `RealmSet`.
+  RealmSet<T> freeze();
 }
 
 class UnmanagedRealmSet<T extends Object?> extends collection.DelegatingSet<T> with RealmEntity implements RealmSet<T> {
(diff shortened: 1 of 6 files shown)


blagoev (approved):

lib/src/set.dart
```diff
@@ -263,12 +272,24 @@ class ManagedRealmSet<T extends Object?> with RealmEntity, SetMixin<T> implement
...
   RealmResults<T> asResults() => RealmResultsInternal.create<T>(realmCore.resultsFromSet(this), realm, _metadata);
+
+  @override
+  RealmSet<T> freeze() {
+    if (isFrozen) {
+      return this;
+    }
+
+    final frozenRealm = realm.freeze();
+    return frozenRealm.resolveSet(this)!;
```
desistefanova, line 283: I think if we return `RealmError` when `resolveSet` returns `null`, we will have better traceability than if we insist that the returned value is not null with "!". I mean that it is better to have a specific error than a common error. I hope this shouldn't happen, so that you can leave it like this if you decide.

test/realm_set_test.dart
```diff
@@ -719,4 +719,52 @@ Future<void> main([List<String>? args]) async {
         ]));
     realm.write(() => testSets.objectsSet.clear());
   });
+
+  test('Set.freeze freezes the set', () {
```
desistefanova, line 723: Probably it will be good to have a test for trying to write to a frozen set. You can ignore this suggestion if you don't think it's necessary to have such test.

desistefanova (approved):
LGTM. I have only two small suggestions.

lib/src/set.dart
nielsenko: No, it should never return null in this case, since we are just resolving in the frozen version of the current realm. Same goes for list and object.

test/realm_set_test.dart
nielsenko: You would have to open a write transaction on the frozen realm. We have tests for that.

PR #1351, opened 2023-07-12
`package:http` now supports more versions: `>=0.13.0 <2.0.0`.

amondnet:
Like other packages, realm should support more versions of the http package.

https://pub.dev/packages/http/changelog
https://github.com/getsentry/sentry-dart/blob/1596141c018a1029c280be2e0a0b2cd4849239ec/dart/pubspec.yaml#L15C1-L16C1
https://github.com/grpc/grpc-dart/pull/629
https://github.com/firebase/flutterfire/blob/81156d1062d3eb3c6c83833887ca054cf66cfa13/packages/firebase_storage/firebase_storage_web/pubspec.yaml#L21

### DIFF
--- a/flutter/realm_flutter/pubspec.yaml
+++ b/flutter/realm_flutter/pubspec.yaml
@@ -33,7 +33,7 @@ dependencies:
     path: ../../generator
   tar: ^0.5.4
   build_runner: ^2.1.0
-  http: ^1.0.0
+  http: '>=0.13.0 <2.0.0'
   cancellation_token: ^2.0.0
 
 dev_dependencies:


blagoev:
Hi, Realm does support multiple versions of the http package but from 1.0 up to 2.0 (excluding).
We have 
`http: ^1.0.0` in the package pubspec.

We don't have plans to support lower than 1.0 of the http package.

Can I ask you why you can't update your dependencies and use `http: ^1.0.0` in your project?

blagoev (changes requested):
We don't have plans on supporting lower than http 1.0

amondnet:
@blagoev We should have a WIDE constraint on package:http – there were no breaks AFAIK. https://pub.dev/packages/http/changelog

PR #1354, opened 2023-07-14
Flexible sync subscribe/unsubscribe API

### LINKED ISSUE #1353: Flex Sync Subscribe API

desistefanova:
Implements `RealmResults.subscribe()` and `RealmResults.unsubscribe()` and `MutableSubscriptionSet.clear({bool unnamedOnly = false})`.
Adds cancellationToken parameter to `Session.waitForDownload()`, `Session.waitForUpload()` and `SubscriptionSet.waitForSynchronization()` 
Fixes #1353
Related to https://github.com/realm/realm-kotlin/pull/1356, https://github.com/realm/realm-js/pull/5772, https://github.com/realm/realm-swift/pull/7755, https://github.com/realm/realm-swift/pull/8244


lib/src/results.dart
```diff
@@ -135,6 +135,85 @@ extension RealmResultsOfObject<T extends RealmObjectBase> on RealmResults<T> {
...
+    Duration? timeout,
+    bool update = false,
+  }) async {
+    bool subscriptionExists = realm.subscriptions.find(this) != null;
+    realm.subscriptions.update((mutableSubscriptions) {
+      mutableSubscriptions.add(this, name: name, update: update);
+    });
+
+    if (waitForSyncMode == WaitForSyncMode.always || (waitForSyncMode == WaitForSyncMode.onCreation && !subscriptionExists)) {
+      Future<void> waitForDownload() => realm.subscriptions.waitForSynchronization().then((value) => realm.syncSession.waitForDownload());
```
nielsenko, line 180: Why not make use async keyword on `waitForDownload` and use `await` here?

lib/src/results.dart
```diff
@@ -135,6 +135,85 @@ extension RealmResultsOfObject<T extends RealmObjectBase> on RealmResults<T> {
...
+      if (timeout == null) {
```
nielsenko, line 181: Should we consider making timeout non-null with a default of fx `const Duration(seconds: 2)` .. I'm not too keen on having a api surface exposed that is guaranteed to block, if the user is offline :-/

lib/src/subscription.dart
```diff
@@ -272,21 +272,22 @@ class MutableSubscriptionSet extends SubscriptionSet {
...
   }
 
-  /// Remove the [query] from the set, if it exists.
+  /// Removes the [query] from the set, if it exists.
   bool removeByQuery<T extends RealmObject>(RealmResults<T> query) {
     return realmCore.eraseSubscriptionByResults(this, query);
   }
 
-  /// Remove the [query] from the set that matches by [name], if it exists.
+  /// Removes the subscriptions from the set that matches by [name], if it exists.
```
nielsenko, line 285:
```suggestion
  /// Remove the subscription from the set that matches by [name], if it exists.
```
You can only have one with a given name

test/subscription_test.dart
```diff
@@ -595,4 +595,145 @@ Future<void> main([List<String>? args]) async {
...
+    realm.close();
+
+    final user1 = await app.logIn(Credentials.anonymous(reuseCredentials: false));
+    final config1 = Configuration.flexibleSync(user1, [Product.schema]);
+    final realm1 = getRealm(config1);
+    final query = realm1.query<Product>('$byTestRun AND name ENDSWITH \$0', ["1"]);
+    final results = await query.subscribe(waitForSyncMode: WaitForSyncMode.never);
+    expect(results.length, 0); // didn't wait for downloading because of WaitForSyncMode.never
+
+    final first = await query.subscribe(waitForSyncMode: WaitForSyncMode.always, timeout: Duration(milliseconds: 1));
```
nielsenko, line 733: Feels brittle to assume that a short wait time will not allow the download to succeed.

lib/src/results.dart
```diff
@@ -135,6 +135,85 @@ extension RealmResultsOfObject<T extends RealmObjectBase> on RealmResults<T> {
...
+        await waitForDownload();
+      } else {
+        await waitForDownload().timeout(timeout, onTimeout: () {});
```
nielsenko, line 184: I think we should mention in the doc comment that a timeout is silent, ie. won't raise an exception. That is not common behaviour in Dart.

nielsenko (changes requested):

lib/src/results.dart
desistefanova: Yes, perhaps it will be better. @blagoev what do you think?

lib/src/results.dart
```diff
@@ -135,6 +135,85 @@ extension RealmResultsOfObject<T extends RealmObjectBase> on RealmResults<T> {
...
+  ///
+  /// The [timeout] is the maximum duration to wait for objects to be downloaded.
+  /// If the time exceeds this limit, the [RealmResults] is returned and the download
+  /// continues in the background.
+  ///
+  /// {@category Sync}
+  Future<RealmResults<T>> subscribe({
+    String? name,
+    WaitForSyncMode waitForSyncMode = WaitForSyncMode.onCreation,
+    Duration? timeout,
```
nielsenko, line 171: I would prefer to take a `CancellationToken`.. people can pass one with a timeout.. and maybe even default to using one with a timeout, if none is passed, but we should discuss that, as it is not inline with the other SDKs as I understand.

desistefanova: Done

lib/src/results.dart
desistefanova: CancellationToken added instead of timeout.

lib/src/native/realm_core.dart
```diff
@@ -2329,8 +2339,8 @@ class _RealmCore {
   }
 
   static void _sessionWaitCompletionCallback(Object userdata, Pointer<realm_sync_error_code_t> errorCode) {
-    final completer = userdata as Completer<void>;
-    if (completer.isCompleted) {
+    final completer = userdata as CancellableCompleter<void>;
+    if (completer.isCompleted || completer.isCancelled) {
       return;
     }
```
nielsenko, line 2345:
```suggestion
    if (completer.isCancelled) {
      return;
    }
```

lib/src/subscription.dart
```diff
@@ -272,21 +273,22 @@ class MutableSubscriptionSet extends SubscriptionSet {
     return Subscription._(realmCore.insertOrAssignSubscription(this, query, name, update));
   }
 
-  /// Remove the [subscription] from the set, if it exists.
+  /// Removes the [subscription] from the set, if it exists.
```
nielsenko, line 276: Why change the tense? If we do, there are doc comments on other methods like `clear` that should change as well..

lib/src/results.dart
```diff
@@ -135,6 +137,99 @@ extension RealmResultsOfObject<T extends RealmObjectBase> on RealmResults<T> {
...
+
+  _SubscribedRealmResult._(RealmResults<RealmObject> results, {this.subscriptionName})
+      : super._(
+          results.handle,
+          results.realm,
+          results.metadata,
+        );
+}
+
+extension RealmResultsOfRealmObject<T extends RealmObject> on RealmResults<T> {
```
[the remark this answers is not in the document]
desistefanova: Because we allow making queries with embedded objects to a local realm. But it is not allowed to make subscriptions on embedded objects for sync realm.

Issue #1357, opened 2023-07-18
Upgrade to newer flutter analyzer


brookman:
### Problem

The current version realm_generator 1.3.0 depends on an older version of the Flutter analyzer package.
This version contains regressions which are fixed in analyzer 6.0.0. The new version makes ours code gen step 400% faster again.

### Solution

Upgrade the dependency.

### How important is this improvement for you?

Would be a major improvement

### Feature would mainly be used with

Local Database only

desistefanova:
Hi @brookman!
Thanks for the reporting! We will definitely do that in some of the next releases. 
I hope it is not a blocker for you.

nielsenko:
@brookman Are you using flutter master channel?

nielsenko:
The following minimal `pubspec.yaml` illustrates the problem with your request:
```yaml
name: upgrade_analyzer
environment:
  sdk: '>=3.0.6 <4.0.0'

dependencies:
  analyzer: ^6.0.0
  flutter:
    sdk: flutter

dev_dependencies:
  flutter_test:
    sdk: flutter
  test: any # no version of test will work with analyzer ^6.0.0 and current flutter stable (3.10.6)
```
You will not be able resolve this with current stable version of flutter. I think it is a bit unacceptable for a package like realm to require people to use the `master` channel or stop using the `test` package 😄 

That said, we will revisit on each new stable release of flutter.

PR #1365, opened 2023-08-01
kn/update analyzer

nielsenko:
- Update flutter and analyzer constraints to support analyzer ^6.0.0

Fix: #1357


desistefanova:
I see, that the sdk version is increased here in order to force using the new analyzer, but I think we should wait until there is a solution on the [Linux issue](https://github.com/dart-lang/sdk/issues/53267) with the new flutter.

nielsenko:
> I see, that the sdk version is increased here in order to force using the new analyzer, but I think we should wait until there is a solution on the [Linux issue](https://github.com/dart-lang/sdk/issues/53267) with the new flutter.

I tend to disagree. Very few people a using realm on linux, even fewer is using it outside flutter.

nielsenko:
@nirinchev The linux issue holding this back has been fixed.

nielsenko:
@nirinchev Sorry, you need to revert 6abe028d which is a temporary fix I did. Note that this PR will force users of the next version to use Flutter 3.13.0 or later (which is latest stable) to get Dart 3.1.0+

sagnik-sanyal:
Hi there, when can we expect this PR to be merged ? We are waiting for this issue to be resolved

nirinchev:
@sagnik-sanyal I've taken over this PR and expect to merge it soon. The reason why it's delayed is because I want to check how many users will be impacted by the flutter upgrade.

nirinchev:
Based on our metrics, around 13% of our users are running on dart 3.1.0+. I'm a little hesitant to force the upgrade at this point unless you folks feel very confident that the flutter community is comfortable with always running latest and greatest.

nielsenko:
👍

PR #1383, opened 2023-08-18
Collections in mixed

nielsenko:
Allow `RealmValue`s (mixed) to contain collections, and support a convenient syntax for constructing `RealmValue`s of nested collections, and drilling into these to extract values.

Here is the suggested interface:
```dart
  final any = RealmValue.from([
    1,
    2,
    {
      'x': {1, 2}
    },
  ]);

  // access list element at index 2,
  // then map element with key 'x',
  // then set element 1, if it exists
  // assuming an int, or null if not found
  final x = any[2]['x'][1].as<int?>();
  assert(x == 1);

  // or, a bit shorter
  final y = any[2]['x'][1]<int?>();
  assert(y == 1);

  // or, shorter yet
  int? z = any[2]['x'][1]();
  assert(z == 1);

  // or, if you are sure
  int w = any[2]['x'][1](); // <-- shortest
  assert(w == 1);

  // or, using a list of indexes
  int u = any[[2, 'x', 1]]();
  assert(u == 1);

  // which allows for a different layout
  int v = any[[
    2,
    'x',
    1,
  ]]();
  assert(v == 1);
```

The code-behind is not done yet

### DIFF
--- a/common/lib/src/realm_types.dart
+++ b/common/lib/src/realm_types.dart
@@ -16,7 +16,6 @@
 //
 ////////////////////////////////////////////////////////////////////////////////
 
-import 'dart:ffi';
 import 'dart:typed_data';
 import 'package:objectid/objectid.dart';
 import 'package:sane_uuid/uuid.dart';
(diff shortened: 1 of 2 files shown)


common/lib/src/realm_types.dart
```diff
@@ -227,3 +292,47 @@ class RealmValue {
   @override
   String toString() => 'RealmValue($value)';
 }
+
+void demo() {
```
nielsenko, line 296: This is a very early draft. Obviously this demo method should not survive

PR #1387, opened 2023-08-25
Core 13.20.1 sync errors changes

### LINKED ISSUE #1252: Sync Client Error Simplification
Related to core updates:

* Sync client error migration to Status from std::error_code - [https://github.com/realm/realm-core/commit/d89a3309766d8280df7395644e3cfe49bd9ecea1](https://github.com/realm/realm-core/commit/d89a3309766d8280df7395644e3cfe49bd9ecea1)

desistefanova:
Fixes #1252 
This PR proposes sync errors simplification. The reason is to achieve a reasonable mapping to the new sync errors structure in the C-API introduced by https://github.com/realm/realm-core/commit/d89a3309766d8280df7395644e3cfe49bd9ecea1

Deprecated sync errors:

- `SyncErrorCategory`
- `SyncClientErrorCode`, `SyncConnectionErrorCode`, `SyncSessionErrorCode`, `SyncWebSocketErrorCode`, `GeneralSyncErrorCode`
- `SyncClientError`, `SyncConnectionError`, `SyncSessionError`, `SyncWebSocketError`, `GeneralSyncError`.

Deprecated `App.localAppName` and `App.localAppVersion`, since they are not used by realm core.


lib/src/native/realm_core.dart
```diff
@@ -3432,3 +3421,28 @@ extension on realm_error {
     return LastError(error, message, userError);
   }
 }
+
+/// @nodoc
+class SyncErrorCodesConstants {
```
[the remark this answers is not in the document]
desistefanova: Yes, it could be. I can change it.

nielsenko (approved):
LGTM

lib/src/native/realm_core.dart
```diff
@@ -3399,25 +3383,30 @@ extension PlatformEx on Platform {
...
+    this.path,
     this.isFatal = false,
     this.isClientResetRequested = false,
     this.originalFilePath,
     this.backupFilePath,
     this.compensatingWrites,
   });
+
+  // @categoryConstant param should be a value from `realm_error_category` class
+  bool _isFromCategory(int categoryConstant) {
```
blagoev, line 3407: we usually move the values in realm_core in Dart and use this one. 
This is what we have done with `SyncErrorCodes` enum below. Lets do the same for the categories and make a private enum with the values so we can safelly use it here as parameter type. then we can remove the error prone int value and the comment on this method

lib/src/native/realm_core.dart
```diff
@@ -3432,3 +3421,38 @@ extension on realm_error {
...
+  syncProtocolInvariantFailed(realm_errno.RLM_ERR_SYNC_PROTOCOL_INVARIANT_FAILED),
+  syncProtocolNegotiationFailed(realm_errno.RLM_ERR_SYNC_PROTOCOL_NEGOTIATION_FAILED),
+  syncServerPermissionsChanged(realm_errno.RLM_ERR_SYNC_SERVER_PERMISSIONS_CHANGED),
+  syncUserMismatch(realm_errno.RLM_ERR_SYNC_USER_MISMATCH),
+  tlsHandshakeFailed(realm_errno.RLM_ERR_TLS_HANDSHAKE_FAILED),
+  wrongSyncType(realm_errno.RLM_ERR_WRONG_SYNC_TYPE),
+  syncWriteNotAllowed(realm_errno.RLM_ERR_SYNC_WRITE_NOT_ALLOWED),
+  authError(realm_errno.RLM_ERR_AUTH_ERROR),
+  unknownError(realm_errno.RLM_ERR_UNKNOWN),
+  notSyncError(-1);
```
blagoev, line 3448: Why do we introduce our own value here when we have `unknownError`. This results in invalid value later in `fromInt` method. Why don't we default to `unknownError`?

lib/src/session.dart
```diff
@@ -246,6 +246,7 @@ enum ProgressMode {
 }
 
 /// The category of a [SyncError].
+@Deprecated("The new errors are not classified by SyncErrorCategory anymore.")
```
blagoev, line 249:
```suggestion
@Deprecated("Sync errors are not classified by SyncErrorCategory anymore.")
```

lib/src/session.dart
```diff
@@ -271,6 +272,7 @@ enum SyncErrorCategory {
 /// These errors will terminate the network connection
 /// (disconnect all sessions  associated with the affected connection),
 /// and the error will be reported via the connection state change listeners of the affected sessions.
+@Deprecated("Handle SyncError or its subclasses instead of checking the error codes.")
```
blagoev, line 275:
```suggestion
@Deprecated("Use SyncError or its subclasses instead.")
```

This repeats on many places. I suggest we use this generic deprecation text in all places

lib/src/native/realm_core.dart
blagoev: This also seems unused?

lib/src/native/realm_core.dart
```diff
@@ -3399,25 +3383,30 @@ extension PlatformEx on Platform {
 /// @nodoc
 class SyncErrorDetails {
   final String message;
-  final SyncErrorCategory category;
+  final int _categoryFlag;
```
blagoev, line 3386: This seems not used

lib/src/native/realm_core.dart
```diff
@@ -3399,25 +3383,30 @@ extension PlatformEx on Platform {
...
+  final String? path;
   final bool isFatal;
   final bool isClientResetRequested;
   final String? originalFilePath;
   final String? backupFilePath;
   final List<CompensatingWriteInfo>? compensatingWrites;
-  const SyncErrorDetails(
+  SyncErrorDetails(
     this.message,
-    this.category,
```
blagoev, line 3412: This is removed, but errors still have categories. And one error can have multiple categories. Kotlin is including this in their errors as detail message property.

blagoev (changes requested):

lib/src/native/realm_core.dart
desistefanova: Discussed and decided to be removed.

lib/src/native/realm_core.dart
desistefanova: Removed `notSyncError`

lib/src/configuration.dart
```diff
@@ -614,32 +614,26 @@ class ClientResetError extends SyncError {
...
 
   /// The [SyncSessionErrorCode] value indicating the type of the sync error.
   /// This property will be [SyncSessionErrorCode.unknown] if `onManualResetFallback` occurs on client reset.
-  SyncSessionErrorCode get sessionErrorCode => SyncSessionErrorCode.fromInt(codeValue);
+  @Deprecated("SessionErrorCode property value is always `unknown`. It will be deleted.")
+  SyncSessionErrorCode get sessionErrorCode => SyncSessionErrorCode.unknown;
 
-  @Deprecated("ClientResetError constructor is deprecated and will be removed in the future")
-  ClientResetError(
+  ClientResetError._(
```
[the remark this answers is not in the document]
desistefanova: I will return back the old constructor and will add a new private one.

lib/src/native/realm_core.dart
```diff
@@ -685,8 +685,8 @@ class _RealmCore {
 
       if (error != nullptr) {
         final err = arena<realm_error>();
-        _realmLib.realm_get_async_error(error, err);
-        completer.completeError(RealmException("Failed to open realm ${err.ref.toLastError().toString()}"));
+        bool asyncErrorAvailable = _realmLib.realm_get_async_error(error, err);
```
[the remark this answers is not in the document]
desistefanova: `asyncErrorAvailable` renamed to `success`.

lib/src/configuration.dart
```diff
@@ -614,32 +614,26 @@ class ClientResetError extends SyncError {
...
   /// when a client reset fails and `onManualResetFallback` occurs. Otherwise, it is [SyncClientErrorCode.unknown]
-  SyncClientErrorCode get code => SyncClientErrorCode.fromInt(codeValue);
+  @Deprecated("Code property value is always `autoClientResetFailure`. It will be deleted.")
+  SyncClientErrorCode get code => SyncClientErrorCode.autoClientResetFailure;
 
   /// The [SyncSessionErrorCode] value indicating the type of the sync error.
   /// This property will be [SyncSessionErrorCode.unknown] if `onManualResetFallback` occurs on client reset.
-  SyncSessionErrorCode get sessionErrorCode => SyncSessionErrorCode.fromInt(codeValue);
+  @Deprecated("SessionErrorCode property value is always `unknown`. It will be deleted.")
+  SyncSessionErrorCode get sessionErrorCode => SyncSessionErrorCode.unknown;
```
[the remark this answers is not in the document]
desistefanova: We deprecate these codes because `SyncClientErrorCode` and `SyncSessionErrorCode` are also deprecated.

lib/src/configuration.dart
```diff
@@ -614,32 +614,26 @@ class ClientResetError extends SyncError {
 
   /// The [ClientResetError] has error code of [SyncClientErrorCode.autoClientResetFailure]
   /// when a client reset fails and `onManualResetFallback` occurs. Otherwise, it is [SyncClientErrorCode.unknown]
-  SyncClientErrorCode get code => SyncClientErrorCode.fromInt(codeValue);
+  @Deprecated("Code property value is always `autoClientResetFailure`. It will be deleted.")
+  SyncClientErrorCode get code => SyncClientErrorCode.autoClientResetFailure;
```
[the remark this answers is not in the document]
desistefanova: We deprecate these codes because `SyncClientErrorCode` and `SyncSessionErrorCode` are also deprecated.

lib/src/configuration.dart
```diff
@@ -603,6 +602,7 @@ class ClientResetError extends SyncError {
   final App? _app;
 
   /// If true the received error is fatal.
+  @Deprecated("ClientResetError is always a fatal error.")
```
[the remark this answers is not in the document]
desistefanova: It is useless. Discussed and decided to be removed.

lib/src/native/realm_core.dart
desistefanova: `_categoryFlag` is removed.

lib/src/native/realm_core.dart
desistefanova: `category` argument is removed.

lib/src/session.dart
```diff
@@ -271,6 +272,7 @@ enum SyncErrorCategory {
 /// These errors will terminate the network connection
 /// (disconnect all sessions  associated with the affected connection),
 /// and the error will be reported via the connection state change listeners of the affected sessions.
+@Deprecated("Use SyncError or its subclasses instead.")
```
blagoev, line 275: This message is fine but the rest omit the `Use` word? This seems like copy/paste error

blagoev (changes requested):
All ready, the deprecation message seems copy/paste error. Should we fix it?

lib/src/session.dart
desistefanova: Thanks for noticing that!

test/session_test.dart
```diff
@@ -279,39 +279,41 @@ Future<void> main([List<String>? args]) async {
...
+  baasTest('WrongSyncTypeError test error handler', (configuration) async {
+    final app = App(configuration);
+    final user = await getIntegrationUser(app);
+    final config = Configuration.flexibleSync(user, syncSchema, syncErrorHandler: (syncError) {
+      expect(syncError, isA<WrongSyncTypeError>());
+      expect(syncError.message, isNotEmpty);
+    });
+    final realm = getRealm(config);
+
+    realm.syncSession.raiseError(SyncErrorCodes.wrongSyncType.code, true);
```
blagoev, line 316: this does not test if the callback was invoked. Should we have an `expect` that checks the callback was invoked otherwise it will be green when in fact the expects in the callback neve happened

blagoev (changes requested):
Its ready but the new test should check if the callback was invoked.

blagoev (approved):

test/session_test.dart
blagoev: ignore. it is our raiseError method so we don't need to verify the callback is invoked.

PR #1389, opened 2023-08-28
Rudimentary geospatial support

nielsenko:
Add rudimentary geospatial support. Example:
```dart
@RealmModel(ObjectType.embeddedObject)
class _Location {
  final String type = 'Point';
  List<double> coordinates = const [0, 0];
}

@RealmModel()
class _Restaurant {
  @PrimaryKey()
  late String name;
  _Location? location;
}

realm.query<Restaurant>('location geoWithin $shape');
```
Supported shapes are `GeoBox`, `GeoCircle`, and `GeoPolygon`

```dart
realm.query<Restaurant>('location geoWithin \$0', [shape]);
```
is also supported, thanks to https://github.com/realm/realm-core/pull/6934, despite C-api not supporting geo shapes. 

It works by allowing string arguments where geo shapes are required and trying to parse such from any given string, so is still implicitly dependent on inlining and hence not as efficient, but it will not force users to inline on the dart side, and hence they will automatically benefit once the C-api is updated, or we have switched to the new binding generator.

**NOTE**: This PR depends on a core-upgrade newer than v13.20.0, and hence a refactoring of the way errors are handled. It is currently  based on top op #1398.

Fixes: #1155


common/lib/src/realm_types.dart
```diff
@@ -224,3 +225,223 @@ class RealmValue {
...
+///
+/// It cannot be persisted as a property on a realm object.
+///
+/// Instead, you must use a custom embedded object with the following structure:
+/// ```dart
+/// @RealmModel(ObjectType.embeddedObject)
+/// class _Location {
+///   final String type = 'Point';
+///   List<double> coordinates = const [0, 0];
+/// }
```
nirinchev, line 242: Does the dart generator support playing around with the accessibility of fields? If so, we could suggest a more user-friendly API, similar to what we do in .NET: https://github.com/realm/realm-dotnet/blob/641f2dc5581776b80979c69abc4c29777313bd0b/CHANGELOG.md?plain=1#L140-L162

common/lib/src/realm_types.dart
```diff
@@ -224,3 +225,223 @@ class RealmValue {
...
+  GeoPoint(this.lat, this.lon) {
+    if (lat < -90 || lat > 90) throw ArgumentError.value(lat, 'lat', 'must be between -90 and 90');
+    if (lon < -180 || lon > 180) throw ArgumentError.value(lon, 'lng', 'must be between -180 and 180');
+  }
+
+  @override
+  bool operator ==(Object other) {
+    if (identical(this, other)) return true;
+    if (other is! GeoPoint) return false;
+    return lat == other.lat && lon == other.lon;
```
nirinchev, line 269: This is a matter of preference, but can this not be rewritten like:

```suggestion
    return other is GeoPoint && lat == other.lat && lon == other.lon;
```

I think it's more readable than the alternating true/false/expression pattern in this and the other `==` overrides.

common/lib/src/realm_types.dart
```diff
@@ -224,3 +225,223 @@ class RealmValue {
...
+  }
+
+  @override
+  int get hashCode => Object.hash(lat, lon);
+
+  @override
+  String toString() => '[$lon, $lat]';
```
nirinchev, line 276: I think it's confusing that the constructor orders arguments like `[lat, lon]` while toString prints them `[lon, lat]` - we should pick one.

common/lib/src/realm_types.dart
```diff
@@ -224,3 +225,223 @@ class RealmValue {
...
+
+  @override
+  String toString() => 'geoBox($southWest, $northEast)';
+}
+
+typedef GeoRing = List<GeoPoint>;
+
+extension on GeoRing {
+  void validate() {
+    if (first != last) throw ArgumentError('Vertices must for a ring (first != last)');
```
nirinchev, line 308: The error string seems to be missing some words.

test/geospatial_test.dart
```diff
@@ -0,0 +1,335 @@
...
+
+  double get longitude => coordinates[0];
+  set longitude(double value) => coordinates[0] = value;
+
+  double get latitude => coordinates[1];
+  set latitude(double value) => coordinates[1] = value;
+}
+
+@RealmModel()
+class _Restaurant {
```
nirinchev, line 43: I realized we don't have any tests for collections of geo points in .NET. Not sure if this is covered by other SDK tests, but might be a good idea to add some for dart at least.

nirinchev (approved):
Looks reasonable - I only have nitty comments.

common/lib/src/realm_types.dart
nielsenko: Then `toString` wins, as this is the order needed by realm-core

common/lib/src/realm_types.dart
nielsenko: I have added getters and setters for latitude and longitude, as well conversion methods between `GeoPoint` and `Location` in the doc comments (similar to how I did in `geospatial_test.dart`)

common/lib/src/realm_types.dart
nielsenko: It is common idiom. The two first lines are identical for all types, except of course for the type tested against. 
You really only have to _think_ about the lines after the first two.

I think I prefer as is, even though yours is one line shorter.

common/lib/src/realm_types.dart
nielsenko: I went with `[lon, lat]` and named required parameters.

common/lib/src/realm_types.dart
nielsenko: Fixed

test/geospatial_test.dart
nielsenko: Isn't that just a test that we can store embedded objects in a list?

nirinchev: Well, not just store - also run geo queries. While I do hope that's covered by Core tests, it never hurts to double check.

nielsenko: Done

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### Enhancements
 * Support efficient `skip` on `RealmResults` ([#1391](https://github.com/realm/realm-dart/pull/1391))
 * Support efficient `indexOf` and `contains` on `RealmResults` ([#1394](https://github.com/realm/realm-dart/pull/1394))
+* Support for performing geo spatial queries using the new classes: `GeoPoint`, `GeoCircle`, `GeoBox` and `GeoPolygon`. See `GeoPoint` documentation on how to persist locations ([#1389](https://github.com/realm/realm-dart/pull/1389))
```
desistefanova, line 6: Just a reminder to move this line up after merging from https://github.com/realm/realm-dart/pull/1387

lib/src/native/realm_core.dart
```diff
@@ -2995,7 +2995,15 @@ void _intoRealmQueryArg(Object? value, Pointer<realm_query_arg_t> realm_query_ar
     realm_query_arg.ref.arg = allocator<realm_value_t>();
     realm_query_arg.ref.nb_args = 1;
     realm_query_arg.ref.is_list = false;
-    _intoRealmValue(value, realm_query_arg.ref.arg.ref, allocator);
+    _intoRealmValueHack(value, realm_query_arg.ref.arg.ref, allocator);
+  }
+}
+
+void _intoRealmValueHack(Object? value, realm_value realm_value, Allocator allocator) {
```
desistefanova, line 3002: I'm not sure if this is a temporary solution looking at the method name ending with `Hack`. If not, maybe we need better name here.

lib/src/native/realm_core.dart
```diff
@@ -2995,7 +2995,15 @@ void _intoRealmQueryArg(Object? value, Pointer<realm_query_arg_t> realm_query_ar
...
+  if (value is GeoShape) {
+    _intoRealmValue(value.toString(), realm_value, allocator);
```
desistefanova, line 3004: Relying on `value.toString()` may cause some potential bugs. Imagine that the users extend the geo classes and override the `toString()`. 
I'm wondering 
- either we can add `toQuery` abstract method to `GeoShaps` that to be dedicated for the queries
- or to encapsulate this logic inside realm_core and implement a converter to query arguments, so we can be sure that the users won't override it.
I would rather prefer the second option, even though it is not aligned to OOP.

lib/src/native/realm_core.dart
nielsenko: This is a temporary solution until the c-api is extended to support geospatial types.

lib/src/native/realm_core.dart
nielsenko: GeoShape is sealed. Users cannot extend these classes.

common/lib/src/realm_types.dart
```diff
@@ -227,3 +228,248 @@ class RealmValue {
   @override
   String toString() => 'RealmValue($value)';
 }
+
+/// A base type for the supported geospatial shapes.
+sealed class GeoShape {}
```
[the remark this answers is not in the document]
nielsenko: You are right. Good spot.

common/lib/src/realm_types.dart
```diff
@@ -227,3 +228,248 @@ class RealmValue {
...
+///
+///   // The rest of the class is just convenience methods
+///   double get lon => coordinates[0];
+///   set lon(double value) => coordinates[0] = value;
+///
+///   double get lat => coordinates[1];
+///   set lat(double value) => coordinates[1] = value;
+///
+///   GeoPoint toGeoPoint() => GeoPoint(lon: lon, lat: lat);
+/// }
```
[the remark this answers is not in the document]
nielsenko: I don't think we should include realm models in the SDK code itself. 

There is a phase 2 to this project, which will introduce persist-able classes. I'm following the other SDKs here and not limiting our freedom too much.

The idea about generating it in the generator and map to GeoPoint I like, but it may conflict with the realm-core design of phase 2.

lib/src/native/realm_core.dart
desistefanova: Discussed and decided no to allow users to extend these classes. All the classes that implement GeoShape should be marked as `final`.

desistefanova (approved):

nielsenko:
Thx @nirinchev ..  just need to rerun generator 👍

nirinchev:
Oh, yeah, needs to add the ignore lint comments that went in a bit earlier. Will do that.

PR #1391, opened 2023-09-01
kn/results skip performance

nielsenko:
This makes `skip` a simple integer addition, and defer handle construction as long as possible.

- Fetch handle in _RealmResultsIterator.current lazily (it may never be called for a given index)
- Support efficient skip on RealmResults

This came about due to https://www.mongodb.com/community/forums/t/do-relationship-properties-get-populated-during-tolist/242260.

`skip` is useful if you wan't to do pagination for various reasons, even though realm doesn't require it. In this case to support an existing data access layer previously build on another DB.

Fixes: #1392

### DIFF
--- a/lib/src/results.dart
+++ b/lib/src/results.dart
@@ -216,18 +216,16 @@ class _RealmResultsIterator<T extends Object?> implements Iterator<T> {
         _index = -1;
 
   @override
-  T get current => _current as T;
+  T get current => _current ??= _results[_index];
 
   @override
   bool moveNext() {
     int length = _results.length;
+    _current = null;
     _index++;
     if (_index >= length) {
-      _current = null;
       return false;
     }
-    _current = _results[_index];
-
     return true;
   }
 }
--- a/lib/src/results.dart
+++ b/lib/src/results.dart
@@ -219,9 +219,9 @@ class _RealmResultsIterator<T extends Object?> implements Iterator<T> {
   int _index;
   T? _current;
 
-  _RealmResultsIterator(RealmResults<T> results, {int index = 0})
+  _RealmResultsIterator(RealmResults<T> results)
       : _results = results,
-        _index = index - 1;
+        _index = -1;
 
   @override
   T get current => _current ??= _results[_index];
--- a/lib/src/results.dart
+++ b/lib/src/results.dart
@@ -30,11 +30,13 @@ import 'realm_object.dart';
 class RealmResults<T extends Object?> extends Iterable<T> with RealmEntity implements Finalizable {
   final RealmObjectMetadata? _metadata;
   final RealmResultsHandle _handle;
+  final int _skipOffset; // to support skip efficiently
 
   final _supportsSnapshot = <T>[] is List<RealmObjectBase?>;
 
-  RealmResults._(this._handle, Realm realm, this._metadata) {
+  RealmResults._(this._handle, Realm realm, this._metadata, [this._skipOffset = 0]) {
     setRealm(realm);
+    assert(length >= 0);
   }
 
   /// Gets a value indicating whether this collection is still valid to use.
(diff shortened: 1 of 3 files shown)


lib/src/results.dart
```diff
@@ -211,23 +219,21 @@ class _RealmResultsIterator<T extends Object?> implements Iterator<T> {
...
 
-  _RealmResultsIterator(RealmResults<T> results)
+  _RealmResultsIterator(RealmResults<T> results, {int index = 0})
       : _results = results,
-        _index = -1;
+        _index = index - 1;
 
   @override
-  T get current => _current as T;
+  T get current => _current ??= _results[_index];
```
nirinchev, line 227: This will be a problem when we reach the end of the iterator. Previously it would return null, now it will throw.

nielsenko: In Dart iteration is controlled by the **bool** returned by `moveNext`. You will never be calling `current` directly. 

If you did you could also make the current implementation throw by calling `current` before `moveNext`.

Notice that no tests where harmed in this refactoring.

nielsenko: Also we do actually throw today when `T` is non-nullable.

nirinchev: I see, if throwing is consistent with the dart API, then it looks good.

nielsenko: Here is a test that illustrates the behaviour:
```dart
import 'package:test/test.dart';

void main(List<String> arguments) {
  test('iterable', () {
    final list = [1];
    final it = list.iterator;
    // you are not supposed to call current before first moveNext
    expect(() => it.current, throwsA(isA<TypeError>()));
    expect(it.moveNext(), isTrue);
    expect(it.moveNext(), isFalse);
    // you are not supposed to call current, if moveNext return false
    expect(() => it.current, throwsA(isA<TypeError>()));
  });
}
```

nielsenko: I have added a test for these edge-cases for `_RealmResultsIterator` as well

nirinchev (commented):
Should we also implement efficient skip for list as well?

nielsenko:
> Should we also implement efficient skip for list as well?

Given that `ManagedRealmList` implements `List` using a `ListMixin` (which now a days is equal to `ListBase`) it actually already uses an efficient implementation for `skip`.

```dart
  Iterable<E> skip(int count) => SubListIterable<E>(this, count, null);
```

`SubListIterable` does the right arithmetics for us.

lib/src/results.dart
```diff
@@ -211,23 +219,21 @@ class _RealmResultsIterator<T extends Object?> implements Iterator<T> {
   int _index;
   T? _current;
 
-  _RealmResultsIterator(RealmResults<T> results)
+  _RealmResultsIterator(RealmResults<T> results, {int index = 0})
```
desistefanova, line 222: The is not used anywhere with the new parameter `index`. I suppose the return value of the `iterator` getter should looks like:
```dart
  @override
  Iterator<T> get iterator {
    ......
    return _RealmResultsIterator(results, _skipOffset);
```

desistefanova: Hm, actually you already add `_skipOffset` when you call `realmCore.resultsGetObjectAt`. So, I'm not sure if we need this new `index` parameter of `_RealmResultsIterator` constructor.

test/results_test.dart
```diff
@@ -18,8 +18,10 @@
 
 // ignore_for_file: unused_local_variable
 
+import 'dart:math';
 import 'dart:typed_data';
 
+import 'package:ffi/ffi.dart';
```
desistefanova, line 24:
```suggestion
```

lib/src/results.dart
nielsenko: You are right, this is remnants from before I realised `skip` returns an iterable.

desistefanova (approved):

PR #1400, opened 2023-09-14
kn/asymmetric sync

nielsenko:
Support asymmetric objects
```dart
@RealmModel(ObjectType.asymmetricObject)
class _Asymmetric {
  @PrimaryKey()
  @MapTo('_id')
  late ObjectId id;

  late List<_Embedded> embeddedObjects;
}
// ...    
realm.write(() {
  realm.ingest(Asymmetric(oid));
});
```

Fixes: #917

### DIFF
--- a/generator/test/error_test_data/asymmetric_external_link.expected
+++ b/generator/test/error_test_data/asymmetric_external_link.expected
@@ -1,0 +1,12 @@
+Linking to asymmetric objects is not allowed
+
+in: asset:pkg/test/error_test_data/asymmetric_external_link.dart:12:16
+   ╷
+10 │ @RealmModel()
+11 │ class _Bad {
+   │       ━━━━ in realm model for 'Bad'
+12 │   _Asymmetric? asymmetric;
+   │                ^^^^^^^^^^ !
+   ╵
+Remove the field
+
--- a/generator/test/good_test_data/asymmetric_object.expected
+++ b/generator/test/good_test_data/asymmetric_object.expected
@@ -6,12 +6,12 @@ class Asymmetric extends _Asymmetric
     with RealmEntity, RealmObjectBase, RealmObject {
   Asymmetric(
     ObjectId id, {
-    Embedded? farther,
+    Embedded? father,
     Embedded? mother,
     Iterable<Embedded> children = const [],
   }) {
     RealmObjectBase.set(this, '_id', id);
-    RealmObjectBase.set(this, 'farther', farther);
+    RealmObjectBase.set(this, 'father', father);
     RealmObjectBase.set(this, 'mother', mother);
     RealmObjectBase.set<RealmList<Embedded>>(
         this, 'children', RealmList<Embedded>(children));
@@ -32,11 +32,11 @@ class Asymmetric extends _Asymmetric
       throw RealmUnsupportedSetError();
 
   @override
-  Embedded? get farther =>
-      RealmObjectBase.get<Embedded>(this, 'farther') as Embedded?;
+  Embedded? get father =>
+      RealmObjectBase.get<Embedded>(this, 'father') as Embedded?;
   @override
-  set farther(covariant Embedded? value) =>
-      RealmObjectBase.set(this, 'farther', value);
+  set father(covariant Embedded? value) =>
+      RealmObjectBase.set(this, 'father', value);
 
   @override
   Embedded? get mother =>
@@ -62,7 +62,7 @@ class Asymmetric extends _Asymmetric
           mapTo: '_id', primaryKey: true),
       SchemaProperty('children', RealmPropertyType.object,
           linkTarget: 'Embedded', collectionType: RealmCollectionType.list),
-      SchemaProperty('farther', RealmPropertyType.object,
+      SchemaProperty('father', RealmPropertyType.object,
           optional: true, linkTarget: 'Embedded'),
       SchemaProperty('mother', RealmPropertyType.object,
           optional: true, linkTarget: 'Embedded'),
--- a/generator/test/good_test_data/asymmetric_object.expected
+++ b/generator/test/good_test_data/asymmetric_object.expected
@@ -1,0 +1,112 @@
+// **************************************************************************
+// RealmObjectGenerator
+// **************************************************************************
+
+class Asymmetric extends _Asymmetric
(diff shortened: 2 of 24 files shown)


nielsenko:
Note that this PR is dependent on dart 3.1.0 which currently breaks linux due to: https://github.com/dart-lang/sdk/issues/53267. Fix is planned to be cherry-picked: https://github.com/dart-lang/sdk/issues/53503.

lib/src/realm_class.dart
```diff
@@ -278,6 +279,18 @@ class Realm implements Finalizable {
     return object;
   }
 
+  /// Ingest an [AsymmetricObject] to the [Realm].
+  ///
+  /// Ingesting is a write only operation. The ingested object will be dead immediately
```
blagoev, line 284: This seems a bit unrefined. 
This is what the docs says
>You cannot read these objects. Once created, they sync to the App Services backend and the linked Atlas database. After syncing to Atlas, Realm deletes asymmetric objects from the device. This keeps the realm file size small even with a heavy insert workload.

Could this be
>Ingesting is a write only operation. The ingested objects sync to the App Services backend and are deleted from the device. It can not be read from the `Realm`.

test/asymmetric_test.dart
```diff
@@ -0,0 +1,117 @@
...
+    final realm = await getSyncRealm(config);
+    realm.syncSession.pause();
+
+    final oid = ObjectId();
+    realm.write(() {
+      realm.ingest(Asymmetric(oid, embeddedObjects: [1, 2, 3].map(Embedded.new)));
+    });
+
+    // Find & query on an Asymmetric object is compile time error, but you can cheat with dynamic
+    expect(realm.dynamic.find('Asymmetric', oid), null); // should this throw instead?
```
blagoev, line 71: On the comment I think this should behave as the case for not found `RealmObject`.

blagoev (approved):
left some small comments. I think its good. Should we have a final discussion on `realm.ingest` vs `realm.add` before we merge?

test/asymmetric_test.dart
nielsenko: I'll remove question

nielsenko:
> left some small comments. I think its good. Should we have a final discussion on `realm.ingest` vs `realm.add` before we merge?

Yes, I cannot make up my mind myself. I'm okay with both, but it cannot be as elegant as in realm-dotnet since we cannot overload `add`. Let us discuss Monday.

lib/src/realm_class.dart
nielsenko: Updated

nielsenko:
> > left some small comments. I think its good. Should we have a final discussion on `realm.ingest` vs `realm.add` before we merge?
> 
> Yes, I cannot make up my mind myself. I'm okay with both, but it cannot be as elegant as in realm-dotnet since we cannot overload `add`. Let us discuss Monday.

I have created a [commit](https://github.com/realm/realm-dart/pull/1400/commits/3864e711dc1c94ab6d850ed2a19314595aa3b4db) that illustrates how it could be implemented, if we go with `add`

generator/test/good_test_data/asymmetric_object.expected
```diff
@@ -0,0 +1,112 @@
...
+class Asymmetric extends _Asymmetric
+    with RealmEntity, RealmObjectBase, RealmObject {
+  Asymmetric(
+    ObjectId id, {
+    Embedded? farther,
+    Embedded? mother,
+    Iterable<Embedded> children = const [],
+  }) {
+    RealmObjectBase.set(this, '_id', id);
+    RealmObjectBase.set(this, 'farther', farther);
```
blagoev, line 14: nit: *father vs farther.

generator/test/error_test_data/asymmetric_external_link.expected
```diff
@@ -0,0 +1,12 @@
...
+
+in: asset:pkg/test/error_test_data/asymmetric_external_link.dart:12:16
+   ╷
+10 │ @RealmModel()
+11 │ class _Bad {
+   │       ━━━━ in realm model for 'Bad'
+12 │   _Asymmetric? asymmetric;
+   │                ^^^^^^^^^^ !
+   ╵
+Remove the field
```
desistefanova, line 11: It would be nice to give to the user some more explanation about this error "Remove the field".

test/asymmetric_test.dart
```diff
@@ -0,0 +1,117 @@
...
+class _Symmetric {
+  @PrimaryKey()
+  @MapTo('_id')
+  late ObjectId id;
+}
+
+Future<void> main([List<String>? args]) async {
+  await setupTests(args);
+
+  Future<Realm> getSyncRealm(AppConfiguration config) async {
```
desistefanova, line 52: Is it allowed to use Asymmetric schema in a local realm configuration? If not, do we have a validation?

nielsenko: No it is not, and yes

nielsenko: Added a test

desistefanova: I don't see it, probably it is in core. If you don't mind, you can add some simple test about it.

generator/test/error_test_data/asymmetric_external_link.expected
nielsenko: The header line of the error is: _Linking to asymmetric objects is not allowed_

Isn't that the context you need? The last line is the suggested action. The only other option would be to make `_Asymmetric` not be asymmetric. Should we suggest that as well?

desistefanova: Sorry, I've missed the title. Yes, it is enough.

nielsenko:
> > left some small comments. I think its good. Should we have a final discussion on `realm.ingest` vs `realm.add` before we merge?
> 
> Yes, I cannot make up my mind myself. I'm okay with both, but it cannot be as elegant as in realm-dotnet since we cannot overload `add`. Let us discuss Monday.

For now we will go with `ingest`. I have parked the refactoring commit that _overload_ `add` on the [kn/asymmetric-add-instead-of-ingest](https://github.com/realm/realm-dart/tree/kn/asymmetric-add-instead-of-ingest) branch

blagoev (approved):

desistefanova (approved):

PR #1402, opened 2023-09-15
kn/asymmetric object allow non embedded links

nielsenko:
- Lift link restriction on asymmetric objects

Fixes: #1403


nielsenko:
This requires realm-core v13.21.0 or later

nielsenko:
@nirinchev I noticed that current CHANGELOG on main says realm-core 13.20.1, but it is really v13.23.2-4-g56a048d9f (ie. v13.23.2 + the XCode 15 fix)

nirinchev:
Yes, I saw that as well, but figured I'll update to the latest stable Core on Friday and just fix it then.

Issue #1409, opened 2023-10-02
.skip doesn't work well with .take


nirinchev:
Reported in [this forum post](https://www.mongodb.com/community/forums/t/new-skip-implementation-doesnt-work-with-take/247144), it appears https://github.com/realm/realm-dart/pull/1391 broke combining skip and take.

dotjon0:
Thanks @nirinchev!

nielsenko:
🙏

PR #1410, opened 2023-10-03
kn/fix skip then iterate bug

nielsenko:
Pass `_skipIndex` when snapshotting results, and add missing test

Fix: #1409


nielsenko:
Anyone?

PR #1445, opened 2023-12-04
Rework app isolate API

nirinchev:
* Adds `App.getById` to be used on background isolates.
* Pulls in https://github.com/realm/realm-core/pull/7182 to start constructing cached apps.
* Logs a warning when `App(AppConfiguration)` is used outside of the main isolate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 ## vNext (TBD)
 
 ### Enhancements
-* None
+* Reworked how creating an `App` instance works across isolates:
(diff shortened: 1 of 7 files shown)


tgoyne (approved):
Obviously I can't judge if the dart code is good but this all looks reasonable to me.

nirinchev:
Thanks @tgoyne, appreciate the review! I mostly needed a high-level sanity check to make sure this approach is sound and in line with how the Core API are supposed to be used.

PR #1457, opened 2024-01-02
Downgrade min flutter to 3.10.0 and min dart to 3.0.0

nielsenko:


nielsenko:
> Perhaps it makes sense to try running the ci.yml to install the minimum supported flutter/dart versions, otherwise we're running this with latest flutter/dart.

I agree (and that is my plan), but perhaps we should land #1441 first to reduce the chance of conflicts?

nirinchev:
Yeah, I was going to do it, but there is a regression in baasaas that we need to resolve first 😑

nielsenko:
CI will still run latest stable for this PR. To see a CI run where we use Flutter 3.10 / Dart 3.0 see #1474 (which will not be merged).

Issue #1448, opened 2023-12-10
Creating Migrations in a local first RealmDb Configuration


rasheedmhd:
### Problem

I am using RealmDb in a Flutter App Project, in my configuration, I am only doing a local storage to the device and no cloud syncing,

```
final config = Configuration.local([
//
]);

final realm = Realm(config);
```

I did changes to my model, and RealmDb says I need to migrate, I was hoping there would be a command to get that done but  I haven't seen anything like that,

I saw this but it doesn't help much in my situation, https://www.mongodb.com/developer/products/realm/realm-schema-migration/
My last resort would be to delete the local db (wherever, I will find it and what ever it looks like) and regenerating the schema again, which by the way doesn't work because it seems RealmDB is using the already created Db, if I knew at which point of the configuration that Db was created it would have been helpful but I don't know.

I was hoping that by `dart run realm generate` creates the Db so if I deleted it a new one would be created but nope!

How do I do migrations in a local config setup ?

### Solution

Can we have something like `dart run realm migrate`?

This could run `dart run realm generate` under the hood, (I saw this based on how I a small part of realm generate works) there could be a gazillion reasons why `realm migrate` shouldn't run it under the hood but what I am hoping to say is to have `realm migrate` rebuild the schema replacing ` schemas.g.dart` and updating the Db for everything to just work.

If this can not be done I hope to learn why and what are the work arounds, I would really know to learn, thanks. 

### Alternatives

I have seen documentation on adding a schema version in the configuration, I haven't seen one that is well written and talks about how to manage that in local configurations, a good documentation on how to do that with local configuration would help tremendously for the time been.

### How important is this improvement for you?

Would be a major improvement

### Feature would mainly be used with

Local Database only

rasheedmhd:
**UDATE**

Someone shared this with me in Ghana Flutter Group and it helped, just too bad for me that I couldn't find it, I just I am just dumb with a dead google fu. https://www.mongodb.com/docs/realm/sdk/flutter/realm-database/model-data/update-realm-object-schema/

It would be create to add a resource like this is the Error Message that appears is VS Code that tells the developer that they need to do a Migration.

PR #1459, opened 2024-01-05
Add MigrationRequiredException

nirinchev:
Fixes https://github.com/realm/realm-dart/issues/1448


nielsenko (approved):
Looks good.

One thing. Previously we mostly just threw RealmException on errors. We might benefit from using a similar pattern of creating a separate exception type in other situations as well.

PR #1461, opened 2024-01-08
Switch to integration tests

nielsenko:
This PR [migrates](https://docs.flutter.dev/release/breaking-changes/flutter-driver-migration) to using integration tests, as recommended by the flutter team. 

To run the test use:
```sh
flutter test integration_test/all_tests.dart
  --dart-define=BAAS_URL=$BAAS_URL
  --dart-define=BAAS_DIFFERENTIATOR=$BAAS_DIFFERENTIATOR 
  --file-reporter=json:test-results.json 
  --suppress-analytics 
  --device-id=... # choose device here
```
If you run BAAS on localhost, remember that on android you should use [10.0.2.2](https://developer.android.com/studio/run/emulator-networking) to reach it.

The PR also:
- Runs android tests roughly 3 times faster.
- Avoid test symlinks, so no duplicated files in your IDE.
- Get rid of relative imports to lib folder (`import '../lib/...';`) following on #1475.
- Allows test reports to be generated, for flutter tests as well.

You can (and should) use `--dart-define` to pass `BASS_URL`, etc. to the host. Tests will still read these values from the shell environment if available, that is if host and driver runs on the same node (ie. linux, macos, and windows). But you need to use `--dart-define` when running on the emulators or real devices. 

Similar data files used in tests needs to be added as assets to the flutter tests project.

Fixes: #1465


nielsenko:
Dropping CHANGELOG entry as this is an internal change to how we structure our tests

flutter/realm_flutter/tests/integration_test/all_tests.dart
```diff
@@ -0,0 +1,69 @@
...
+import '../../../../test/realm_logger_test.dart' as realm_logger_test;
+import '../../../../test/realm_map_test.dart' as realm_map_test;
+import '../../../../test/realm_object_test.dart' as realm_object_test;
+import '../../../../test/realm_set_test.dart' as realm_set_test;
+import '../../../../test/realm_test.dart' as realm_test;
+import '../../../../test/realm_value_test.dart' as realm_value_test;
+import '../../../../test/results_test.dart' as results_test;
+import '../../../../test/session_test.dart' as session_test;
+import '../../../../test/subscription_test.dart' as subscription_test;
+import '../../../../test/user_test.dart' as user_test;
```
nielsenko, line 30: This is still maintained by hand for now. Integration tests can run each file separately just fine, but it is prohibitively slow to build and deploy a separate app per test file.

I have created an issue to fix this: https://github.com/realm/realm-dart/issues/1490

flutter/realm_flutter/tests/integration_test/all_tests.dart
```diff
@@ -0,0 +1,69 @@
...
+  group('realm_logger_test.dart', realm_logger_test.main);
+  group('realm_map_test.dart', realm_map_test.main);
+  group('realm_object_test.dart', realm_object_test.main);
+  group('realm_set_test.dart', realm_set_test.main);
+  group('realm_test.dart', realm_test.main);
+  group('realm_value_test.dart', realm_value_test.main);
+  group('results_test.dart', results_test.main);
+  group('session_test.dart', session_test.main);
+  group('subscription_test.dart', subscription_test.main);
+  group('user_test.dart', user_test.main);
```
nielsenko, line 68: Currently we need to add groups manually as new test files are added (similar to how it always was).

I have created an issue to fix this: https://github.com/realm/realm-dart/issues/1490

flutter/realm_flutter/tests/integration_test/all_tests.dart
```diff
@@ -0,0 +1,69 @@
...
+  final bytes = data.buffer.asUint8List(data.offsetInBytes, data.lengthInBytes);
+  await File(toPath).writeAsBytes(bytes);
+}
+
+void main() {
+  IntegrationTestWidgetsFlutterBinding.ensureInitialized();
+
+  // To support both dart test and flutter integration test we pass an alternative
+  // copyFile function. Remember to add any needed files as assets in pubspec.yaml.
+  configuration_test.copyFile = _copyBundledFile;
```
nielsenko, line 43: Trick used to get rid of the duplicated `flavor_helpers.dart` that are imported via relative imports.

flutter/realm_flutter/tests/android/app/build.gradle
```diff
@@ -21,21 +22,31 @@ if (flutterVersionName == null) {
...
+    sourceSets {
+        main.java.srcDirs += 'src/main/kotlin'
+    }
+
     defaultConfig {
-        applicationId "io.realm.tests"
-        minSdkVersion 16
-        targetSdkVersion 30
+        // TODO: Specify your own unique Application ID (https://developer.android.com/studio/build/application-id.html).
+        applicationId "com.example.tests"
```
nirinchev, line 45: Should we specify an app id here - like `com.mongodb.realm-dart-tests`?

flutter/realm_flutter/tests/linux/CMakeLists.txt
```diff
@@ -1,13 +1,13 @@
...
+cmake_minimum_required(VERSION 3.10)
 project(runner LANGUAGES CXX)
 
 # The name of the executable created for the application. Change this to change
 # the on-disk name of your application.
 set(BINARY_NAME "tests")
 # The unique GTK application identifier for this application. See:
 # https://wiki.gnome.org/HowDoI/ChooseApplicationID
-set(APPLICATION_ID "io.realm.tests")
+set(APPLICATION_ID "com.example.tests")
```
nirinchev, line 10: Same here - I guess it doesn't really matter, but should we use a more branded id?

flutter/realm_flutter/tests/pubspec.yaml
```diff
@@ -21,19 +21,14 @@ dependencies:
...
+    sdk: flutter
 
 flutter:
-  uses-material-design: true
   assets:
-    - data/realm_files/old-format.realm
-
-dependency_overrides:
-  test_api: ^0.6.1
+    - test/data/realm_files/old-format.realm
```
nirinchev, line 34: If we're defining the test data files here, can we get rid of the symlink and instead use `../../.../../test/data/realm_files/old-format.realm`?

src/realm_dart_decimal128.cpp
```diff
@@ -25,13 +25,13 @@ RLM_API realm_decimal128_t realm_dart_decimal128_from_string(const char* string)
     return to_decimal128(result);
 }
 
+// This buffer is reused between calls, hence it is thread local
+thread_local char _decimal128_to_string_buffer[34];
```
nirinchev, line 29: This appears to be pulled from another PR.

test/app_test.dart
```diff
@@ -24,23 +24,27 @@ import 'package:test/expect.dart' hide throws;
...
 import 'test.dart';
 
-Future<void> main([List<String>? args]) async {
-  await setupTests(args);
+void main() {
+  setupTests();
 
   test('AppConfiguration can be initialized', () {
-    Configuration.defaultRealmPath = path.join(Configuration.defaultStoragePath, Configuration.defaultRealmName);
+    Configuration.defaultRealmPath = path.join(
```
nirinchev, line 35: Did something change about the max line length here?

test/baas_helper.dart
```diff
@@ -97,38 +86,41 @@ class BaasHelper {
...
+      final baasAasApiKey = Env.baasAasApiKey();
+      if (baasAasApiKey != null) {
+        if (cluster != null) {
+          throw "BAAS_BAASAAS_API_KEY can't be combined with BAAS_CLUSTER";
         }
 
-        (baasUrl, _) = await BaasClient.retry(() => BaasClient.getOrDeployContainer(baasaasApiKey, differentiator));
+        (baasUrl, _) = await BaasClient.retry(() => BaasClient.getOrDeployContainer(
+              baasAasApiKey,
+              differentiator!,
```
nirinchev, line 112: Is this safe? As far as I can tell, there's nothing that guarantees that a differentiator is supplied.

test/test.dart
```diff
@@ -388,27 +388,22 @@ void xtest(String? name, dynamic Function() testFunction, {dynamic skip, Map<Str
...
-  testName = _testArgs["name"];
 
+void setupTests() {
   setUpAll(() async {
-    baasHelper = await BaasHelper.setupBaas(_testArgs);
-  });
+    baasHelper = await BaasHelper.setupBaas();
 
-  setUp(() {
     Realm.logger = Logger.detached('test run')
```
nirinchev, line 396: This seems to have been moved from setUp to setUpAll - is this intentional? This way, if we replace the logger in a test, wouldn't that logger continue to be used in subsequent tests?

test/test.dart
```diff
@@ -388,27 +388,22 @@ void xtest(String? name, dynamic Function() testFunction, {dynamic skip, Map<Str
...
       ..level = Level.ALL
       ..onRecord.listen((record) {
-        if (record.level.value >= RealmLogLevel.warn.value) {
-          print('${record.time} ${record.level.name}: ${record.message}');
-        }
-
```
nirinchev, line 408: Is this intentional? I guess there's still value in printing warnings/errors, even if the test succeeds.

test/configuration_test.dart
```diff
@@ -638,3 +636,9 @@ Future<void> main([List<String>? args]) async {
     expect(() => realm.write(() {}), throws<RealmException>("Number of active versions (3) in the Realm exceeded the limit of 2"));
   });
 }
+
+Future<void> _copyFile(String fromPath, String toPath) async {
+  await File(fromPath).copy(toPath);
+}
+
+var copyFile = _copyFile; // default, but allow integration_test to override
```
nirinchev, line 644: Do we want this to live in `configuration_test.dart` or more generally in `test.dart`? I feel like that's where we put helper functions like this one.

nirinchev (commented):
Good stuff - only minor nits from me.

flutter/realm_flutter/tests/pubspec.yaml
nielsenko: We need the asset path and the actual path to be the same currently, for the file to be copied correctly. But I guess we could do something clever.

src/realm_dart_decimal128.cpp
nielsenko: Yes, sorry.. will fix.

test/configuration_test.dart
nielsenko: I'll move it. Currently it is only used by `configuration_test.dart`

test/test.dart
nielsenko: It is intentional, but perhaps a bit opinionated :-). If the test fails, it will be included in the trace regardless. I'm okay with adding it back, if you prefer.

flutter/realm_flutter/tests/linux/CMakeLists.txt
nielsenko: Forgot to add `--org` flag when I re-created the platform projects. I have returned to using `io.realm`.

test/app_test.dart
nielsenko: Apparently so.. fixed

test/baas_helper.dart
nielsenko: You are right. Will fix and use 'local' as default.

flutter/realm_flutter/tests/pubspec.yaml
nielsenko: I suggest leaving the data symlink for now.

test/test.dart
nielsenko: Yes, the logger is static to the isolate anyway. It will not be torn down regardless. Notice how the log tests use `Isolate.run` to not mess with the primary logger. We might as well move it here.

nirinchev: This doesn't sound correct to me - maybe I'm missing something, but the Realm Core logger is very much a singleton, so setting it up on a different isolate just ends up replacing it.

nielsenko: I'm not sure if something has happened in my absence. But the Dart side logger is per isolate, and the static core side logger will forward messages to each isolate logger and keep track of the log level of each.

nirinchev: Hm, I'll try and take a look.

nirinchev: Yeah, you're right - I had completely forgotten about how this was done in Dart. Let's keep it like this then.

nirinchev (approved):

PR #1464, opened 2024-01-09
Update to Core next-major

nirinchev:
This will be the base PR on top of which collections in mixed will be based.


test/realm_map_test.dart
```diff
@@ -630,12 +630,12 @@ testNotifications<T>(RealmMap<T> Function(TestRealmMaps) accessor, TestCaseData<
       realm.add(testObject);
     });
 
-    // final managedMap = accessor(testObject);
-    // await runManagedNotificationTests(testData, managedMap);
+    final managedMap = accessor(testObject);
+    await runManagedNotificationTests(testData, managedMap, testObject);
```
nirinchev, line 634: These were accidentally committed as commented out.

nielsenko (approved):
Looks good.

I'm wondering about the timing? Will we be releasing before next-major is deemed stable? If so, we should probably wait with merging this a bit.

nirinchev:
next-major in Core should be released soon, though I'm just putting this as a baseline on top of which I'll be implementing collections in mixed. I don't think we need to merge this until collections in mixed is ready.

Issue #1451, opened 2023-12-14
Cannot Open Realm in Background Isolate


rsanford-hh2:
### What happened?

See https://github.com/realm/realm-dart/issues/1055 for a similar issue... 

I'm trying to follow the doc from here : https://www.mongodb.com/docs/realm/sdk/flutter/sync/sync-multiple-processes/... 

The basic instructions from that are to have a single realm using the flexible sync config and then any other workers using a disconnected sync configuration.

I'm trying to do that and am getting the following exception.
`RealmException (RealmException: Error opening realm at path /Users/ebenezer/.hh2mobile.realm. Error code: 2013 . Message: Realm at path '/Users/ebenezer/.hh2mobile.realm' already opened with different sync configurations.)`

### Repro steps

Look at the code in the code snippet below. It does the following
- Starts up a flexible sync realm
- adds subscriptions to listen for changes to a collection of objects
- spawns an isolate that will open the same realm with a disconnected sync config
  - the isolate will loop to scan for new items to process
- the main thread will loop and run a query every so often 

Note that the way the writeRealm is initialized is the same as the issue that I mention in the "What happened?" section except that reporter used a memory config where I'm using a disconnected sync config.

What am I doing wrong?

### Version

Dart SDK version: 3.2.3 (stable) (Tue Dec 5 17:58:33 2023 +0000) on "macos_arm64"

### What Atlas Services are you using?

Atlas Device Sync

### What type of application is this?

Dart standalone application

### Client OS and version

macOS Sonoma 14.2

### Code snippets
```
import 'dart:async';
import 'dart:io';
import 'dart:isolate';

import 'package:path/path.dart' as path;
import 'package:realm_dart/realm.dart';

import 'package:mongodart/realmmodels/models.dart';

String get _homeDir => Platform.environment['HOME'] ?? (Platform.environment['APPDATA'] != null ? path.join(Platform.environment['APPDATA']!, 'Local') : null) ?? Platform.environment['USERPROFILE'] ?? '.';
String get _realmPath => path.join(_homeDir, '.hh2mobile.realm');

final schemaObjects = [RealmReceipt.schema, RealmFileRecordLinkChunk.schema, RealmCreditCard.schema, RealmFileRecordLink.schema, RealmUser.schema];
final writeConfig = Configuration.disconnectedSync(schemaObjects, path: _realmPath);
final writeRealm = Realm(writeConfig);

class _tusConfig {
[...]

Issue #1433, opened 2023-11-22
[Bug]: Unable to open realm on background isolate


geosebas:
### What happened?

Code that worked and still work on realm 1.5.0 stop working since update. I downgraded for now, so it's not urgent. 

I currently don't have time to create a sample and test sample with the new version, but since it work nicely on 1.5 and not with 1.6 and I didn't see anything on the changelog related to this, I assume the problem here is on your side. 

### Repro steps

Nearly a years ago I got the same problem (issue here https://github.com/realm/realm-dart/issues/1055) 

### Version

Flutter 3.16.0

### What Atlas Services are you using?

Both Atlas Device Sync and Atlas App Services

### What type of application is this?

Flutter Application

### Client OS and version

With realm 1.6.0 on both android and ios

### Stacktrace of the exception/crash you're getting

```shell
flutter: [ERROR] : PlayerSearchHelper.initData cannot be completed :
flutter: RealmException: Error opening realm at path /Users/geoffreysebastianelli/Library/Developer/CoreSimulator/Devices/3770FA2B-8574-473C-A9A7-ECDF0BD3F659/data/Containers/Data/Application/376EA245-397C-4FDC-9660-CB4A02CD5437/Documents/mongodb-realm/flex.638118ad44927095598ea5fa.realm. Error code: 2013 . Message: Realm at path '/Users/geoffreysebastianelli/Library/Developer/CoreSimulator/Devices/3770FA2B-8574-473C-A9A7-ECDF0BD3F659/data/Containers/Data/Application/376EA245-397C-4FDC-9660-CB4A02CD5437/Documents/mongodb-realm/flex.638118ad44927095598ea5fa.realm' already opened with different sync user.
flutter: #0      _RealmCore.throwLastError.<anonymous closure> (package:realm/src/native/realm_core.dart:156:7)
flutter: #1      using (package:ffi/src/arena.dart:124:31)
flutter: #2      _RealmCore.throwLastError (package:realm/src/native/realm_core.dart:150:5)
flutter: #3      _RealmLibraryEx.invokeGetPointer (package:realm/src/native/realm_core.dart:2993:17)
flutter: #4      _RealmCore.openRealm (package:realm/src/native/realm_core.dart:661:32)
flutter: #5      Realm._openRealm (package:realm/src/realm_class.dart:232:22)
flutter: #6      new Realm._ (package:realm/src/realm_class.dart:178:98)
[...]

PR #1470, opened 2024-01-11
Fix use of http transport from background isolate

nielsenko:
After the introduction of realm app caching hot-restart and background isolates no longer works reliably with realm.

Fixes: #1467 
Fixes: #1466
Fixes: #1451 
Fixes: #1433 (only half fixed by #1445)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,6 +5,12 @@
     "version": "0.2.0",
     "configurations": [
         {
+            "name": "Attach",
+            "type": "lldb",
+            "request": "attach",
+            "pid": "${command:pickProcess}",
+        },
+        {
             "name": "Flutter: Attach to Device",
             "type": "dart",
             "request": "attach"
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -153,6 +153,13 @@
     "__mutex_base": "cpp",
     "coroutine": "cpp",
     "ranges": "cpp",
-    "span": "cpp"
+    "span": "cpp",
+    "__bit_reference": "cpp",
+    "__bits": "cpp",
+    "__config": "cpp",
+    "__debug": "cpp",
+    "__split_buffer": "cpp",
+    "__tuple": "cpp",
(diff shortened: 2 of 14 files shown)


nielsenko:
The issues fixed in this PR was introduced with , but was never released, so I didn't update the CHANGELOG.

Since Apps are now cached and shared we need to be vigilant to always route to the correct isolate in cases such as `App.logIn` as we potentially need to do the request in another isolate than the one calling, and hence handle async native callbacks in both. 

Hence we need to ensure all data is available not only when the callback is scheduled, but also when it actually runs. The current realm C-api expects callback to be executed synchronously, so we need to deep-copy all non-value arguments and handle lifetime explicitly. This involves a bit of simple and dumb native code, such as: 
```c++
RLM_API void realm_dart_user_completion_callback(realm_userdata_t userdata, realm_user_t* user, const realm_app_error_t* error)
{
    // we need to make a deep copy of error, because the message pointer points to stack memory
    auto error_copy = realm_app_error_copy(error);

    // take an extra ref to the user, so that it doesn't get deleted before we invoke the callback
    std::shared_ptr<realm::SyncUser> user_copy; 
    if (user != nullptr) {
        user_copy = realm_user(*user);
    }

    ud->scheduler->invoke([ud, error = std::move(error_copy), user = realm_user(user_copy)]() mutable { 
        (reinterpret_cast<realm_app_user_completion_func_t>(ud->dart_callback))(ud->handle, &user, error.get());
    });
}
```

Going forward we could:

1. Re-work app caching in core to be better suited for Darts memory model,
2. Create Dart specific C-api with the new binding generator introduced for realm-js, that is geared towards async callbacks,
3. Wait for https://github.com/dart-lang/sdk/issues/54554#issuecomment-1887127643, which should allow us to get rid of these trampolines, or
4. Just keep adding these trampolines as needed.

BTW: Note that (unrelated to the fix) the Dart native interface has been bumped to 2.2, which is backwards compatible with 2.0 that we used previously.

PR #1472, opened 2024-01-15
MVP of fixing the app cache issues

nirinchev:


nirinchev:
Closing this in favor of https://github.com/realm/realm-dart/pull/1470. The main issue with this approach is that isolates don't map reliably to threads so using thread local storage is not safe.

PR #1474, opened 2024-01-16
kn/run ci with flutter 3.10

nielsenko:
- Run ffigen after downgrade
- Use flutter 3.10.0 on CI (temporary test)
- Hack to run driver tests with Flutter 3.10 (Dart 3.0)

This PR should not be merged. It exists to test realm with Flutter 3.10 (Dart 3.0), instead of latest stable.

NOTE: We should rework CI to run the full suite on multiple Flutter / Dart versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -307,6 +307,7 @@ jobs:
       - name: Setup Flutter
         uses: subosito/flutter-action@v2
         with:
+          flutter-version: '3.10.0'
           channel: 'stable'
 
       - name: Install dependencies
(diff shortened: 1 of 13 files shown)


nielsenko:
Notice the dependency_override hack needed to actually run the tests with flutter 3.10
```yaml
dependency_overrides:
  realm_common:
    path: ../../../common 
  realm_generator: 1.6.0
```
This is due to both `realm` and `test` pulling in `analyzer` which is pinned by `realm` ^1.6.1 at ^6.0.0. All versions of `test` supporting this is incompatible with `flutter_test` (and `flutter_drive` for that matter) from the 3.10 SDK.

PR #1475, opened 2024-01-17
Avoid lib symlink

nielsenko:
- Export realm_dart.dart from realm package


nielsenko:
There should be no user visible change, so I have skipped the CHANGELOG entry.

flutter/realm_flutter/tests/lib
```diff
@@ -1 +1 @@
-../lib
\ No newline at end of file
+../../../lib/
```
nirinchev, line 1: I know this is not a new thing, but I'm curious why we need to symlink lib inside the tests app? AFAICT the test project takes a dependency on the `realm` package, so I would expect it to pull the lib source code from there.

nirinchev (approved):
Looks good. So now the realm package in pub.dev will have a dependency on realm_dart for the dart source code and just add some native sources that pull the binaries automatically rather than relying on an install command?

flutter/realm_flutter/tests/lib
nielsenko: Indeed.. I just wanted to do the test refactoring in a separate PR as to not drown this one in small import updates.

nielsenko:
> Looks good. So now the realm package in pub.dev will have a dependency on realm_dart for the dart source code and just add some native sources that pull the binaries automatically rather than relying on an install command?

Yes, and we just re-export all that is exported by realm_dart. It should not make a difference to our users.

PR #1491, opened 2024-01-25
Always use persistent handles when passing dart objects as userdata to callbacks

nielsenko:
Fixes: #1492

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,13 @@
 * None
 
 ### Fixed
-* None
+* Fix a possible hang (or in rare cases crash) during notification handling. (Issue [#1492](https://github.com/realm/realm-dart/issues/1492))
 
 ### Compatibility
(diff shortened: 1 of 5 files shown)


nielsenko:
I have not added new tests for this, as it needs to run fairly long to provoke it.

nirinchev (approved):
This looks good - is there a way to run a test to see if we leak any resources with this change? (I don't think we should but it's the kind of change that carries this risk)

nielsenko:
> This looks good - is there a way to run a test to see if we leak any resources with this change? (I don't think we should but it's the kind of change that carries this risk)

We could look into using the `vmService` like I do here: https://gist.github.com/nielsenko/3f3cda354931d9179d1028e3df03340d.

Or perhaps better to use the _fairly_ new [leak_tracker](https://pub.dev/packages/leak_tracker) package

I have created a separate issue for that https://github.com/realm/realm-dart/issues/1494

Issue #1485, opened 2024-01-23
Avoid nested packages


nielsenko:
Instead if nesting `realm_common`, `realm_generator`, and `realm` package directories within the `realm_dart` package directory, I suggest moving to a setup like:
```
realm-dart/
  realm/
    packages/
      realm_common/
      realm_generator/
      realm_dart/
      realm/
  ejson/
    packages/
      ...
  atlas/
    packages/
      ...   
```
This way no packages exists within the directory structure of another package.

nirinchev:
As part of this, we should try and remove the src symlink in flutter ios.

nielsenko:
> As part of this, we should try and remove the src symlink in flutter ios.

Many(/most) symlinks will die, but the symlinks in the flutter platform plugins will remain until we complete #1515.

nielsenko:
> > As part of this, we should try and remove the src symlink in flutter ios.
> 
> Many(/most) symlinks will die, but the symlinks in the flutter platform plugins will remain until we complete #1515.

Here I'm referring to the symlinks to the binary folder, ie:
```
realm_dart.xcframework@ -> ../../../binary/ios/realm_dart.xcframework
```
I realise now, that is not what you are asking about 🤦

nirinchev:
Oh, great! Yeah, I was surprised this had to stay so created a quick PR to see what happens if I remove it 😄 Apparently, it can be removed, which makes me happy - it always irked me that we ship the entire Core with our flutter package.

PR #1495, opened 2024-01-26
kn/new repo structure

nielsenko:
Fixes: #1485


.github/workflows/publish-release.yml
```diff
@@ -85,41 +86,41 @@ jobs:
...
-        rm flutter/realm_flutter/ios/src
-        rm flutter/realm_flutter/ios/realm_dart.xcframework
-        rm flutter/realm_flutter/linux/binary
-        rm flutter/realm_flutter/windows/binary
-        rm example/binary
+        rm -rf packages/realm/android/src/main/cpp/lib
+        rm packages/realm/ios/src
+        rm packages/realm/ios/realm_dart.xcframework
+        rm packages/realm/linux/binary
+        rm packages/realm/windows/binary
```
nirinchev, line 110: I know you just updated some paths, but I can see we're not removing the macos binary symlink and it's not obvious to me why.

.github/workflows/publish-release.yml
```diff
@@ -85,41 +86,41 @@ jobs:
...
+        cp -Lr ../packages/realm realm
         cd realm
         flutter pub publish --dry-run || true
       working-directory: release
 
     - name: Package realm_dart
       run: |
         mkdir realm_dart
-        cp -Lr ../bin ../example ../lib ../test realm_dart
+        cp -Lr ../package/realm_dart/bin ../package/realm_dart/example ../package/realm_dart/lib ../package/realm_dart/test realm_dart
```
nirinchev, line 123: Do we need this convoluted copying now that `realm_dart` is in a dedicated folder?

.gitmodules
```diff
@@ -1,4 +1,4 @@
-[submodule "src/realm-core"]
-	path = src/realm-core
-	url = https://github.com/realm/realm-core
+[submodule "packages/realm_dart/src/realm-core"]
```
nirinchev, line 1: Can the native bits live outside of `realm_dart` or do they have to be in a subfolder? Not a strong opinion, but I mildly prefer to have the native bits - both core and our custom native code - outside of the dart packages.

packages/realm/android/src/main/cpp/lib/arm64-v8a
```diff
@@ -0,0 +1 @@
+../../../../../../realm_dart/binary/android/arm64-v8a/
```
nirinchev, line 1: Can we have a single symlink in `packages/realm/android/src/main/cpp/lib` pointing `../../../../../../realm_dart/binary/android/`? Or is there something else in the source dir we don't want to link here?

nirinchev (commented):
The realm.podspec now seems to refer to non-existent files in `HEADER_SEARCH_PATHS` (now that we deleted the src symlink).

Overall looks good - we'll need to test the release workflow out, but I don't see anything concerning

.github/workflows/publish-release.yml
nielsenko: No, I decided to split the branch to reduce the size of the PR, so it ended up being  addressed in the follow up PR #1516. 

Actually I'm wondering why we even store packages in the release tarball and upload to s3? They are available for perpetuity on pub.dev?

packages/realm/android/src/main/cpp/lib/arm64-v8a
nielsenko: No, and we could. I was just doing the least intrusive change. Will fix.

.gitmodules
nielsenko: I was thinking about that as well.

The long term goal here is to move to use the new native-asset feature. They suggest keeping the native src, in `<package>/src` and the dart src in `<package>/lib/src`, so I'm inclined to stick with that.

However, I do intent to move all dart code in `realm_dart`, except the bindings (`native/realm_bindings.dart`) to `realm` and have `realm_dart` be a simple shim around the native code... Once native-assets land.

Would that sit with you?

.github/workflows/publish-release.yml
nielsenko: 🤔 .. seems like an inherited mistake. 

It doesn't make it into the pub packages (or the packages.tar.gz) though. Just checked 1.9.0.
I suppose we are "saved" by the way tar handles soft links pointing outside the folder, so all these removes are really redundant? 

Anyway, goes back to my question, why are we even tarball'ing the packages and storing on s3?

nielsenko: No that doesn't compute, since we use `cp -Lr` which follows links. It is not removed, because there is no symlink to remove for macos.

Still, why upload these to both pub.dev and s3? I know, I should know, but it currently escapes me 😄

packages/realm/android/src/main/cpp/lib/arm64-v8a
nielsenko: Fixed

.github/workflows/publish-release.yml
nirinchev: We upload to S3 for prerelease builds. Since the release workflow is running the Staging part first, regardless of whether we deploy to production or not, those will be uploaded to S3. I believe these symlinks point into the void, which is why we wanted to remove them, though it was a long time ago and it's a bit fuzzy.

.gitmodules
nirinchev: I would prefer that we don't ship the native source code with the dart packages since it's unusable, which is why I'd like to see it live outside the packages. Even with native assets, I'd probably prefer that we ship just the prebuilds unless we believe there's a use case where someone might want to build the library locally.

nielsenko: The src won't be in the package on pub.dev. It isn't today either. It will just be a submodule you can call in with git - if you need to build the native assets. For end users it will just be downloaded from S3.

.github/workflows/publish-release.yml
nielsenko: You can push pre-release packages to pub.dev as well. This makes it much easier for end-users to consume them. Also, it we don't like that, we can run a private pub.dev server.. as I recall it is even using mongodb 😄 

Anyway, I haven't changed the fact that we upload to S3.. just something to consider in the future.

PR #1521, opened 2024-02-09
Update .vscode cmake.sourceDirectory

nirinchev:


nielsenko:
The title seems to imply that the CMakePreset file will be moved, but the PR doesn't actually do that.
Could you elaborate a bit on what you want to achieve here?

analysis_options.yaml
```diff
@@ -29,6 +29,7 @@ linter:
   rules:
     avoid_relative_lib_imports: false
     package_api_docs: true
+    dangling_library_doc_comments: false
```
nielsenko, line 32: I prefer that we don't disable recommended lints, but let us keep it for now, and fix the code later.

nielsenko (approved):
The actual changes looks good to me, but I'm a bit confused by the title?

PR #1546, opened 2024-03-08
RDART-973: Add support for the new progress notifications

nirinchev:
Fixes https://github.com/realm/realm-dart/issues/1564

TODO:
- [x] Changelog


packages/realm_dart/test/session_test.dart
```diff
@@ -191,28 +192,97 @@ void main() {
...
+    uploadRealm.write(() {
+      uploadRealm.add(NullableTypes(ObjectId(), differentiator, stringProp: generateRandomString(50)));
+    });
+
+    await uploadRealm.syncSession.waitForUpload();
+    await downloadRealm.syncSession.waitForDownload();
+
+    expect(uploadRealm.all<NullableTypes>().length, downloadRealm.all<NullableTypes>().length);
+    expect(uploadData.callbacksInvoked, uploadCallbacks);
+    expect(downloadData.callbacksInvoked, downloadCallbacks);
```
nielsenko, line 230: Could you elaborate the explanation of this part a bit? Why will we not see more upload callbacks here, when we add and wait for upload?

packages/realm_dart/test/test.dart
```diff
@@ -418,7 +418,7 @@ void setupTests() {
 
     Realm.logger.setLogLevel(LogLevel.detail);
     Realm.logger.onRecord.listen((record) {
-      testing.printOnFailure('${record.category} ${record.level.name}: ${record.message}');
+      testing.printOnFailure('${DateTime.now().toUtc()} ${record.category} ${record.level.name}: ${record.message}');
```
nielsenko, line 421: 👍

nielsenko (approved):
LGTM.

Perhaps you could elaborate a bit on the explanation of the test?

nielsenko:
Some test failures needs to be resolved

packages/realm_dart/test/session_test.dart
nirinchev: Because the subscription we setup is for `ProgressMode.forCurrentlyOutstandingWork`, which ends when you've reached 100% and no further callbacks will be invoked as more data comes in/goes out.

Issue #1547, opened 2024-03-08
Upgrade CI to xcode 15 in prep for April app store requirements


nielsenko:
> Starting in April 2024, apps submitted to the App Store must be built with Xcode 15 and the iOS 17 SDK, tvOS 17 SDK, or watchOS 10 SDK (or later).

https://developer.apple.com/news/?id=khzvxn8a

nielsenko:
This is only relevant to apps for now.

PR #1548, opened 2024-03-08
RDART-962: Use xcode 15.4

nielsenko:
Fixes: #1547 

- Use macos-14 runner for arm64 and macos-13 runner for x86_64 (see https://github.com/actions/runner-images?tab=readme-ov-file)
- Use xcode 15.4 which changes a bit how to build universal binaries.

### DIFF
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -52,7 +52,7 @@ jobs:
         run: echo "ANDROID_NDK=$ANDROID_NDK_LATEST_HOME" >> $GITHUB_ENV
 
       - name: Select XCode for MacOS
-        if: startsWith(inputs.runner, 'macos')
+        if: runner.os == 'macOS'
         run: |
           xcodes installed
           sudo xcodes select 15.4
--- a/.github/workflows/build-native.yml
+++ b/.github/workflows/build-native.yml
@@ -51,9 +51,11 @@ jobs:
         if: startsWith(matrix.build, 'android')
         run: echo "ANDROID_NDK=$ANDROID_NDK_LATEST_HOME" >> $GITHUB_ENV
 
-      - name: Downgrade XCode for MacOS
-        if: matrix.build == 'macos'
-        run: sudo xcodes select 14.3.1
+      - name: Select XCode for MacOS
+        if: startsWith(inputs.runner, 'macos')
+        run: |
+          xcodes installed
+          sudo xcodes select 15.4
 
       - name: Build
         if: steps.check-cache.outputs.cache-hit != 'true'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -139,7 +139,7 @@ jobs:
     secrets: inherit
     with:
       runner: macos-14
-      architecture: arm
+      arch: arm64
       differentiator: dma${{ github.run_id }}${{ github.run_attempt }}
 
   cleanup-cluster-dart-macos-arm:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,9 +19,9 @@ jobs:
     name: Build Windows
     uses: ./.github/workflows/build-native.yml
     with:
-     runner: windows-latest
-     binary: windows
-     build: '["windows"]'
+      runner: windows-latest
+      binary: windows
+      build: '["windows"]'
 
   build-macos:
     name: Build MacOS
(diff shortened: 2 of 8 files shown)


.github/workflows/dart-desktop-tests.yml
```diff
@@ -3,18 +3,10 @@ name: Dart desktop tests
 on:
   workflow_call:
     inputs:
-      os:
```
nirinchev, line 6: I'm not sure I'm a huge fan of the changes in this file. I understand that it removes some redundant pieces of information, but it complicates the setup somewhat and forces readers to familiarize themselves with the architectures of the various github-hosted runners.

.github/workflows/ci.yml
```diff
@@ -296,7 +290,12 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          submodules: 'recursive'
+          submodules: "recursive"
+
+      - name: Select XCode for MacOS
```
nirinchev, line 295: We need to increase the ulimit here it seems.

CHANGELOG.md
```diff
@@ -51,7 +52,7 @@
     @override
     String toString() => 'Stuff{id: $id}';
   }
-
+
```
nirinchev, line 55:
```suggestion

```

.github/workflows/dart-desktop-tests.yml
nielsenko: The previous code had assumptions about the architecture, that didn't actually match the runners, implicitly assuming arch was `x86_64` unless `architecture` was set to `arm`. That is not correct for macos-14, which is being rolled out as `macos-latest` as of yesterday (https://github.com/github/roadmap/issues/926).

One thing I regret is that I cannot show the arch in the job title, as we don't know what runner is being used yet, so it is only shown in the steps. And obviously I would have preferred for github action expressions to be expressive enough to lowercase a simple string 🙄 

What if I reintroduce the `architecture` argument, and make a step that explicitly test that it actually match `runner.arch`?

CHANGELOG.md
nielsenko: I prefer a new-line after the class. I think it okay to correct small blemishes like this, even in the `CHANGELOG.md`

packages/realm_dart/scripts/build-macos.sh
```diff
@@ -7,4 +7,4 @@ set -o pipefail
 cd "$(dirname "$0")/.."
 
 cmake --preset macos
-cmake --build --preset macos --config MinSizeRel -- -destination "generic/platform=macOS"
+cmake --build --preset macos --config Release -- -destination "generic/platform=macOS"
```
nielsenko, line 10: `MinSizeRel` becomes `-Os` which apparently `clang++` (from XCode 15.x) don't like, hence this change.

packages/realm_dart/CMakePresets.json
```diff
@@ -168,7 +168,7 @@
       "configurePreset": "macos",
       "nativeToolOptions": [
         "-destination",
-        "platform=macOS"
+        "generic/platform=macOS"
```
nielsenko, line 171: As of XCode 15.3 it is apparently important to use `generic/platform=macOS` during preset, to actually end up with a universal binary (ie. supporting both `arm64` and `x86_64`).

The downside of this change is that we also build for `x86_64` on our dev machines, but given that we use `ccache` it is perhaps a non-issue in practice.

.github/workflows/dart-desktop-tests.yml
nirinchev: Sure, I'd be fine with that.

.github/workflows/build-native.yml
```diff
@@ -51,9 +51,11 @@ jobs:
         if: startsWith(matrix.build, 'android')
         run: echo "ANDROID_NDK=$ANDROID_NDK_LATEST_HOME" >> $GITHUB_ENV
 
-      - name: Downgrade XCode for MacOS
-        if: matrix.build == 'macos'
-        run: sudo xcodes select 14.3.1
+      - name: Select XCode for MacOS
+        if: startsWith(inputs.runner, 'macos')
```
nirinchev, line 55:
```suggestion
        if: runner.os == 'macOS'
```

.github/workflows/build-native.yml
```diff
@@ -51,9 +51,11 @@ jobs:
...
+        run: |
+          xcodes installed
```
nirinchev, line 57: Why do we need this?

.github/workflows/dart-desktop-tests.yml
```diff
@@ -56,18 +61,20 @@ jobs:
           melos setup
 
       - name: Bump ulimit on macos
-        run: ulimit -n 10240
-        if: ${{ contains(inputs.os, 'macos') }}
+        run: |
+          ulimit -n
```
nirinchev, line 65: Why do we do this?

.github/workflows/flutter-desktop-tests.yml
```diff
@@ -42,26 +38,33 @@ jobs:
...
           submodules: false
 
+      - name: Check runner.arch
+        if: ${{ inputs.arch != runner.arch }}
+        run: false # fail if arch is not as expected
+        shell: bash
+
+      - name: Set RUNNER_OS
+        # there is no such thing as ${{ tolower(runner.os) }}, hence this abomination ¯\_(ツ)_/¯ 
+        run: echo ${{ runner.os }} | awk '{print "RUNNER_OS=" tolower($0)}' >> $GITHUB_ENV
```
nirinchev, line 48: [nit] Rather than set is as an env variable, a more idiomatic approach would be to set it as the output of the step and use it later as `steps.set-runner-os.outputs.os` - it's more verbose, but it's nicer than stashing everything in env.

nirinchev (approved):

.github/workflows/build-native.yml
nielsenko: We don't, but I find it useful to see what versions are available when looking at the output of the step

.github/workflows/dart-desktop-tests.yml
nielsenko: We don't, but I find it useful to see the current limit in the output of the step before setting it.

PR #1554, opened 2024-03-14
RDART-941: Clear collections in mixed when reassigned

nirinchev:
Works except when reassigning collections to the same managed value, which depends on https://github.com/realm/realm-core/issues/7422.

### DIFF
--- a/packages/realm_dart/lib/src/native/realm_core.dart
+++ b/packages/realm_dart/lib/src/native/realm_core.dart
@@ -3021,6 +3021,10 @@ class _RealmCore {
           collectionHandle = listHandle;
 
           final list = realm.createList<RealmValue>(listHandle, null);
+
+          // Necessary since Core will not clear the collection if the value was already a collection
+          list.clear();
+
           for (final item in value.value as List<RealmValue>) {
             list.add(item);
           }
@@ -3030,6 +3034,10 @@ class _RealmCore {
           collectionHandle = mapHandle;
 
           final map = realm.createMap<RealmValue>(mapHandle, null);
+
+          // Necessary since Core will not clear the collection if the value was already a collection
+          map.clear();
+
           for (final kvp in (value.value as Map<String, RealmValue>).entries) {
             map[kvp.key] = kvp.value;
           }
--- a/packages/realm_dart/test/realm_value_test.dart
+++ b/packages/realm_dart/test/realm_value_test.dart
@@ -509,6 +509,8 @@ void main() {
         expect(obj.oneAny.type, RealmValueType.list);
         expectMatches(obj.oneAny, [true, 5.3]);
 
+        writeIfNecessary(realm, () => obj.oneAny = RealmValue.from(['foo']));
+
         writeIfNecessary(realm, () => obj.oneAny = RealmValue.from(999));
         expectMatches(obj.oneAny, 999);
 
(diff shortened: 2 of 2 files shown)


nielsenko (commented):
Will core change their behaviour, and hence this is something we need to get back to, or is this by design on cores part?

packages/realm_dart/lib/src/native/realm_core.dart
```diff
@@ -3030,6 +3034,10 @@ class _RealmCore {
           collectionHandle = mapHandle;
 
           final map = realm.createMap<RealmValue>(mapHandle, null);
+
+          // Necessary since Core will not clear the collection if the value was already a collection
```
nielsenko, line 3038: I wonder if it would be better to track what keys are not updated, and the remove only those, instead of removing all followed by setting them again.

I believe those to operations look quite different in the oplog.

Something similar holds for list, ie. only clear the elements in the old-list from length og the new-list (if any), and update the rest.

elle-j: I guess it also depends on what operations are expected when you e.g. set a field to a new collection. In this case, it may be preferable/expected to have the items cleared, then set new ones, rather than removing some items and updating some items.

nielsenko: It is definitely a choice, but it would avoid https://github.com/realm/realm-core/issues/7422.

elle-j: I don't think we want to avoid that though (that issue) as it could make it easier for the SDKs, and the SDKs wouldn't have to call into Core for every item insert. It's definitely an interesting discussion to have, also for Core if we do delegate to them.

nielsenko: Eventually the behavior should move to core, as to avoid the repeated calls. I'm more interested in what behavior we want. 

Even if we go with _clear, then assign_, I think this should be handled by core, not the individual SDKs.

Also, to be consistent we should allow setting all collections, not just those in a mixed.

packages/realm_dart/test/realm_value_test.dart
```diff
@@ -566,13 +568,186 @@ void main() {
...
+          {'foo': 'bar'},
+          5.3
+        ]));
+        if (isManaged) {
+          realm.write(() => realm.add(obj));
+        }
+
+        final list = obj.oneAny.asList();
+
+        writeIfNecessary(realm, () => list[1] = RealmValue.from({'new': 5}));
```
elle-j, line 595: What do you think about asserting/expecting that `list[1]` is a Realm map, right before reassigning to a new map. That would guard for potential manual changes to `oneAny` list at the top of the test and ensure that it's testing reassigning from one map to another.

elle-j (approved):
LGTM, nice test updates 🙂

nielsenko (approved):

packages/realm_dart/lib/src/native/realm_core.dart
elle-j: Thanks for the clarification! Regarding the last example, for the SDKs having the `Mixed` wrapped (e.g. the `RealmValue` class), I don't think the expectation necessarily is that `myList.add(2);` also should be acting on `myPodo.mixed` due to the intermediate API that they'd have to go through `RealmValue.from()`.

In this case, I'd say that a user's expectations may instead be based on what the API docs of `RealmValue.from()` says that it returns (e.g. the same reference as `myList` or something else (the latter in this case)).

I think the ambiguity of what users would expect becomes more ambiguous for the JS SDK where the syntax is not differentiated as we don't wrap the `Mixed` value.

PR #1565, opened 2024-03-14
RDART-975: Update core to v14.2.0-12-g95c6efce8

nielsenko:
Fix: #1568 

- **Update realm-core to v14.2.0-12-g95c6efce8**
- **Update Changelog**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -108,19 +108,18 @@
 * Added `RealmObjectBase.objectSchema` that returns the schema for this object. In most cases, this would be the schema defined in the model, but in case the Realm is opened as dynamic (by providing an empty collection for schemaObjects in the config) or using `FlexibleSyncConfiguration`, it may change as the schema on disk changes. (Issue [#1449](https://github.com/realm/realm-dart/issues/1449))
 * Added `Realm.schemaChanges` that returns a stream of schema changes that can be listened to. Only dynamic and synchronized Realms will emit schema changes. (Issue [#1449](https://github.com/realm/realm-dart/issues/1449))
 
-* Improve performance of object notifiers with complex schemas and very simple changes to process by as much as 20% ([PR #7424](https://github.com/realm/realm-core/pull/7424)).
-* Improve performance with very large number of notifiers as much as 75% ([PR #7424](https://github.com/realm/realm-core/pull/7424)).
-* Add support to synchronize collections embedded in Mixed properties and other collections (except sets) ([PR #7353](https://github.com/realm/realm-core/pull/7353) Core v14.2.0-12-g95c6efce8).
-* Improve performance of change notifications on nested collections somewhat ([PR #7402](https://github.com/realm/realm-core/pull/7402) Core v14.2.0-12-g95c6efce8).
-* Improve performance of aggregate operations on Dictionaries of objects, particularly when the dictionaries are empty ([PR #7418](https://github.com/realm/realm-core/pull/7418) Core v14.2.0-12-g95c6efce8)
+* Improve performance of object notifiers with complex schemas and very simple changes to process by as much as 20% ([Core 14.2.0).
+* Improve performance with very large number of notifiers as much as 75% (Core 14.2.0).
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -103,19 +103,26 @@
   ```
 * Added `RealmValueType` enum that contains all the possible types that can be wrapped by a `RealmValue`. (PR [#1469](https://github.com/realm/realm-dart/pull/1469))
 * Added support for accessing `Set` and `Map` types using the dynamic object API - `obj.dynamic.getSet/getMap`. (PR [#1533](https://github.com/realm/realm-dart/pull/1533))
-
+* Improve performance of object notifiers with complex schemas and very simple changes to process by as much as 20% ([PR #7424](https://github.com/realm/realm-core/pull/7424)).
```
nirinchev, line 106: I think for those we'd probably want to just have `Core v14.2.0` or similar, since it's not super easy to determine whether the PR link is going to be to the dart repo or the core one.

```suggestion
* Improve performance of object notifiers with complex schemas and very simple changes to process by as much as 20% (Core v14.2.0).
```

CHANGELOG.md
```diff
@@ -103,19 +103,26 @@
...
+* Improve performance with very large number of notifiers as much as 75% ([PR #7424](https://github.com/realm/realm-core/pull/7424)).
+* Add support to synchronize collections embedded in Mixed properties and other collections (except sets) ([PR #7353](https://github.com/realm/realm-core/pull/7353) Core v14.2.0-12-g95c6efce8).
```
nirinchev, line 108: We likely want to update to 14.3.0 before we release publicly, so these changelog entries should ideally refer to the tagged version of Core.

CHANGELOG.md
```diff
@@ -103,19 +103,26 @@
...
+* Improve performance of change notifications on nested collections somewhat ([PR #7402](https://github.com/realm/realm-core/pull/7402) Core v14.2.0-12-g95c6efce8).
```
nirinchev, line 109: We haven't released nested collections, so there's no reason to list this improvement.

CHANGELOG.md
```diff
@@ -103,19 +103,26 @@
...
 
 ### Fixed
 * If you have more than 8388606 links pointing to one specific object, the program will crash. (Core 14.0.0)
 * A Realm generated on a non-apple ARM 64 device and copied to another platform (and vice-versa) were non-portable due to a sorting order difference. This impacts strings or binaries that have their first difference at a non-ascii character. These items may not be found in a set, or in an indexed column if the strings had a long common prefix (> 200 characters). (Core 14.0.0)
 * Ctor arguments appear in random order on generated classes, if the realm model contains many properties. (PR [#1531](https://github.com/realm/realm-dart/pull/1531))
+* Fixed an issue where removing realm objects from a List with more than 1000 items could crash. ([#7414](https://github.com/realm/realm-core/pull/7414), Core 14.2.0)
+* Fixed conflict resolution bug which may result in an crash when the AddInteger instruction on Mixed properties is merged against updates to a non-integer type ([PR #7353](https://github.com/realm/realm-core/pull/7353), Core v14.2.0-12-g95c6efce8).
```
nirinchev, line 117: We don't have RealmInteger support in dart, so this is likely not relevant.

CHANGELOG.md
nielsenko: I was thinking we could update then refs then. I'm a little hesitant to refer to 14.3 when it is not released yet, and 14.2 does not contain the fix.

CHANGELOG.md
nielsenko: I'm not a big fan of dropping information like this. I know we did in the past, and I'll update to match in this PR, but I think we should reconsider. 

These are links, easy to follow, if you want more info, and it is actually a bit less editing to leave them in.

CHANGELOG.md

CHANGELOG.md

nielsenko:
Included some cleanup of .gitignore and .pubignore I spotted along the way

CHANGELOG.md
nirinchev: Fair enough, we just need to remember to go through those and fix them.

nirinchev (approved):

PR #1574, opened 2024-03-15
RDART-977: Hierarchical logger

nielsenko:
- **Update realm-core**
- **Implement and test new logger**

Fixes: #1578

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,7 @@
 
   Also, note that setting a level is no longer local to the current isolate, but shared accross all isolates. At the core level there is just one process wide logger.
 
-  Categories form a hirarchy and setting the log level of a parent category will override the level of the current children. The hierarchy is exposed in a type safe manner with:
+  Categories form a hierarchy and setting the log level of a parent category will override the level of its children. The hierarchy is exposed in a type safe manner with:
   ```dart
   sealed class RealmLogCategory {
     /// All possible log categories.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,7 +71,7 @@
     realm.sync.server,
     ...
   ```
-  the `onRecord` stream now pumps `RealmLogRecord`s that includes the category the message was logged to.
+  The `onRecord` stream now pumps `RealmLogRecord`s that include the category the message was logged to.
 
   If you want to hook up realm logging with conventional dart logging you can do:
   ```dart
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,40 @@
   ```
 * Removed `SchemaObject.properties` - instead, `SchemaObject` is now an iterable collection of `Property`. (Issue [#1449](https://github.com/realm/realm-dart/issues/1449))
 * `SyncProgress.transferredBytes` and `SyncProgress.transferableBytes` have been consolidated into `SyncProgress.progressEstimate`. The values reported previously were incorrect and did not accurately represent bytes either. The new field better conveys the uncertainty around the progress being reported. With this release, we're reporting accurate estimates for upload progress, but estimating downloads is still unreliable. A future server and SDK release will add better estimations for download progress. (Issue [#1562](https://github.com/realm/realm-dart/issues/1562))
+* `Realm.logger` is no longer settable, and no longer implements `Logger` from package `logging`. In particular you can no longer call `Realm.logger.level =`. Instead you should call `Realm.logger.setLogLevel(RealmLogLevel level, {RealmLogCategory? category})` that takes an optional category. If no category is exlicitly given, then `RealmLogCategory.realm` is assumed.
+
+  Also, note that setting a level is no longer local to the current isolate, but shared between accross all isolates. Maintaining the illution that there exists a logger per isolate, became untennable with the new API. At the core level there is just one process wide logger.
+
+  Categories form a hirarchy and setting the log level of a parent category will override the level of the current children. The hierarchy is exposed in a type safe manner with:
+  ```dart
+  sealed class RealmLogCategory {
+    /// All possible log categories.
+    static final values = [
+    realm,
+    realm.app,
+    realm.sdk,
+    realm.storage,
+    realm.storage.notification,
+    realm.storage.object,
+    realm.storage.query,
+    realm.storage.transaction,
+    realm.sync,
+    realm.sync.client,
+    realm.sync.client.changeset,
+    realm.sync.client.network,
+    realm.sync.client.reset,
+    realm.sync.client.session,
+    realm.sync.server,
+    ...
+  ```
+  the `onRecord` stream now pumps `RealmLogRecord`s that includes the category the message was logged to.
+
+  If you want to hook up realm logging with conventional dart logging you can do:
+  ```dart
+  Realm.logger.onRecord.forEach((r) => Logger(r.category.toString()).log(r.level.level, r.message));
+  ```
+  If no isolate subscribes to `Realm.logger.onRecord` then the logs will by default be send to stdout.
+
 
 ### Enhancements
 * Realm objects can now be serialized as [EJSON](https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@
 * `SyncProgress.transferredBytes` and `SyncProgress.transferableBytes` have been consolidated into `SyncProgress.progressEstimate`. The values reported previously were incorrect and did not accurately represent bytes either. The new field better conveys the uncertainty around the progress being reported. With this release, we're reporting accurate estimates for upload progress, but estimating downloads is still unreliable. A future server and SDK release will add better estimations for download progress. (Issue [#1562](https://github.com/realm/realm-dart/issues/1562))
 * `Realm.logger` is no longer settable, and no longer implements `Logger` from package `logging`. In particular you can no longer call `Realm.logger.level =`. Instead you should call `Realm.logger.setLogLevel(RealmLogLevel level, {RealmLogCategory? category})` that takes an optional category. If no category is exlicitly given, then `RealmLogCategory.realm` is assumed.
 
-  Also, note that setting a level is no longer local to the current isolate, but shared between accross all isolates. Maintaining the illution that there exists a logger per isolate, became untennable with the new API. At the core level there is just one process wide logger.
+  Also, note that setting a level is no longer local to the current isolate, but shared accross all isolates. At the core level there is just one process wide logger.
 
   Categories form a hirarchy and setting the log level of a parent category will override the level of the current children. The hierarchy is exposed in a type safe manner with:
   ```dart
@@ -77,7 +77,7 @@
   ```dart
   Realm.logger.onRecord.forEach((r) => Logger(r.category.toString()).log(r.level.level, r.message));
   ```
-  If no isolate subscribes to `Realm.logger.onRecord` then the logs will by default be send to stdout.
+  If no isolate subscribes to `Realm.logger.onRecord` then the logs will by default be sent to stdout.
 
 
 ### Enhancements
--- a/packages/realm_dart/lib/src/logging.dart
+++ b/packages/realm_dart/lib/src/logging.dart
@@ -8,7 +8,7 @@ import 'package:logging/logging.dart';
 import 'native/realm_core.dart';
 
 // Using classes to make a fancy hierarchical enum
-sealed class RealmLogCategory {
+sealed class LogCategory {
   /// All possible log categories.
   static final values = [
     realm,
@@ -32,12 +32,12 @@ sealed class RealmLogCategory {
   static final realm = _RealmLogCategory();
 
   final String _name;
-  final RealmLogCategory? _parent;
+  final LogCategory? _parent;
 
-  RealmLogCategory._(this._name, this._parent);
+  LogCategory._(this._name, this._parent);
 
   /// Returns `true` if this category contains the given [category].
-  bool _contains(RealmLogCategory category) {
+  bool _contains(LogCategory category) {
     var current = category;
     while (current != this) {
       final isRoot = current == realm;
@@ -54,18 +54,18 @@ sealed class RealmLogCategory {
   @override
   String toString() => _toString;
 
-  static final _map = {for (final category in RealmLogCategory.values) category.toString(): category};
+  static final _map = {for (final category in LogCategory.values) category.toString(): category};
 
-  /// Returns the [RealmLogCategory] for the given [category] string.
+  /// Returns the [LogCategory] for the given [category] string.
   /// Will throw if the category is not recognized.
-  factory RealmLogCategory.fromString(String category) => _map[category]!;
+  factory LogCategory.fromString(String category) => _map[category]!;
 }
 
-class _LeafLogCategory extends RealmLogCategory {
-  _LeafLogCategory(super.name, RealmLogCategory super.parent) : super._();
+class _LeafLogCategory extends LogCategory {
+  _LeafLogCategory(super.name, LogCategory super.parent) : super._();
 }
 
-class _RealmLogCategory extends RealmLogCategory {
+class _RealmLogCategory extends LogCategory {
   _RealmLogCategory() : super._('Realm', null);
 
   late final app = _LeafLogCategory('App', this);
@@ -74,15 +74,15 @@ class _RealmLogCategory extends RealmLogCategory {
   late final sync = _SyncLogCategory(this);
 }
 
-class _SyncLogCategory extends RealmLogCategory {
-  _SyncLogCategory(RealmLogCategory parent) : super._('Sync', parent);
+class _SyncLogCategory extends LogCategory {
+  _SyncLogCategory(LogCategory parent) : super._('Sync', parent);
 
   late final client = _ClientLogCategory(this);
   late final server = _LeafLogCategory('Server', this);
 }
 
-class _ClientLogCategory extends RealmLogCategory {
-  _ClientLogCategory(RealmLogCategory parent) : super._('Client', parent);
+class _ClientLogCategory extends LogCategory {
+  _ClientLogCategory(LogCategory parent) : super._('Client', parent);
 
   late final changeset = _LeafLogCategory('Changeset', this);
   late final network = _LeafLogCategory('Network', this);
@@ -90,8 +90,8 @@ class _ClientLogCategory extends RealmLogCategory {
   late final session = _LeafLogCategory('Session', this);
 }
 
-class _StorageLogCategory extends RealmLogCategory {
-  _StorageLogCategory(RealmLogCategory parent) : super._('Storage', parent);
+class _StorageLogCategory extends LogCategory {
+  _StorageLogCategory(LogCategory parent) : super._('Storage', parent);
 
   late final notification = _LeafLogCategory('Notification', this);
   late final object = _LeafLogCategory('Object', this);
@@ -102,7 +102,7 @@ class _StorageLogCategory extends RealmLogCategory {
 /// Specifies the criticality level above which messages will be logged
 /// by the default sync client logger.
 /// {@category Realm}
-enum RealmLogLevel {
+enum LogLevel {
   /// Log everything. This will seriously harm the performance of the
   /// sync client and should never be used in production scenarios.
   all(Level.ALL),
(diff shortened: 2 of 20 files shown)


packages/realm_dart/lib/src/logging.dart
```diff
@@ -0,0 +1,182 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+import 'dart:async';
+
+import 'package:logging/logging.dart';
+
+import 'native/realm_core.dart';
+
+// Using classes to make a fancy hierarchical enum
+sealed class RealmLogCategory {
```
nirinchev, line 11: Can we use `LogCategory` here? I know we do sometimes prefix stuff with `Realm`, but I'd prefer if we only do that for things that go in a Realm (database) and not for other API.

packages/realm_dart/lib/src/logging.dart
```diff
@@ -0,0 +1,182 @@
...
+  error(Level.SEVERE),
+
+  /// Log only fatal errors.
+  fatal(Level.SHOUT),
+
+  /// Turn off logging.
+  off(Level.OFF),
+  ;
+
+  static const logToValues = [
```
nirinchev, line 133: Should this be public?

packages/realm_dart/lib/src/logging.dart
```diff
@@ -0,0 +1,182 @@
...
+    RealmLogLevel.debug,
+    RealmLogLevel.detail,
+    RealmLogLevel.info,
+    RealmLogLevel.warn,
+    RealmLogLevel.error,
+    RealmLogLevel.fatal,
+  ];
+
+  final Level level;
+  const RealmLogLevel(this.level);
```
nirinchev, line 144: Public?

packages/realm_dart/lib/src/logging.dart
```diff
@@ -0,0 +1,182 @@
...
+    RealmLogLevel.trace,
+    RealmLogLevel.debug,
+    RealmLogLevel.detail,
+    RealmLogLevel.info,
+    RealmLogLevel.warn,
+    RealmLogLevel.error,
+    RealmLogLevel.fatal,
+  ];
+
+  final Level level;
```
nirinchev, line 143: Missing docs for this one

packages/realm_dart/lib/src/logging.dart
```diff
@@ -0,0 +1,182 @@
...
+  const RealmLogLevel(this.level);
+
+  static final _valuesMap = {for (var value in RealmLogLevel.values) value.index: value};
+  factory RealmLogLevel.fromInt(int code) => _valuesMap[code]!;
```
nirinchev, line 147: Similarly, should this be private?

packages/realm_dart/lib/src/logging.dart
```diff
@@ -0,0 +1,182 @@
...
+}
+
+/// A record of a log message from the Realm SDK.
+typedef RealmLogRecord = ({RealmLogCategory category, RealmLogLevel level, String message});
```
nirinchev, line 151: Similarly to `RealmLogCategory`, can we make this a `LogRecord` or will it conflict with built-in types?

packages/realm_dart/lib/src/logging.dart
```diff
@@ -0,0 +1,182 @@
...
+  );
+
+  const RealmLogger();
+
+  void setLogLevel(RealmLogLevel level, {RealmLogCategory? category}) {
+    category ??= RealmLogCategory.realm;
+    realmCore.setLogLevel(level, category: category);
+  }
+
+  Stream<RealmLogRecord> get onRecord => _controller.stream;
```
nirinchev, line 167: Docs

packages/realm_dart/lib/src/logging.dart
```diff
@@ -0,0 +1,182 @@
...
+/// A logger that logs messages from the Realm SDK.
+class RealmLogger {
+  static final _controller = StreamController<RealmLogRecord>.broadcast(
+    onListen: () => realmCore.loggerAttach(),
+    onCancel: () => realmCore.loggerDetach(),
+  );
+
+  const RealmLogger();
+
+  void setLogLevel(RealmLogLevel level, {RealmLogCategory? category}) {
```
nirinchev, line 162: Docs

packages/realm_dart/lib/src/realm_class.dart
```diff
@@ -510,26 +510,9 @@ class Realm implements Finalizable {
     return realmCore.realmEquals(this, other);
   }
 
-  static Logger _logger = realmCore.defaultRealmLogger;
-
-  /// The logger to use for Realm logging in this `Isolate`
+  /// The logger to use for Realm logging in all [Isolate]s
```
nirinchev, line 513: I think the `Isolate` bit is somewhat unclear here - not sure why it's relevant. I think we could remove that and instead add info on how to subscribe for messages instead. Something like:

```suggestion
  /// The logger that will emit log messages from the database and sync operations. To receive log
  /// messages, use `logger.onRecord.forEach` and pass in a callback to be invoked for every item.
```

packages/realm_dart/lib/src/logging.dart
```diff
@@ -0,0 +1,182 @@
...
+
+  static final _valuesMap = {for (var value in RealmLogLevel.values) value.index: value};
+  factory RealmLogLevel.fromInt(int code) => _valuesMap[code]!;
+}
+
+/// A record of a log message from the Realm SDK.
+typedef RealmLogRecord = ({RealmLogCategory category, RealmLogLevel level, String message});
+
+/// A logger that logs messages from the Realm SDK.
+class RealmLogger {
```
nirinchev, line 154: The docs could use a code snippet for how you setup listening.

packages/realm_dart/lib/src/logging.dart
```diff
@@ -0,0 +1,182 @@
...
+    return true;
+  }
+
+  // cache the toString result
+  late final String _toString = (_parent == null ? _name : '$_parent.$_name');
+  @override
+  String toString() => _toString;
+
+  static final _map = {for (final category in RealmLogCategory.values) category.toString(): category};
+  factory RealmLogCategory.fromString(String message) => _map[message]!;
```
nirinchev, line 56: Should this be public? Also, should we rename the arg to `category`?

packages/realm_dart/lib/src/scheduler.dart
```diff
@@ -22,9 +24,11 @@ class Scheduler {
     _receivePort.handler = (dynamic message) {
       if (message is List) {
         // currently the only `message as List` is from the logger.
-        final level = message[0] as int;
-        final text = message[1] as String;
-        Realm.logger.log(LevelExt.fromInt(level), text);
+        final category = RealmLogCategory.fromString(message[0] as String);
+        // category is a string to avoid converting back and forth between RealmLogCategory and String
```
nirinchev, line 28: This comment seems incorrect/unclear since `category` is in fact `RealmLogCategory`.

packages/realm_dart/test/app_test.dart
```diff
@@ -5,6 +5,7 @@ import 'dart:convert';
 import 'dart:io';
 import 'dart:isolate';
 import 'package:logging/logging.dart';
+import 'package:realm_dart/src/logging.dart';
```
nirinchev, line 8: This is suspicious. I would expect everything relevant to have been exported by `realm_dart`.

packages/realm_dart/test/realm_logger_test.dart
```diff
@@ -1,211 +1,105 @@
...
-        // turn on second isolate
-        Realm.logger.level = RealmLogLevel.trace;
+  test('RealmLogger hookup logging', () async {
+    Realm.logger.setLogLevel(RealmLogLevel.off);
+    Realm.logger.setLogLevel(RealmLogLevel.all, category: RealmLogCategory.realm.sdk);
 
-        // log a another message
-        await logTestMessage(RealmLogLevel.trace, "3");
+    final logger = Logger.detached('Test');
+    Realm.logger.onRecord.forEach((r) => logger.log(r.level.level, r.message));
```
nirinchev, line 83: Do we need to clean this up?

packages/realm_dart/test/realm_logger_test.dart
```diff
@@ -1,211 +1,105 @@
...
+    Logger.root.level = Level.ALL;
 
-    final expected = [
-      const LoggedMessage(RealmLogLevel.error, "2"),
-      const LoggedMessage(RealmLogLevel.error, "2"),
-      const LoggedMessage(RealmLogLevel.error, "first only"),
-      const LoggedMessage(RealmLogLevel.trace, "3")
-    ];
+    expectLater(Logger('Realm').onRecord, emits(isA<LogRecord>().having((r) => r.level, 'level', Level.SEVERE).having((r) => r.message, 'message', 'error')));
+    expectLater(
```
nirinchev, line 100: I'm not sure what's happening here - is the hierarchical logging something we get for free from the framework?

nirinchev (changes requested):
Looks good - I have some questions as well as a few docs reminders. I only glossed over the C++ code and it seems reasonable to me, but might want to get another pair of eyes on it (@elle-j perhaps?).

I believe the following tests should work, but it might be a good idea to verify they do:

1. Log level can be updated dynamically:
  a. Subscribe at trace level
  b. Log at trace level; verify log is received
  c. Set log level to info
  d. Log at trace level; verify log is not received
2. onRecord broadcast to multiple subscribers
  a. `final sub1 = onRecord.listen(...);`
  b. `final sub2 = onRecord.listen(...);`
  c. Emit message, verify both callbacks are invoked
  d. Cancel sub1
  e. Emit message, verify only sub2 callback is invoked

CHANGELOG.md
```diff
@@ -45,6 +45,40 @@
...
+* `Realm.logger` is no longer settable, and no longer implements `Logger` from package `logging`. In particular you can no longer call `Realm.logger.level =`. Instead you should call `Realm.logger.setLogLevel(RealmLogLevel level, {RealmLogCategory? category})` that takes an optional category. If no category is exlicitly given, then `RealmLogCategory.realm` is assumed.
+
+  Also, note that setting a level is no longer local to the current isolate, but shared between accross all isolates. Maintaining the illution that there exists a logger per isolate, became untennable with the new API. At the core level there is just one process wide logger.
```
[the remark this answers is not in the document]
nielsenko: I'll drop the "Maintaining the illusion.. " part

packages/realm_dart/lib/src/logging.dart
```diff
@@ -0,0 +1,182 @@
...
+
+  /// The root category for all log messages from the Realm SDK.
+  static final realm = _RealmLogCategory();
+
+  final String _name;
+  final RealmLogCategory? _parent;
+
+  RealmLogCategory._(this._name, this._parent);
+
+  bool contains(RealmLogCategory category) {
```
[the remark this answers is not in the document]
nielsenko: If not, it should simply be removed. It is not used by the implementation.

packages/realm_dart/lib/src/logging.dart
nielsenko: We could. 

I'm am/was little hesitant to call `RealmLogRecord` for `LogRecord` as that is the same name the [logging](https://pub.dev/packages/logging) package use, and I foresee that people will want to connect with that package.

And once I started using a `Realm` prefix on that it felt more natural to use it throughout the interface - including here.

Obviously any name conflict can always be resolved with a simple `import x show y as z`, but since we are using the `Realm` prefix so many other places I added it for consistency.

So, it was a conscious choice, but I don't have strong opinions. Now you got the context, I'll let you do the call 😄

packages/realm_dart/lib/src/logging.dart
nielsenko: If not, it should. simply be removed, as it is only used by the tests.

packages/realm_dart/lib/src/logging.dart
nielsenko: Will fix

nielsenko: This is an enum.. it is not like you can go construct one on the fly.

packages/realm_dart/lib/src/logging.dart
nielsenko: See discussion about `RealmLogCategory`.

packages/realm_dart/lib/src/realm_class.dart
nielsenko: Okay, I'll reword a little. Don't think we should mention `forEach`, just the `Realm.logger.onRecord` stream.

packages/realm_dart/lib/src/scheduler.dart
nielsenko: old cruft, will fix.

packages/realm_dart/lib/src/logging.dart
nielsenko: Removed altogether. You can always construct an enum with `Enum.values[index]` anyway. We only use this pattern if the index and the integer code don't match up.

packages/realm_dart/lib/src/logging.dart
nielsenko: I'll rename argument, but I think we should keep it public.

packages/realm_dart/test/realm_logger_test.dart
nielsenko: ?

CHANGELOG.md
```diff
@@ -45,6 +45,40 @@
...
+* `Realm.logger` is no longer settable, and no longer implements `Logger` from package `logging`. In particular you can no longer call `Realm.logger.level =`. Instead you should call `Realm.logger.setLogLevel(RealmLogLevel level, {RealmLogCategory? category})` that takes an optional category. If no category is exlicitly given, then `RealmLogCategory.realm` is assumed.
+
+  Also, note that setting a level is no longer local to the current isolate, but shared accross all isolates. At the core level there is just one process wide logger.
+
+  Categories form a hirarchy and setting the log level of a parent category will override the level of the current children. The hierarchy is exposed in a type safe manner with:
```
elle-j, line 52:
```suggestion
  Categories form a hierarchy and setting the log level of a parent category will override the level of its children. The hierarchy is exposed in a type safe manner with:
```

CHANGELOG.md
```diff
@@ -45,6 +45,40 @@
...
+    realm.sync,
+    realm.sync.client,
+    realm.sync.client.changeset,
+    realm.sync.client.network,
+    realm.sync.client.reset,
+    realm.sync.client.session,
+    realm.sync.server,
+    ...
+  ```
+  the `onRecord` stream now pumps `RealmLogRecord`s that includes the category the message was logged to.
```
elle-j, line 74:
```suggestion
  The `onRecord` stream now pumps `RealmLogRecord`s that include the category the message was logged to.
```

packages/realm_dart/lib/src/logging.dart
```diff
@@ -0,0 +1,178 @@
...
+  final RealmLogCategory? _parent;
+
+  RealmLogCategory._(this._name, this._parent);
+
+  /// Returns `true` if this category contains the given [category].
+  bool contains(RealmLogCategory category) {
+    var current = category;
+    while (current != this) {
+      if (current == realm) {
+        // root
```
elle-j, line 44: Alternative:
```suggestion
      final isRoot = current == realm;
      if (isRoot) {
```

packages/realm_dart/lib/src/logging.dart
```diff
@@ -0,0 +1,178 @@
...
+    realmCore.setLogLevel(level, category: category);
+  }
+
+  Stream<RealmLogRecord> get onRecord => _controller.stream;
+
+  void _raise(RealmLogRecord record) {
+    _controller.add(record);
+  }
+
+  void _log(RealmLogLevel level, Object message, {RealmLogCategory? category}) {
```
elle-j, line 169: In what other types would we expect the `message` to be passed as other than `String`?

packages/realm_dart/test/realm_logger_test.dart
```diff
@@ -1,211 +1,115 @@
...
-        });
-
-        RealmInternal.logMessageForTesting(level, "123");
-
-        return await completer.future;
+  group('All levels', () {
+    for (var level in RealmLogLevel.values) {
+      test('Realm.logger supports log level $level', () {
+        Realm.logger.setLogLevel(RealmLogLevel.off);
+        Realm.logger.setLogLevel(RealmLogLevel.all, category: RealmLogCategory.realm.sdk);
```
elle-j, line 29: Why is the `off` necessary here?

packages/realm_dart/test/realm_logger_test.dart
```diff
@@ -1,211 +1,115 @@
...
-        Realm.logger.level = RealmLogLevel.off;
-
-        // log another message. second isoalte should not get it
-        await logTestMessage(RealmLogLevel.error, "first only");
+    final trace = Isolate.run(() async {
+      return (await Realm.logger.onRecord.first).message;
+    });
+    await Future<void>.delayed(const Duration(milliseconds: 100)); // yield
+    expectLater(trace, completion('Hey'));
+    Realm.logger.log(RealmLogLevel.trace, 'Hey');
```
elle-j, line 85: Do you mind sharing what the issue and solution was with this piece of code that we discussed earlier in the office?

elle-j (approved):
Interesting seeing how this is now done in Dart! LGTM 🙂

packages/realm_dart/test/realm_logger_test.dart
nielsenko: It suppress all other trace than what is send to `RealmLogCategory.realm.sdk` ..like sync stuff in background.

packages/realm_dart/lib/src/logging.dart
nielsenko: Dart's default logger package allows objects to be passed for message. This just follows that convention. Previously I considered keeping log method public, since that would be less of a breaking change. However, I decided against that.

packages/realm_dart/test/realm_logger_test.dart
nielsenko: A native side bug in what became bd49967b.

packages/realm_dart/lib/src/logging.dart
elle-j: Ah okay, thanks 👍

packages/realm_dart/lib/src/logging.dart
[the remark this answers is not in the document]
nirinchev: Yeah, I don't see immediate value of it, so we could probably leave it out for now and add it or something like it if we get feedback from users.

packages/realm_dart/test/realm_logger_test.dart
elle-j: Great digging, might bug you about some of those changes at a later time once the release is out 👍

packages/realm_dart/lib/src/logging.dart
nirinchev: I see - I don't have a strong opinion there, it's just a little annoying that we have to use a worse API for our stuff to avoid conflicting with another package, but I guess that's the price we pay for not being first-party.

packages/realm_dart/test/realm_logger_test.dart
nirinchev: Sorry, was late and wasn't super clear. I meant that when you do `Realm.logger.onRecord.listen` you get a subscription you can then cancel. With `forEach` you don't seem to get that, so wouldn't that code continue executing even after this particular test completes.

nielsenko: 😬 eagle eye..

nirinchev:
> This is tested in the "Match levels" test.

I think this is testing what I suggested by accident between test runs, rather than more explicitly. In the `match levels` test, we're only setting the log level once and then verifying only the correct categories are routed to the callback, but it doesn't validate that you can change the level dynamically - the validation only happens because we then run the test with a different log level.

> This is basic broadcast stream functionality. We are just using a StreamController<T>.broadcast here.

I agree that such a test would pass, but the usage of `StreamController<T>.broadcast` is not part of the public API and as such is subject to change at any point. A year in the future, someone might decide that `.broadcast` is inefficient and refactor that and with no tests to verify the desired behavior is preserved, we may end up missing the regression.

nirinchev:
Another test which we could add (or I may have missed) is validating that we have mapped all the native category names to managed `RealmLogCategory`. You can use `realm_get_category_names` to get the names Core uses and compare them to `RealmLogCategory.values`.

packages/realm_dart/test/realm_logger_test.dart
nielsenko: Fixed

nielsenko:
> Another test which we could add (or I may have missed) is validating that we have mapped all the native category names to managed `RealmLogCategory`...

Added, and vice-versa.

nielsenko:
Requested tests added

nirinchev (approved):

PR #1576, opened 2024-03-18
RDART-976: Fix install command

nirinchev:
No changelog since we're not keeping dedicated entries for alpha releases.

### DIFF
--- a/packages/realm_dart/lib/src/cli/install/install_command.dart
+++ b/packages/realm_dart/lib/src/cli/install/install_command.dart
@@ -31,8 +31,7 @@ class InstallCommand extends Command<void> {
     populateOptionsParser(argParser);
   }
 
-  Directory getBinaryPath(Package realmPackage, {required bool isFlutter}) {
-    final realmPackagePath = realmPackage.root.toFilePath();
+  Directory getBinaryPath(String realmPackagePath, {required bool isFlutter}) {
     if (isFlutter) {
       return Directory(switch (options.targetOsType) {
         TargetOsType.android => path.join(realmPackagePath, 'android', 'src', 'main', 'cpp', 'lib'),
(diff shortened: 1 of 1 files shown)


packages/realm_dart/lib/src/cli/install/install_command.dart
```diff
@@ -100,7 +99,7 @@ class InstallCommand extends Command<void> {
     await versionFile.writeAsString(version.toString());
   }
 
-  Future<Package> getPackage(String name) async {
```
nielsenko, line 103: Why do you prefer to traffic in typeless `String`s over `Package`?

nirinchev: Because we don't use anything from the package except for the root folder. I think structuring the code that way makes it more explicit - it doesn't force the reader to figure out what the Package API is and/or wonder what exactly it is being used for.

nielsenko: Could we use `Directory`/`File`?

nirinchev: We could - just don't see much value in it, since `path.join` takes strings and the Directory/File API doesn't seem to do any validations either, so it'll be just wrapping the paths. I guess one benefit is that it's going to be a bit more explicit about what the return type is, but I feel this is reasonably well achieved by the method naming.

nielsenko: I guess I'm just a bit allergic to use _typeless_ values, hence why I made it a `Package` in the first place. What are you actually trying to fix with this PR. It is not immediately apparent to me.

nielsenko (approved):

PR #1581, opened 2024-03-18
RDART-978: Add support RealmValue.fromJson

nirinchev:
Fixes https://github.com/realm/realm-dart/issues/1580

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,19 +1,5 @@
 ## vNext (TBD)
 
-### Enhancements
-* None
-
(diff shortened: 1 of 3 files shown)


nielsenko (commented):
Obviously nothing technically wrong with this PR, but do we want to add this?

Typically in Dart is for `fromJson` does not take a `String` argument, but a `Map` (see https://docs.flutter.dev/data-and-backend/serialization/json#serializing-json-inside-model-classes).

nielsenko (approved):

PR #1582, opened 2024-03-18
Fix release workflow

nirinchev:
Looks like the release workflow deleted a bunch of symlinks. Reverting this as well as adding a `git reset --hard` post package publish to make sure we're in a clean state.


nielsenko (approved):
Apparently these where lost when "reverting" 4b06f7412a2f17c25d04941d80c5240d4656b903

PR #1586, opened 2024-03-19
Schema migration support

nirinchev:


packages/realm_dart/lib/src/configuration.dart
```diff
@@ -340,6 +342,9 @@ class FlexibleSyncConfiguration extends Configuration {
   /// Called when opening a `Realm` for the first time, after process start.
   final ShouldCompactCallback? shouldCompactCallback;
 
+  /// TODO
```
nirinchev, line 345: Reminder

packages/realm_dart/src/realm_dart_sync.cpp
```diff
@@ -233,8 +233,10 @@ RLM_API bool realm_dart_sync_after_reset_handler_callback(realm_userdata_t userd
 RLM_API void realm_dart_async_open_task_callback(realm_userdata_t userdata, realm_thread_safe_reference_t* realm, const realm_async_error_t* error)
 {
     auto ud = reinterpret_cast<realm_dart_userdata_async_t>(userdata);
-    ud->scheduler->invoke([ud, realm, error]() {
-        (reinterpret_cast<realm_async_open_task_completion_func_t>(ud->dart_callback))(ud->handle, realm, error);
+    realm_async_error_t* error_clone = error ? error->clone() : nullptr;
```
nirinchev, line 236: This was a pre-existing bug that I noticed - the passed in `error` pointer gets destroyed immediately after we exit this function, which means that when the scheduler executes the callback, it's no longer pointing to anything meaningful.

packages/realm_dart/test/baas_helper.dart
```diff
@@ -52,16 +54,6 @@ extension on String? {
   }
 }
 
-enum AppNames {
```
nirinchev, line 55: This was moved to `baas_client`, which is why it's re-exported by the helper.

packages/realm_dart/lib/src/cli/atlas_apps/baas_client.dart
```diff
@@ -29,14 +29,46 @@ class BaasAuthHelper {
   }
 }
 
+enum AppNames {
```
nirinchev, line 32: Moved from `baas_helper` to make sure we have a single source of truth for the names.

packages/realm_dart/test/user_test.dart
```diff
@@ -56,14 +56,14 @@ void main() {
 
   baasTest('User link credentials', (configuration) async {
     final app = App(configuration);
-    final user1 = await app.logIn(Credentials.anonymous());
```
nirinchev, line 59: This entire file is just cleanup changes with no actual functional impact.

packages/realm_dart/.gitignore
```diff
@@ -1,4 +1,3 @@
 /binary/
 /build-native/
 /mongodb-realm/
-*.realm.*
```
nielsenko, line 4: 🤦  .. this was added to exclude the actual `realm` database files, not the generated `.realm.dart` files.. things has been a bit hectic lately

packages/realm_dart/lib/src/cli/atlas_apps/baas_client.dart
```diff
@@ -29,14 +29,46 @@ class BaasAuthHelper {
...
+class JsonSchemaDefinition {
+  final String collectionName;
+  final List<JsonSchemaProperty> properties;
+
+  JsonSchemaDefinition(this.collectionName, this.properties);
+}
+
+class JsonSchemaProperty {
+  final String name;
+  final String bsonType;
```
nielsenko, line 53: Should we make an enum for these?

packages/realm_dart/src/realm_dart_sync.cpp
```diff
@@ -233,8 +233,10 @@ RLM_API bool realm_dart_sync_after_reset_handler_callback(realm_userdata_t userd
...
+    ud->scheduler->invoke([ud, realm, error_clone]() {
+        (reinterpret_cast<realm_async_open_task_completion_func_t>(ud->dart_callback))(ud->handle, realm, error_clone);
+        delete error_clone;
```
nielsenko, line 239: ✅

packages/realm_dart/test/baas_helper.dart
```diff
@@ -148,7 +140,7 @@ class BaasHelper {
   Future<void> _waitForInitialSync(BaasApp app) async {
     while (true) {
       try {
-        print('Validating initial sync is complete...');
+        print('Validating initial sync for $app is complete...');
```
nielsenko, line 143: Should we log instead?

packages/realm_dart/test/embedded_test.dart
```diff
@@ -31,8 +31,7 @@ void main() {
   });
 
   baasTest('Synchronized Realm with orphan embedded schemas throws', (configuration) async {
-    final app = App(configuration);
-    final user = await getIntegrationUser(app);
+    final user = await getIntegrationUser(appConfig: configuration);
```
nielsenko, line 34: A common theme 😄

packages/realm_dart/test/sync_migration_test.dart
```diff
@@ -0,0 +1,259 @@
+// Copyright 2022 MongoDB, Inc.
```
nielsenko, line 1: Mostly a question. Shouldn't this say 2024? Ie. the year the file was first added.

packages/realm_dart/test/sync_migration_test.dart
```diff
@@ -0,0 +1,259 @@
...
+  late String? stringValue;
+  late ObjectId? objectIdValue;
+  late Uuid? uuidValue;
+  late Uint8List? binaryValue;
+}
+
+// This has to match the server-side schema declared in baas_client::_nullablesSchemaV1
+@RealmModel()
+@MapTo("Nullables")
+class _NullablesV1 {
```
nielsenko, line 38: I find this to be a bit confusing name. Would 
`_V0SchemaWithNullables` and `_V1SchemaWithoutNullables` perhaps be better.

packages/realm_dart/test/sync_migration_test.dart
nirinchev: It should 👀

packages/realm_dart/test/baas_helper.dart
nirinchev: We could here, but we're using print statements in the rest of the baas setup code, so I felt it's fine to keep it in. Also, the way we've setup the logger is to only log on failure, while I think those one-time logs are useful anyway.

packages/realm_dart/lib/src/cli/atlas_apps/baas_client.dart
nielsenko, line 32:
```suggestion
enum AppName {
```
I don't think we should use plural on `enums`. It looks odd in argument lists and on call-sites.

packages/realm_dart/lib/src/cli/atlas_apps/baas_client.dart
```diff
@@ -243,25 +306,26 @@ class BaasClient {
...
-      await Future.delayed(Duration(seconds: 5));
+      await Future<Object?>.delayed(Duration(seconds: 5));
     }
 
     print('Initial sync for ${app.name} is complete.');
   }
 
-  Future<void> _createAppIfNotExists(List<BaasApp> existingApps, String appName, String appSuffix, {String? confirmationType}) async {
-    final existingApp = existingApps.firstWhereOrNull((a) => a.name == appName);
+  Future<void> _createAppIfNotExists(List<BaasApp> existingApps, AppNames appName) async {
```
nielsenko, line 325:
```suggestion
  Future<void> _createAppIfNotExists(List<BaasApp> existingApps, AppName appName) async {
```
Looks odd here due to plural

packages/realm_dart/lib/src/cli/atlas_apps/baas_client.dart
```diff
@@ -321,15 +391,18 @@ class BaasClient {
     await _updateFunction(app, 'confirmFunc', confirmFuncId, source ?? _confirmFuncSource);
   }
 
-  Future<BaasApp> _createApp(String name, String suffix, {String? confirmationType}) async {
-    final uniqueName = "$name$suffix";
+  Future<BaasApp> _createApp(AppNames appName) async {
+    final uniqueName = "${appName.name}$_appSuffix";
     print('Creating app $uniqueName');
 
+    final runConfirmationFunction = appName != AppNames.autoConfirm && appName != AppNames.emailConfirm;
```
nielsenko, line 398: Looks odd here due to plural
```suggestion
    final runConfirmationFunction = appName != AppName.autoConfirm && appName != AppName.emailConfirm;
```

packages/realm_dart/lib/src/native/realm_bindings.dart
```diff
@@ -3999,21 +3999,23 @@ class RealmLibrary {
     ffi.Pointer<ffi.Void> userdata,
     int transferred_bytes,
     int total_bytes,
+    double estimate,
```
nielsenko, line 4002: I know this is not part of this PR per se, but just a side effect of updating realm-core, but what is the relation between these 3 args. I assume it is not `estimate = transferred_bytes / total_bytes`.

nielsenko (approved):
I have a few comments, but nothing to prevent an approval

PR #1587, opened 2024-03-20
Update melos to 5.2.0

nielsenko:
- Use new steps feature (https://github.com/invertase/melos/pull/664)
- Drop analyze and format scripts, as these are now internal commands to melos
- Renamed format script to custom_format
- Renamed analyze:deps to lint:deps

### DIFF
--- a/melos.yaml
+++ b/melos.yaml
@@ -25,21 +25,21 @@ command:
 
 scripts:
   setup:
-    run: >-
-      dart pub global activate combine_coverage &&
-      dart pub global activate coverage &&
-      dart pub global activate coverde &&
-      dart pub global activate dependency_validator &&
-      dart pub global activate ffigen &&
-      dart pub global activate melos &&
-      dart pub global activate pana &&
-      dart pub global activate --source path $MELOS_ROOT_PATH/packages/realm_dart/dev/
+    steps:
+      - dart pub global activate combine_coverage
+      - dart pub global activate coverage
+      - dart pub global activate coverde
+      - dart pub global activate dependency_validator
+      - dart pub global activate ffigen
+      - dart pub global activate melos
+      - dart pub global activate pana
+      - dart pub global activate --source path $MELOS_ROOT_PATH/packages/realm_dart/dev/
 
   build:
-    run: >-
-      melos run build:native &&
-      melos run build:binding &&
-      melos run build:dart
+    steps:
+      - build:native
+      - build:binding
+      - build:dart
       
   build:native:
     exec: dev build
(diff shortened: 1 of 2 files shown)


spydon:
Wow, happy to see the new feature in practice, so much cleaner!

nielsenko:
Added no-changelog, as this is not a user visible feature, and we _sadly_ don't use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) (yet) 😄

melos.yaml
```diff
@@ -115,7 +115,7 @@ scripts:
     packageFilters:
       published: true
 
-  format:
+  custom_format:
```
elle-j, line 118: Was `format` conflicting with some other convention?

melos.yaml
```diff
@@ -226,12 +217,10 @@ scripts:
...
-      melos run test &&
-      melos run coverage &&
-      melos run analyze &&
-      melos run lint
+    steps:
+      - build
+      - test
+      - coverage
+      - analyze
+      - lint
```
elle-j, line 225: How come the `clean` and `bootstrap` commands are not explicitly needed in this case anymore?

elle-j (approved):
Much cleaner with `step`s 🙂

melos.yaml
nielsenko: The new steps feature is for chaining _scripts_. Currently it doesn't support _commands_ (made a comment about it [here](https://github.com/invertase/melos/issues/653#issuecomment-2008977433)). Thinking about it a bit, it was perhaps a bit much to require a `clean` anyway, so I removed that, and hence `bootstrap` is not needed either.

melos.yaml
nielsenko: Yes, as of 5.1.0 `format` is a build-in command in melos. I made an issue on it [here](https://github.com/invertase/melos/issues/661), but I guess we will at most get a warning. Anyway, the default `melos format` is a bit too simple for us (currently), hence the re-name.

melos.yaml
elle-j: Wouldn't it be necessary in some cases to clean? Or why do you think it isn't needed?

melos.yaml
elle-j: Gotcha 👍

melos.yaml
nielsenko: It should be rare, but [`melos clean`](https://melos.invertase.dev/commands/clean) basically brings us closer to a fresh start.

nielsenko: We are not currently using the melos _ci_ script on our actual CI. It is intended to go through all (/most) of the motions the github actions does, but locally.. but to be honest it is work in progress.

In some ways it does more, in some ways it does less.

elle-j: Oh yeah, thanks for the reminder

nirinchev (approved):

PR #1601, opened 2024-03-22
[RDART-968] Add support for keypath filtering on object

papafe:
This PR adds the `changeFor([List<String>? keyPaths]` method to the generated realm classes to support keypath filtering on object notifications

TODO 
- [x] Changelog
- [x] Documentation
- [x] Fix generator tests
- [x] Tests
- [x] Fix backlink tests
- [x] Fix mapped property tests


papafe:
@nirinchev @nielsenko It's still a little rough, but I thought it would be a good idea if you could give me some initial feedback, so I know I'm going in the right direction

packages/realm_dart/test/realm_object_test.dart
```diff
@@ -74,9 +74,390 @@ class _BoolValue {
...
+        tno.link?.intProperty = 23;
+      });
+
+      await verifyNotifications<TestNotificationObject>(tno, externalChanges, null);
+
+      subscription.cancel();
+    });
+
+    // TODO Need to decide what we expect here....
+    test('mappedProperty', () async {
```
papafe, line 293: This test at the moment passes, but we should decide what kind of behaviour we want to have with mapped properties. 
At the moment we need to pass the "managed" name when we subscribe, but we get the mapped name in `changes`

packages/realm_dart/lib/src/realm_object.dart
```diff
@@ -483,8 +483,17 @@ mixin RealmObjectBase on RealmEntity implements RealmObjectBaseMarker, Finalizab
...
 
+  Stream<RealmObjectChanges<RealmObjectBase>> changesFor([List<String>? keyPaths]) =>
+      throw RealmError("Invalid usage. Use the generated inheritors of RealmObject");
+
   /// @nodoc
   static Stream<RealmObjectChanges<T>> getChanges<T extends RealmObjectBase>(T object) {
+    return getChangesFor<T>(object, null);
+  }
+
+  //TODO we need to decide if we want to have keyPaths be nullable here of we force developers to use getChanges in that case
```
papafe, line 494: What do you think?

papafe:
This is more or less reviewable. I'm still trying to understand why the backlink tests are failing (it seems that no backlink notifications are raised).
Apart from that I left some notes in the PR for things I'm not sure about, and I'll keep the list of TODOs updated in the PR description

packages/realm_dart/lib/src/native/realm_core.dart
```diff
@@ -1794,16 +1794,47 @@ class _RealmCore {
...
+      if (keyPaths != null && classKey != null) {
+        final length = keyPaths.length;
+        final keypathsNative = arena<Pointer<Char>>(length);
+
+        for (int i = 0; i < length; i++) {
+          keypathsNative[i] = keyPaths[i].toCharPtr(arena);
+        }
+
+        kpNative = _realmLib.invokeGetPointer(() => _realmLib.realm_create_key_path_array(object.realm.handle._pointer, classKey, length, keypathsNative));
+      }
```
nielsenko, line 1811: LGTM

Perhaps get the `classKey` from the `realmObject` instead of passing it as an argument?

packages/realm_dart/lib/src/native/realm_core.dart
```diff
@@ -1794,16 +1794,47 @@ class _RealmCore {
...
+        final length = keyPaths.length;
+        final keypathsNative = arena<Pointer<Char>>(length);
+
+        for (int i = 0; i < length; i++) {
+          keypathsNative[i] = keyPaths[i].toCharPtr(arena);
+        }
+
+        _realmLib.invokeGetPointer(() => _realmLib.realm_create_key_path_array(object.realm.handle._pointer, classKey, length, keypathsNative));
+      }
+    });
```
nielsenko, line 1837: Perhaps extract common code between `verifyKeyPath` and `subscribeObjectNotifications`

packages/realm_dart/lib/src/realm_object.dart
nielsenko: As a user, you will not be using these, but the property `changes` and the method `changesFor`, so I don't think it matter that much.

packages/realm_dart/lib/src/realm_object.dart
```diff
@@ -719,12 +728,24 @@ extension RealmObjectChangesInternal<T extends RealmObject> on RealmObjectChange
...
+        throw RealmException("It is not allowed to have empty key paths.");
+      }
+      realmCore.verifyKeyPath(realmObject, keyPaths, classKey);
+    }
+  }
 
   @override
   RealmNotificationTokenHandle subscribe() {
-    return realmCore.subscribeObjectNotifications(realmObject, this);
+    return realmCore.subscribeObjectNotifications(realmObject, this, keyPaths, classKey);
```
nielsenko, line 748: I notice you pass in both the `realmObject` and the `classKey` where the later can be inferred from the former. I suppose this is to avoid the overhead of calling `realmObject.realm.metadata.getByType(T).classKey;` twice. 

Maybe that is a bit of premature optimization. In particular since we eventually want to avoid creating `keypathsNative` twice anyway.

nielsenko (commented):
So far so good 👍 . I'll take another look once you set it ready-for-review.

For other reviewers. The double creation of the `keypathNative` array is a small blemish we already discussed and will address in a later PR

packages/realm_dart/lib/src/realm_object.dart
papafe: I think I put this comment in the wrong place, but I was actually referring to `changesFor`, that at the moment can accept a null argument. Should we keep it like this? Ot should we make the keypath argument non-nullable and force developers to use `changes` for the default notifications?

packages/realm_dart/lib/src/native/realm_core.dart
papafe: Definitely

packages/realm_dart/lib/src/realm_object.dart
```diff
@@ -701,7 +730,10 @@ class RealmObjectChanges<T extends RealmObjectBase> implements Finalizable {
   /// The property names that have changed.
   List<String> get properties {
     final propertyKeys = realmCore.getObjectChangesProperties(_handle);
-    return object.realm.getPropertyNames(object.runtimeType, propertyKeys);
+    return object.realm
+        .getPropertyNames(object.runtimeType, propertyKeys)
+        .map((e) => object.objectSchema.firstWhere((element) => element.mapTo == e || element.name == e).name)
+        .toList();
```
papafe, line 736: Not sure this is the way we want to do this.

nielsenko: While potentially inefficient, it is correct, and good enough for now.

nielsenko (approved):
LGTM

packages/realm_dart/lib/src/realm_object.dart
nielsenko: I think we should allow `null` as well, so `changes` and `changesFor(null)` is essentially the same.

PR #1607, opened 2024-03-27
RDART-992: Const initializer evaluation is too simple

### LINKED ISSUE #1606: Realm generator rejects const values saying initialization must be const
### What happened?

Using constant classes or negative numbers used to work for initialization.  When upgrading from 1.6 to 2.0, this no longer works:

```[SEVERE] realm:realm_generator on lib/model/data/picture.dart:

Field initializers must be constant

in: package:sustain/model/data/picture.dart:19:14
    ╷
16  │ @RealmModel()
17  │ class _Example {
    │       ━━━━━━━━ in realm model for 'Example'
... │
19  │   double y = -1;
    │              ^^ Must be const
    ╵
Ensure the default value for field "y" is const

```

### Repro steps

Create a class with integer or float field and try to initialize it to a constant negative value.
Presumably this would work (or rather fail to work) with any other class that provides a const constructor.

### Version

2.0.0

Local Database only

Flutter Application

### Client OS and version

Linux

### Code snippets

```
@RealmModel()
class _Example {
  double x = 0;
  double y = -1;
  int z = -1;
}
```

_No response_

### Relevant log output

```shell
[SEVERE] realm:realm_generator on lib/model/data/picture.dart:

Field initializers must be constant

in: package:sustain/model/data/picture.dart:19:14
    ╷
16  │ @RealmModel()
17  │ class _Example {
    │       ━━━━━━━━ in realm model for 'Example'
... │
19  │   double y = -1;
    │              ^^ Must be const
    ╵
Ensure the default value for field "y" is const


```

nielsenko:
Fix #1606

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 * Improve file compaction performance on platforms with page sizes greater than 4k (for example arm64 Apple platforms) for files less than 256 pages in size (Core 14.4.0).
 
 ### Fixed
+* Using `-1` as an initializer would fail. ([#1606](https://github.com/realm/realm-dart/issues/1606))
 
 ### Compatibility
 * Realm Studio: 13.0.0 or later.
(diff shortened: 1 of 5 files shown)


papafe (approved):
looks good

elle-j (approved):
Good fix 👍

PR #1612, opened 2024-04-02
RDART-992: Handle Identifer expression as well

nielsenko:
As mentioned in https://github.com/realm/realm-dart/issues/1606#issuecomment-2028498843 identifier expressions are not considered when establishing const-ness. This PR fixes that.


nielsenko:
Added `no-changelog` label, as this is a fix to a fix that is not yet released.

elle-j (approved):
👍

Issue #1595, opened 2024-03-21
Run Time Issue


Jyoshnarani:
### What happened?

A bug happened!

Application is going to crash

![ImportedPhoto_1711000799717](https://github.com/realm/realm-dart/assets/58125446/65c61412-3073-4f3f-8b4a-4a4b5a1f28f1)

### Repro steps

Showing error  

======== Exception caught by widgets library =======================================================
The following RealmError was thrown building Builder:
Realm error : Invalid argument(s): Failed to load dynamic library 'librealm_dart.so': dlopen failed: library "librealm_dart.so" not found
Invalid argument(s): Failed to load dynamic library '/system/bin/librealm_dart.so': dlopen failed: library "/system/bin/librealm_dart.so" not found
Invalid argument(s): Failed to load dynamic library 'librealm_dart.so': dlopen failed: library "librealm_dart.so" not found

The relevant error-causing widget was: 

### Version

Dart version 3.3.1

### What Atlas Services are you using?

Local Database only

### What type of application is this?

Flutter Application

### Client OS and version

Android 13

nielsenko:
Hi @Jyoshnarani 

What version of Realm? (2.0.0?)

Device or emulator (or both)?

Did the problem occur after upgrading an existing project, or when creating a new one? 

If upgrading, then from what version?

Did you try `flutter clean`?

If that didn't help. Did you try re-creating the android project:
```shell
rm -rf android
flutter create . --platforms=android # ,...
```

Jyoshnarani:
What version of Realm? (2.0.0?) -  **2.0.0 (I faced issue in 1.9.0 also)**
Device or emulator (or both)? - **both**
Did the problem occur after upgrading an existing project, or when creating a new one? -  **existing project**
If upgrading, then from what version? - **1.8.0**
Did you try flutter clean? - **yes**

nielsenko:
@Jyoshnarani  What desktop are you on? The instructions for re-creating the android flutter project is for macos, and linux.

Jyoshnarani:
@nielsenko , It's android flutter project

nielsenko:
Yes, but your development machine? What OS is that?

Jyoshnarani:
Desktop - Windows

nielsenko:
Could you try re-creating the android platform project, and report back.

nielsenko:
I have tried upgrading and downgrading between `1.8.0`, `1.9.0`, and `2.0.0` and running on an android emulator. 

[...]

PR #1614, opened 2024-04-02
 RDART-983: Refactor how we open dynamic library to give better error message

nielsenko:
This PR aims to give a better error message, if we fail to load the native shared library. 

In the process It changes how we attempt to open it.

First of all it only tries one place when running under the flutter vm, which is the most common scenario. This is not only more correct, but it also improves the error message on failure by not referring to unrelated paths.

Secondly, it will report a hint to add `realm` package if it fails. It is an easy mistake to only add `realm_dart`. We have seen one example already in #1595.

Fix: #1595 

#### Dart script:
```
dart run bin/init_realm_dart.dart
Unhandled exception:
Realm error : Could not open librealm_dart.dylib. Tried:
- "librealm_dart.dylib"
- "/Users/kasper/.puro/shared/caches/e76c956498841e1ab458577d3892003e553e4f3c/dart-sdk/bin/librealm_dart.dylib"
/Users/kasper/Projects/mongodb/experiments/init_realm_dart/binary/macos/librealm_dart.dylib
Hint: Did you forget to run `dart run realm_dart install`?
#0      _openRealmLib.throwError (package:realm_dart/src/init.dart:71:5)
#1      _openRealmLib (package:realm_dart/src/init.dart:97:5)
#2      initRealm (package:realm_dart/src/init.dart:120:24)
...
```

#### Dart pre-compiled exe:
```
./bin/init_realm_dart.exe
Unhandled exception:
Realm error : Could not open librealm_dart.dylib. Tried:
- "librealm_dart.dylib"
- "/Users/kasper/Projects/mongodb/experiments/init_realm_dart/bin/librealm_dart.dylib"
- "/Users/kasper/Projects/mongodb/experiments/init_realm_dart/binary/macos/librealm_dart.dylib"
Hint: Did you forget to run `dart run realm_dart install`?
#0      _openRealmLib.throwError (package:realm_dart/src/init.dart:71)
#1      _openRealmLib (package:realm_dart/src/init.dart:97)
#2      initRealm (package:realm_dart/src/init.dart:120)
```

#### Flutter app: 
```
[ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: Realm error : Could not open realm_dart. Tried:
- "/Users/kasper/Library/Developer/CoreSimulator/Devices/FB6A66E7-50AF-4C35-8118-83EFE909C4D3/data/Containers/Bundle/Application/80555876-B519-4C9E-A280-5CD535D09A4B/Runner.app/Frameworks/realm_dart.framework/realm_dart"
Hint: Did you forget to add a dependency on the realm package?
#0      _openRealmLib.throwError (package:realm_dart/src/init.dart:71:5)
#1      _openRealmLib (package:realm_dart/src/init.dart:84:29)
#2      initRealm (package:realm_dart/src/init.dart:120:24)
...
```

### DIFF
--- a/packages/realm_dart/lib/src/init.dart
+++ b/packages/realm_dart/lib/src/init.dart
@@ -71,10 +71,10 @@ DynamicLibrary _openRealmLib() {
     throw RealmError(
       [
         'Could not open $libName. Tried:',
-        (candidatePaths.join('\n')),
+        candidatePaths.map((p) => '- "$p"').join('\n'),
         isFlutterPlatform //
-            ? 'Did you forget to add a dependency on realm package?'
-            : 'Did you forget to run `dart run realm_dart install`?'
+            ? 'Hint: Did you forget to add a dependency on the realm package?'
+            : 'Hint: Did you forget to run `dart run realm_dart install`?'
       ].join('\n'),
     );
   }
--- a/packages/realm_dart/lib/src/init.dart
+++ b/packages/realm_dart/lib/src/init.dart
@@ -5,24 +5,17 @@ import 'dart:ffi';
 import 'dart:io';
 
 import 'package:path/path.dart' as p;
-import 'package:realm_common/realm_common.dart';
 
 import '../realm.dart' as realm show isFlutterPlatform;
 import '../realm.dart' show realmBinaryName;
 import 'cli/common/target_os_type.dart';
 import 'cli/metrics/metrics_command.dart';
 import 'cli/metrics/options.dart';
+import 'realm_class.dart';
 
 DynamicLibrary? _library;
 
-void _debugWrite(String message) {
-  assert(() {
-    print(message);
-    return true;
-  }());
-}
-
-String _getBinaryPath(String libName) {
+String _getPluginPath(String libName) {
   if (Platform.isAndroid) {
     return libName;
   }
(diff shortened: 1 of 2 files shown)


packages/realm_dart/lib/src/init.dart
```diff
@@ -65,26 +58,39 @@ String get _exeDirName => p.dirname(Platform.resolvedExecutable);
...
-  // Try to open lib from various candidate paths
-  List<String> errMessages = [];
-  for (final open in [
-    () => _open(libName), // just ask OS..
-    () => _open(p.join(_exeDirName, libName)), // try finding it next to the executable
-    if (root != null) () => _open(p.join(root, 'binary', Platform.operatingSystem, libName)), // try finding it relative to project
-    () => _open(_getBinaryPath(libName)), // find it where it is installed by plugin
-  ]) {
+
+  DynamicLibrary? tryOpen(String path) {
```
elle-j, line 62: Would you say that there's a preference in Dart to have e.g. `tryOpen()` and `throwError()` local within `_openRealmLib()` rather than private functions defined outside of the parent function?

packages/realm_dart/lib/src/init.dart
```diff
@@ -65,26 +58,39 @@ String get _exeDirName => p.dirname(Platform.resolvedExecutable);
...
+      return null;
     }
   }
-  throw RealmError(errMessages.join('\n'));
-}
 
-DynamicLibrary _open(String lib) => DynamicLibrary.open(lib);
+  Never throwError(Iterable<String> candidatePaths) {
+    final message = 'Could not open $libName at any of the following paths:\n${candidatePaths.join('\n')}'
+        '${isFlutterPlatform ? '\nDid you forget to add a dependency on realm package?' : 'Did you forget to run `dart run realm_dart install`?'}';
```
elle-j, line 72: Perhaps this could make it somewhat easier to parse.

(P.S. I think PR description examples showed an earlier iteration)
```suggestion
    final message = 'Could not open $libName at any of the following paths:\n- ${candidatePaths.join('\n- ')}'
        '${isFlutterPlatform ? '\nHint: Did you forget to add a dependency on realm package?' : '\nHint: Did you forget to run `dart run realm_dart install`?'}';
```

packages/realm_dart/lib/src/init.dart
```diff
@@ -65,26 +58,39 @@ String get _exeDirName => p.dirname(Platform.resolvedExecutable);
...
+    final candidatePaths = [
+      libName, // just ask OS..
+      p.join(_exeDirName, libName), // try finding it next to the executable
+      if (root != null) p.join(root, 'binary', Platform.operatingSystem, libName), // try finding it relative to project
+    ];
+    DynamicLibrary? lib;
+    for (final path in candidatePaths) {
+      lib = tryOpen(path);
+      if (lib != null) return lib;
+    }
```
elle-j, line 90: Could also extract this to a similar function as `tryOpen()` but take `paths` instead of `path` if you'd like.

elle-j (approved):
Nice with improved error messages 🙂 Just have some small suggestions.

packages/realm_dart/lib/src/init.dart
nielsenko: No, I wouldn't per se.

It signals/conveys/ensures that these can (and will) only be called from within the declaring function, and it also allows the inner functions to refer to state in the outer.

I could use the same trick other places in this file, but I decided to try and keep the PR tractable instead.

packages/realm_dart/lib/src/init.dart
nielsenko: Since that would only be called from one place, it seems a little excessive to me.

packages/realm_dart/lib/src/init.dart
elle-j: Alright 🙂

packages/realm_dart/lib/src/init.dart
```diff
@@ -5,24 +5,17 @@ import 'dart:ffi';
 import 'dart:io';
 
 import 'package:path/path.dart' as p;
-import 'package:realm_common/realm_common.dart';
```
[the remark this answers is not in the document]
nielsenko: To access `RealmError`, but that is exported from `realm_class.dart` so no need.

packages/realm_dart/lib/src/init.dart
nielsenko: Hmm.. I think I'm missing a push 😬

nielsenko: Added the '- ' prefix, the 'Hint: ' prefix, and wrapped paths in ""

PR #1617, opened 2024-04-03
RDART-996: Don't serialize backlinks

### LINKED ISSUE #1616: EJson export doesn't support backlinks
### Problem

Backlinks in a Realm object cause a stack overflow error when exporting to EJson.  This is somewhat expected, as following backlinks will certainly lead to circular references.

### Solution

Ideally we would be able to annotate fields to specify how we want to export them.  Maybe we some fields excluded, or other to be shallowly exported (id only).

### Alternatives

Currently we can write a custom export function, but then we don't get to use the magic of auto-generated code.

Would be a major improvement

Local Database only

nielsenko:
Backlinks are computed values and cannot be deserialized, and shouldn't be serialized either. Especially since we don't handle cycles during serialization gracefully, and backlinks are guarenteed cycles.

Fix #1616

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Fixed
 * Using valid const, but non-literal expressions, such as negation of numbers, as an initializer would fail. (Issue [#1606](https://github.com/realm/realm-dart/issues/1606))
-* Backlinks mistakenly included in EJson serialization ([Issue #1616](https://github.com/realm/realm-dart/issues/1616))
+* Backlinks mistakenly included in EJson serialization. ([Issue #1616](https://github.com/realm/realm-dart/issues/1616))
 
 ### Compatibility
 * Realm Studio: 13.0.0 or later.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 ### Fixed
 * Using valid const, but non-literal expressions, such as negation of numbers, as an initializer would fail. (Issue [#1606](https://github.com/realm/realm-dart/issues/1606))
+* Backlinks mistakenly included in EJson serialization ([Issue #1616](https://github.com/realm/realm-dart/issues/1616))
```
elle-j, line 9: Wondering if it's worth adding `since 2.0.0`? The current CHANGELOG pattern seems to only use `since` for Core related changes though.
```suggestion
* Backlinks mistakenly included in EJson serialization. ([Issue #1616](https://github.com/realm/realm-dart/issues/1616))
```

packages/realm_generator/test/good_test_data/all_types.expected
elle-j, line 1: Unsure if there's something testing that backlinks are not attempted to be de/serialized?

nielsenko: Added test case

CHANGELOG.md
nielsenko: We don't usually do that.  Also this bug has been there forever in the sense that it existed in the very first release supporting EJson.

PR #1621, opened 2024-04-08
New documentation URL

kneth:


nirinchev (commented):
We should probably update credentials.dart and user.dart as those have API docs that point to the old docs location.

nielsenko (approved):
LGTM

I have updated the other doc urls @nirinchev mentioned in the branch `kn/suggestions/kneth/new-docs-url`.

kneth:
I have merged suggestions from @nielsenko

Issue #1632, opened 2024-04-16
RealmValue.operator== bug


nielsenko:
`a == a` should always be true, also when a is a `RealmValue` containing a collection. Ie. reference equality implies equality.

PR #1633, opened 2024-04-16
RDART-1006: Reference equality implies equality

nielsenko:
Fixes: #1632


nielsenko:
One thing to be aware of still is that:
```dart
final x = [RealmValue.from(1)];
final xrv = RealmValue.from(x);
expect(xrv, xrv); // ref eq
expect(xrv, RealmValue.from(x)); // value ref eq

final y = [1];
final yrv = RealmValue.from(y);
expect(yrv, yrv); // ref eq
expect(RealmValue.from(y), isNot(RealmValue(y))); // value not ref eq, as internal list is mapped to RealmValue's
```

nirinchev (approved):

papafe (approved):

Issue #292, opened 2022-03-11
Generated RealmObject constructor should use named instead of positional argument


calvintam236:
For example, you have a datamodel class like this:
```dart
@RealmModel()
class _Hello {
  @PrimaryKey()
  final String key;

  final String sth1;

  final String sth2;

  final String sth3;

  final List<String> listA;

  final List<String> listB;
}
```

and this will generate a RealmObject class with this constructor:

```dart
class Hello extends _Hello with RealmObject {
  Hello(
    String key,
    String sth1,
    String sth2,
    String sth3, {
    Iterable<String> listA = const [],
    Iterable<String> listB = const [],
  }) {
    RealmObject.set(this, 'key', key);
    RealmObject.set(this, 'sth1', sth1);
    RealmObject.set(this, 'sth2', sth2);
    RealmObject.set(this, 'sth3', sth3);
    RealmObject.set<List<String>>(this, 'listA', listA.toList());
    RealmObject.set<List<String>>(this, 'listB', listB.toList());
  }
  ...
}
```

Now, when you create a new `Hello` object, you need to know exactly the order of the arguments of the constructor:
```dart
Hello(key, sth1, sth2, sth3, listA, listB);
```

THIS IS NOT FRIENDLY.

It should simply use named arguments in the constructor so order of the arguments does not matter & provide some flexibility (like skipping `listA` for default value and passing `listB` at the same time):
```dart
Hello(
  key: key,
  ...
  sth3: sth3,
  listB: listB,
)
```

blagoev:
@calvintam236 Thanks for posting. 
This is true only for required properties. If you have a nullable (optional in Realm) property in your model than it is generated as named. The order should be the same as defined in your data model. 
Also is this really an issue if you have intellisense in the editor.

calvintam236:
> @calvintam236 Thanks for posting. This is true only for required properties. If you have a nullable (optional in Realm) property in your model than it is generated as named. The order should be the same as defined in your data model. Also is this really an issue if you have intellisense in the editor.
> 
> EDIT: for non nullable properties with default values we do generate named arguments as well.

Thanks for getting back.

1. Even they are in the same order as that data model class, you still need to open that file to confirm all arguments are in correct order.
2. If I have 10 positional arguments in one constructor, does that make sense to take them positional?
3. Please also keep in mind that NOT all developers use IDE to code. (I'm using code editor and run analyze tool on terminal. A lot of IDEs out there are performance heavy.)

[...]

PR #1641, opened 2024-04-18
RDART-1010: Allow configuration of generator per realm model class

nielsenko:
Support per realm model configuration of the generator via the `GeneratorConfig` class. Currently we only support specifying the constructor style to use, as this has been a very frequent request (see #292).

  ```dart
  // somewhere define your preferred generator configuration and a const RealmModel that uses it
  const config = GeneratorConfig(ctorStyle: CtorStyle.allNamed);
  const realmModel = RealmModel.using(generatorConfig: config);

  @realmModel // <-- use here
  class _Person {
    late String name;
    int age = 42;
  }
  ```
  will generate a constructor like:
  ```dart
  Person({
    required String name,
    int age = 42,
  }) { ... }
  ```

Fixes: #292
Fixes: #1642

### DIFF
--- a/packages/realm_common/lib/src/realm_common_base.dart
+++ b/packages/realm_common/lib/src/realm_common_base.dart
@@ -20,7 +20,7 @@ enum ObjectType {
   /// to query or modify it.
   asymmetricObject('AsymmetricObject', 2);
 
-  const ObjectType([this._className = 'Unknown', this._flags = -1]);
+  const ObjectType(this._className, this._flags);
 
   final String _className;
   final int _flags;
--- a/packages/realm_common/lib/src/realm_common_base.dart
+++ b/packages/realm_common/lib/src/realm_common_base.dart
@@ -23,7 +23,6 @@ enum ObjectType {
   const ObjectType([this._className = 'Unknown', this._flags = -1]);
 
   final String _className;
-
   final int _flags;
 }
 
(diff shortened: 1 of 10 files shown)


packages/realm_common/lib/src/realm_types.dart
```diff
@@ -85,18 +85,12 @@ enum RealmCollectionType {
...
-      default:
-        return "none";
-    }
-  }
+  String get plural => switch (this) {
+      RealmCollectionType.list => 'lists',
+      RealmCollectionType.set => 'sets',
+      RealmCollectionType.map => 'maps',
+      _ => 'none'
+    };
```
nielsenko, line 93: A small drive by edit. Not related to this PR

packages/realm_generator/test/good_test_data/private_fields.expected
```diff
@@ -13,15 +13,15 @@ class WithPrivateFields extends _WithPrivateFields
 
   WithPrivateFields(
     String _plain, {
-    int _withDefault = 0,
+    int withDefault = 0,
```
nielsenko, line 16: This is actually fixing a bug in version 2.1.0. Named arguments cannot start with underscore.

packages/realm_generator/test/good_test_data/all_named_ctor.expected
```diff
@@ -0,0 +1,82 @@
...
+// **************************************************************************
+
+// ignore_for_file: type=lint
+class Person extends _Person with RealmEntity, RealmObjectBase, RealmObject {
+  static var _defaultsSet = false;
+
+  Person({
+    required String name,
+    int age = 42,
+  }) {
```
nielsenko, line 16: The interesting bit

packages/realm_common/lib/src/realm_common_base.dart
```diff
@@ -33,15 +32,49 @@ extension ObjectTypeInternal on ObjectType {
...
 ///
 /// {@category Annotations}
 class RealmModel {
-  /// The base type of the object
-  final ObjectType type;
-
-  /// Creates a new instance of [RealmModel] specifying the desired base type.
-  const RealmModel([this.type = ObjectType.realmObject]);
+  /// The base type of the generated object class
+  final ObjectType baseType;
```
nielsenko, line 61: Are you okay with this?

Dararii:
Can someone tell me how to implement this in Real: 2.1.0? 

These lines are resulting in references not found:
const config = GeneratorConfig(ctorStyle: CtorStyle.allNamed);
const realmModel = RealmModel.using(generatorConfig: config);

What class should I import?

nielsenko:
This feature is not released yet. It will go out with the next release.

PR #1650, opened 2024-04-24
Temporary file descriptor fix

nielsenko:
After upgrading xcode/macos some tests are prone to run out of file-descriptors. Simply bumping ulimit on the host running the simulator don't work.

This PR tags some test suites as using `many_files` and exclude those when testing on iOS.


nielsenko:
Not needed

PR #1651, opened 2024-04-25
RDART-1015: Update core to v14.6.2

nielsenko:
Fixes: #1653

### DIFF
--- a/packages/realm_dart/ffigen.yaml
+++ b/packages/realm_dart/ffigen.yaml
@@ -27,6 +27,7 @@ comments:
 compiler-opts:
   - '-DRLM_NO_ANON_UNIONS'
   - '-DFFI_GEN'
+  - '-DREALM_APP_SERVICES'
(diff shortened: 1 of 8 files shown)


packages/realm_dart/ffigen.yaml
```diff
@@ -27,6 +27,7 @@ comments:
 compiler-opts:
   - '-DRLM_NO_ANON_UNIONS'
   - '-DFFI_GEN'
+  - '-DREALM_APP_SERVICES'
```
nirinchev, line 30: this should be on, by default, shouldn't it? Why do we have to define it here?

nielsenko: It is on by default in cmake. This is the ffigen config.

nielsenko:
Updated test to match semantics introduced by https://github.com/realm/realm-core/issues/7452

PR #1652, opened 2024-04-30
RDART-950: Update nullability annotations for base url API

nirinchev:
Updates nullability annotations - the getter should never return `null`, while the setter should accept `null` to indicate the developer wants to reset the base url to the default value.

Fixes https://github.com/realm/realm-dart/issues/1523.

### DIFF
--- a/packages/realm_dart/lib/src/app.dart
+++ b/packages/realm_dart/lib/src/app.dart
@@ -219,16 +219,17 @@ class App implements Finalizable {
 
   /// Returns the current value of the base URL used to communicate with the server.
   @experimental
-  Uri? get baseUrl {
-    return Uri.tryParse(realmCore.getBaseUrl(this) ?? '');
+  Uri get baseUrl {
+    return Uri.parse(realmCore.getBaseUrl(this));
   }
 
   /// Temporarily overrides the [baseUrl] value from [AppConfiguration] with a new [baseUrl] value
-  /// used for communicating with the server.
+  /// used for communicating with the server. If set to `null`, the app will revert to the default
+  /// base url.
   ///
   /// The App will revert to using the value in [AppConfiguration] when it is restarted.
   @experimental
-  Future<void> updateBaseUrl(Uri baseUrl) async {
+  Future<void> updateBaseUrl(Uri? baseUrl) async {
     return await realmCore.updateBaseUrl(this, baseUrl);
   }
 
--- a/packages/realm_dart/lib/src/app.dart
+++ b/packages/realm_dart/lib/src/app.dart
@@ -218,6 +218,9 @@ class App implements Finalizable {
   }
 
   /// Returns the current value of the base URL used to communicate with the server.
+  ///
+  /// If an [updateBaseUrl] operation is currently in progress, this value will not
+  /// be updated with the new value until that operation has completed.
   @experimental
   Uri get baseUrl {
     return Uri.parse(realmCore.getBaseUrl(this));
--- a/packages/realm_dart/lib/src/app.dart
+++ b/packages/realm_dart/lib/src/app.dart
@@ -227,6 +227,10 @@ class App implements Finalizable {
   /// used for communicating with the server. If set to `null`, the app will revert to the default
   /// base url.
   ///
+  /// If this operation fails, the app will continue to use the original base URL. If another [App]
+  /// operation is started while this function is in progress, that request will use the original
+  /// base URL location information.
+  ///
   /// The App will revert to using the value in [AppConfiguration] when it is restarted.
   @experimental
   Future<void> updateBaseUrl(Uri? baseUrl) async {
--- a/packages/realm_dart/test/test.dart
+++ b/packages/realm_dart/test/test.dart
@@ -597,7 +597,7 @@ Future<void> baasTest(
 
   skip = shouldSkip(skip);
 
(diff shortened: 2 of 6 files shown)


packages/realm_dart/test/test.dart
```diff
@@ -597,7 +597,7 @@ Future<void> baasTest(
 
   skip = shouldSkip(skip);
 
-  test('[BAAS] $name', () async {
```
nirinchev, line 600: Had to remove that because the VSCode test plugin was now complaining with `No tests match regular expression ^$name` - note the ^ at the start, which was failing to match the renamed test that had a `[BAAS]` prefix.

nielsenko (approved):

packages/realm_dart/lib/src/app.dart
```diff
@@ -219,16 +219,17 @@ class App implements Finalizable {
...
-  Uri? get baseUrl {
-    return Uri.tryParse(realmCore.getBaseUrl(this) ?? '');
+  Uri get baseUrl {
+    return Uri.parse(realmCore.getBaseUrl(this));
   }
 
   /// Temporarily overrides the [baseUrl] value from [AppConfiguration] with a new [baseUrl] value
-  /// used for communicating with the server.
+  /// used for communicating with the server. If set to `null`, the app will revert to the default
+  /// base url.
```
elle-j, line 228: I'm wondering if the DX and the intent would be clearer if the SDKs provided a separate zero-arguments API for reverting to the default base URL, rather than passing `null` (or empty string for the SDKs accepting a string) to the update-API, even if that's how the Core API currently works. Thoughts?

packages/realm_dart/lib/src/app.dart
```diff
@@ -219,16 +219,17 @@ class App implements Finalizable {
 
   /// Returns the current value of the base URL used to communicate with the server.
```
elle-j, line 220: Just added this to the JS API docs as well (taken from Core). May be valuable to point out the extra information.
```suggestion
  /// Returns the current value of the base URL used to communicate with the server.
  ///
  /// If an [updateBaseUrl] operation is currently in progress, this value will not
  /// be updated with the new value until that operation has completed.
```

packages/realm_dart/lib/src/app.dart
```diff
@@ -219,16 +219,17 @@ class App implements Finalizable {
...
-    return Uri.tryParse(realmCore.getBaseUrl(this) ?? '');
+  Uri get baseUrl {
+    return Uri.parse(realmCore.getBaseUrl(this));
   }
 
   /// Temporarily overrides the [baseUrl] value from [AppConfiguration] with a new [baseUrl] value
-  /// used for communicating with the server.
+  /// used for communicating with the server. If set to `null`, the app will revert to the default
+  /// base url.
   ///
```
elle-j, line 229: Similar as a previous comment 👍 
```suggestion
  ///
  /// If this operation fails, the app will continue to use the original base URL. If another [App]
  /// operation is started while this function is in progress, that request will use the original
  /// base URL location information.
  ///
```

elle-j (approved):
Nice 🙂 Just left a comment that would be good to know if the SDKs should align on regarding reverting the url.

packages/realm_dart/test/test.dart
nielsenko: I have another inbound PR where `baas` is added as a tag instead, so vscode test-runner still can figure out how to run the test.

packages/realm_dart/lib/src/app.dart
nirinchev: I think providing zero-argument API will be more confusing as it's not obvious what `app.updateBaseUrl()` will update it to. In this case I feel `null` == `default` is a reasonable assumption developers can make. Alternatively, we could provide another method `app.resetBaseUrl()` to the same effect, but I'm not convinced this functionality justifies 3 methods on the app class.

I'm leaning toward leaving the API proposed in this PR (i.e. accepting null/"" for reseting the url to the default) and then re-evaluating based on user feedback/broader discussion as we revisit the API and move it out of the experimental phase.

elle-j: With zero-argument API I meant a different (3rd) API taking zero args, e.g. `resetBaseUrl()` like you mentioned. The flexible part atm is that it's experimental, so alrighty we can see at a later point if it'd make sense or not 🙂

nielsenko: I prefer to just have a single `updateBaseUrl` method and have a `null` argument reset it.

One thing to consider is to align on the API across SDKs. From the rumblings in the office it seems we are at risk of diverting a bit.

elle-j: From the original design we should have accepted only strings (or Uri), so for TS, we're currently only allowing the empty string as a reset (passing `null` would do the same but we haven't added it to the type). If we should also allow `null` to align more, we could do that as well. For the JS SDK, in the future it may be more idiomatic for us to either use a 3rd API or an options arg like `{ reset: true }`.

nirinchev: I'd say we should allow `null` here - I don't think it's obvious that you can pass an empty string to achieve the reset behavior, while `null`/`undefined` is special enough to convey that better.

PR #1669, opened 2024-05-06
RDART-1021: Wire up some basic dynamic setting and change notifications

nirinchev:
Fixes https://github.com/realm/realm-dart/issues/1668

Additionally, had to add `Realm.dynamic.create` and `RealmObject.dynamic.set` to be able to test.

TODO:
- [x] Change notifications for collections of objects
- [x] Tests for setting of different property types
- [x] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,9 @@
 ### Enhancements
 * Added support for creating and storing a RealmObject using the `Realm.dynamic` API: `realm.dynamic.create("Person", primaryKey: 123)`. (PR [#1669](https://github.com/realm/realm-dart/pull/1669))
 * Added support for setting properties on a RealmObject using the dynamic API: `obj.dynamic.set("name", "Peter")`. (PR [#1669](https://github.com/realm/realm-dart/pull/1669))
+* Listening for `.changes` on a dynamic object (obtained via the `realm.dynamic` API) no longer throws. (Issue [#1668](https://github.com/realm/realm-dart/issues/1668))
 
 ### Fixed
-* Listening for `.changes` on a dynamic object (obtained via the `realm.dynamic` API) no longer throws. (Issue [#1668](https://github.com/realm/realm-dart/issues/1668))
 * Private fields did not work with default values. (Issue [#1663](https://github.com/realm/realm-dart/issues/1663))
 
 ### Compatibility
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,11 @@
 ## vNext (TBD)
 
 ### Enhancements
-* None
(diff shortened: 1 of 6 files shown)


nirinchev:
I don't think this should conflict too much with #1550, but happy to merge the handles PR first.

packages/realm_dart/test/dynamic_realm_test.dart
```diff
@@ -222,6 +226,95 @@ void main() {
...
+      actual.dynamic.set('intProp', 42);
+      actual.dynamic.set('nullableIntProp', -42);
+
+      actual.dynamic.set('decimalProp', Decimal128.fromDouble(500));
+      actual.dynamic.set('nullableDecimalProp', Decimal128.infinity);
+
+      actual.dynamic.set('realmValueProp', RealmValue.from([true, 5]));
+    });
+
+    expect(actual.dynamic.get('stringProp'), isNot(expected.stringProp));
```
elle-j, line 261: Would it not be preferable to test against the exact updated value instead of `isNot` here?

elle-j (approved):
LGTM 🙂

packages/realm_dart/test/dynamic_realm_test.dart
nirinchev: Sure, we could also do that - I was mostly lazy here.

CHANGELOG.md
```diff
@@ -1,10 +1,11 @@
 ## vNext (TBD)
 
 ### Enhancements
-* None
+* Added support for creating and storing a RealmObject using the `Realm.dynamic` API: `realm.dynamic.create("Person", primaryKey: 123)`. (PR [#1669](https://github.com/realm/realm-dart/pull/1669))
+* Added support for setting properties on a RealmObject using the dynamic API: `obj.dynamic.set("name", "Peter")`. (PR [#1669](https://github.com/realm/realm-dart/pull/1669))
 
 ### Fixed
-* None
+* Listening for `.changes` on a dynamic object (obtained via the `realm.dynamic` API) no longer throws. (Issue [#1668](https://github.com/realm/realm-dart/issues/1668))
```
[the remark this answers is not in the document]
nirinchev: I guess you can argue both ways - it was reported as a bug, but technically we didn't intend it to work, so maybe it's an enhancement.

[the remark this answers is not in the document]
elle-j: Gotcha, I think I would see it more of an enhancement then.

packages/realm_dart/test/dynamic_realm_test.dart
elle-j: Then we can catch potential incorrect updates 👍

PR #1681, opened 2024-05-21
RDART-1034: Update realm-core to v14.7.0

nielsenko:
Fixeds: #1685


.github/workflows/build-native.yml
```diff
@@ -51,6 +51,10 @@ jobs:
         if: startsWith(matrix.build, 'android')
         run: echo "ANDROID_NDK=$ANDROID_NDK_LATEST_HOME" >> $GITHUB_ENV
 
+      - name: Downgrade XCode for MacOS
```
nirinchev, line 54: We need to move this job to macos-12 for this to work.

nielsenko: Using XCode 14.3.1 instead.

nirinchev (approved):

Issue #1670, opened 2024-05-07
Realm binary files take a lot of space in apk file, and include binary files for non-target platform.


zhxst:
### What happened?

My previous Flutter app takes about 63 MB in apk release build.
After import Realm, the apk file size become 131 MB, which is double from before.

By unpack the release package. It turns out that Realm include binary files for all the target platform.
By default, flutter only build for `android-arm`, `android-arm64` and `android-x64`. But realm also include `android-x86` binary files. Which takes 18.6MB extra space.

Here is another example , which only target `android-arm64` when build:
![image](https://github.com/realm/realm-dart/assets/11647836/bacb02e4-e3e2-481a-9fd1-0e5671a34845)

So please make realm package follow the `flutter --target-platform`, this will save a lot of space for release build apks.

Thank you.

### Repro steps

1.  Use any Flutter project which import realm.
2. Build android apk, use `--target-platform` or `--split-per-abi`.
3. Unpack the generate apks. Check the realm binary files inside apk.

### Version

3.19.5

### What Atlas Services are you using?

Local Database only

### What type of application is this?

Flutter Application

### Client OS and version

macOS 13.6

nirinchev:
@nielsenko do you know if Google Play repackages flutter apps to remove binaries for architectures that don't match the install device? I remember this used to work for .NET apps but I'm not sure if there's anything in Flutter that would prevent this mechanism from working.

nielsenko:
I'm not sure where you get the 161MB from. If I do `flutter build apk` on a basic project using realm, I get an apk size of 36.8MB. 
```sh
flutter build apk

Directory: '/Users/kasper/.puro/shared/pub_cache/hosted/pub.dev/realm-2.2.1/android/src/main/cpp/lib'
Realm binaries for 2.2.1 already downloaded
Realm install command finished.
Running Gradle task 'assembleRelease'...                            5.6s
✓ Built build/app/outputs/flutter-apk/app-release.apk (36.8MB)
```
If I use `flutter build apk --split-per-abi` I get:
```sh
flutter build apk --split-per-abi

Directory: '/Users/kasper/.puro/shared/pub_cache/hosted/pub.dev/realm-2.2.1/android/src/main/cpp/lib'
Realm binaries for 2.2.1 already downloaded
Realm install command finished.
Running Gradle task 'assembleRelease'...                            7.8s
✓ Built build/app/outputs/flutter-apk/app-armeabi-v7a-release.apk (9.8MB)
[...]

PR #1682, opened 2024-05-22
RDART-1022: Drop x86 as target

nielsenko:
Reduce android bundle size.

There is no need to build for `x86` as flutter itself does not support it.

Fixes: #1670


fealebenpae (approved):

nirinchev:
Needs an update to the gha workflows it seems.

Issue #1676, opened 2024-05-16
argument value for 'return_value' is null


zzyzy:
### What happened?

Realm operations will get ArgumentError exception in Flutter 3.22. The app does not crash in Profile/Release mode.

This happens to both Android and iOS.

This does not happen if we downgrade to Flutter 3.19.

argument value for 'return_value' is null
RealmLibrary.realm_scheduler_perform_work (d:\Users\zeezee\AppData\Local\Pub\Cache\hosted\pub.dev\realm_dart-2.2.1\lib\src\native\realm_bindings.dart:8685)
_RealmCore.invokeScheduler (d:\Users\zeezee\AppData\Local\Pub\Cache\hosted\pub.dev\realm_dart-2.2.1\lib\src\native\realm_core.dart:654)
new Scheduler._.<anonymous closure> (d:\Users\zeezee\AppData\Local\Pub\Cache\hosted\pub.dev\realm_dart-2.2.1\lib\src\scheduler.dart:32)
_RawReceivePort._handleMessage (d:\Users\zeezee\fvm\versions\3.22.0\bin\cache\pkg\sky_engine\lib\_internal\vm\lib\isolate_patch.dart:184)

![image](https://github.com/realm/realm-dart/assets/6206954/6736ac51-125e-44fa-8130-5c0e5d3afec4)

### Repro steps

1. Update app to Flutter 3.22
2. Run the app in debug mode (make sure "Uncaught Exceptions" is checked)
3. The exception will be thrown whenever any Realm operations takes place

### Version

3.22

### What Atlas Services are you using?

Local Database only

### What type of application is this?

Flutter Application

### Client OS and version

Windows 11 and macOS Sonoma 14.1.1

### Stacktrace of the exception/crash you're getting

```shell
argument value for 'return_value' is null
RealmLibrary.realm_scheduler_perform_work (d:\Users\zeezee\AppData\Local\Pub\Cache\hosted\pub.dev\realm_dart-2.2.1\lib\src\native\realm_bindings.dart:8685)
_RealmCore.invokeScheduler (d:\Users\zeezee\AppData\Local\Pub\Cache\hosted\pub.dev\realm_dart-2.2.1\lib\src\native\realm_core.dart:654)
new Scheduler._.<anonymous closure> (d:\Users\zeezee\AppData\Local\Pub\Cache\hosted\pub.dev\realm_dart-2.2.1\lib\src\scheduler.dart:32)
[...]

PR #1683, opened 2024-05-22
 RDART-1028: Use Zone.current.bindUnaryCallbackGuarded for RawReceivePort.handler

nielsenko:
Fixes: #1676

### DIFF
--- a/packages/realm_dart/lib/src/scheduler.dart
+++ b/packages/realm_dart/lib/src/scheduler.dart
@@ -21,6 +21,17 @@ class Scheduler {
 
   Scheduler._() {
     _receivePortFinalizer.attach(this, _receivePort, detach: this);
+    // There be dragons here!!!
+    //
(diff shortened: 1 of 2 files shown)


packages/realm_dart/lib/src/scheduler.dart
```diff
@@ -20,25 +21,25 @@ class Scheduler {
...
-        final text = message[2] as String;
-        Realm.logger.raise((category: category, level: level, message: text));
-      } else if (message is int) {
-        realmCore.invokeScheduler(message);
-      } else {
-        Realm.logger.log(LogLevel.error, 'Unexpected Scheduler message type: ${message.runtimeType} - $message');
-      }
-    };
-
+    _receivePort.handler = Zone.current.bindUnaryCallbackGuarded(_handle);
```
nirinchev, line 24: Can we add a comment what that does and why it's needed?

nirinchev (approved):

Issue #1374, opened 2023-08-10
Ignore Realm from Web.


rdunlocked18:
### Problem
So I have one application that needs to be built on Native Platforms and Web. Moreover, When try to build web project the realm will break the whole web support. Is there any way that I can ignore realm binaries from the web build? I know its not supported on WEB but is there any fallback that we can use to ignore ?

TODO :

1. Ignore realm.dart on web 

I tried to remove this by [Stackoverflow](https://stackoverflow.com/a/60153630) this method but doesn't seem to work.

Please help. Thanks.

### Screenshot
![](https://github.com/realm/realm-dart/assets/29021926/5f342093-1c06-4643-b78f-5a286d8d46ae)

### How important is this improvement for you?
Dealbreaker

### Feature would mainly be used with
Local Database only

victorcode1:
Please

richard457:
We all need this.

nirinchev:
I haven't been able to find an easy way to achieve this. If you folks have any suggestions, we'd be more than willing to take pull requests.

shockbytes:
I have a similar problem. I need Realm to migrate data from my legacy Android App to the new Flutter App. 
Since the Flutter App also supports Web as a platform, I would need to have web support for Realm too (although I won't use it, it just should compile)

dotjon0:
We have the same problem. +1 for this. If Realm is not compatible for a target OS/platform (i.e. web in this case) Realm should not block compiling... Flutter is a multi OS/platform framework, but it appears with Realm its not.

nielsenko:
The typical trick to use is:
```dart
import 'facade.dart'
    if (dart.library.html) 'facade_html.dart' // don't use realm here
    if (dart.library.io) 'facade_io.dart'; // here you can use realm
```
but since you ask, I suspect this is not sufficient for you. Could you elaborate a bit on what you hope for?

rdunlocked18:
Hi All,

@nirinchev , @richard457 , @shockbytes , @panalgin ,  @nielsenko , @dotjon0  @

So this has worked for me from like the time when I raised this thing. After spending many hours, I found that the trick I am using is working and hence being used in my live code. This won't work if all your code is dumped in a single file. This is a bit of hustle but once you are on to it can be easily used. 💯  I hope this works. 😀

1. I am currently using CLEAN hence was easy to manage.

Let's Say you have a class called ```UserInfoRepository``` that has all the methods for io and web.

[...]

PR #1699, opened 2024-05-29
RDART-866: Minimal web support

nielsenko:
The aim of this PR is to allow package realm to be included in flutter web projects without breaking compilation. It is **not** intended to actually make realm work runtime with flutter web.

It is safe to use unmanaged realm object in web projects, except if they include `Decimal128` fields.

Fixes: #1374


packages/realm_dart/lib/src/realm_web.dart
```diff
@@ -0,0 +1,11 @@
+// Copyright 2024 MongoDB, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//export 'realm_class.dart' hide RealmInternal;
```
nirinchev, line 4: Should this be deleted?

nirinchev (approved):
Looks good - I skimmed over large portions of repetitive changes, but those seem to be covered in tests, so they seem to work. One thing to consider is extracting the web exception somewhere so we don't have to repeat it 58329529 times.

nielsenko:
> Looks good - I skimmed over large portions of repetitive changes, but those seem to be covered in tests, so they seem to work. One thing to consider is extracting the web exception somewhere so we don't have to repeat it 58329529 times.

I have add a base-class on all we handles, and have it implement `noSuchMethod`. Also the actual throw has been refactored into the support function `webNotSupported`.

packages/realm_dart/lib/src/handles/native/collection_handle_base.dart
```diff
@@ -3,23 +3,22 @@
...
   CollectionHandleBase(super.root, super.pointer, super.size);
 }
-
 void createCollection(Realm realm, RealmValue value, Pointer<realm_list> Function() createList, Pointer<realm_dictionary> Function() createMap) {
   CollectionHandleBase? collectionHandle;
   try {
     switch (value.collectionType) {
       case RealmCollectionType.list:
-        final listHandle = ListHandle(createList(), realm.handle);
+        final listHandle = ListHandle(createList(), realm.handle as RealmHandle);
```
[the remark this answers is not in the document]
nielsenko: Not fan of that.. I prefer all methods on the platform specific handles (including ctors) to traffic in platform specific handles. The problem is more in the `handle` extension method on `Realm`.

packages/realm_dart/lib/src/handles/native/mutable_subscription_set_handle.dart
```diff
@@ -86,5 +93,11 @@ class MutableSubscriptionSetHandle extends SubscriptionSetHandle {
     });
   }
 
+  @override
   void clear() => realmLib.realm_sync_subscription_set_clear(_mutablePointer).raiseLastErrorIfFalse();
+
+  @override
+  // Workaround for weird compiler bug
+  // ignore: unnecessary_overrides
+  SubscriptionHandle? findByResults(covariant ResultsHandle results) => super.findByResults(results);
```
nirinchev, line 102: Can we add a bit more details for the weird compiler bug? Just so that we don't have to guess what that meant 6-12 months from now.

packages/realm_dart/lib/src/handles/web/decimal128.dart
```diff
@@ -0,0 +1,39 @@
...
+  static Decimal128? tryParse(String source) => throw UnimplementedError('web not supported');
+
+  /// Parses a string into a [Decimal128]. Throws a [FormatException] if the string is not a valid [Decimal128].
+  factory Decimal128.parse(String source) => throw UnimplementedError('web not supported');
+
+  /// Converts a `int` into a [Decimal128].
+  factory Decimal128.fromInt(int value) => throw UnimplementedError('web not supported');
+
+  /// Converts a `double` into a [Decimal128].
+  factory Decimal128.fromDouble(double value) => throw UnimplementedError('web not supported');
```
nirinchev, line 35: Can these not use `webNotSupported`?

packages/realm_dart/lib/src/realm_object.dart
```diff
@@ -175,6 +175,7 @@ class RealmCoreAccessor implements RealmAccessor {
           if (listMetadata != null && _isTypeGenericObject<T>()) {
             switch (listMetadata.schema.baseType) {
               case ObjectType.realmObject:
+                //ManagedRealmList<RealmObject>._(handle, object.realm, listMetadata);
```
nirinchev, line 178: Why do we need this comment?

packages/realm_dart/test/utils/web/platform_util.dart
```diff
@@ -0,0 +1,11 @@
+// Copyright 2024 MongoDB, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+import '../platform_util.dart' as intf;
+
+class PlatformUtil implements intf.PlatformUtil {
+  const PlatformUtil();
+
+  @override
+  noSuchMethod(Invocation invocation) => throw UnsupportedError('web not supported');
```
nirinchev, line 10: `webNotSupported()`?

nirinchev (approved):

nirinchev:
No, `Decimal128` won't be supported because it relies on a native implementation. Unfortunately, this type is not implemented in pure dart and it's not obvious that such an implementation would be easy/possible with the primitives we have there.

nielsenko:
> could we fallback to a String in the case the RealmModel has a Decimal128 type then (for web only)? We can then deal with it there onwards. Otherwise we are going to have lots of Realm Models we cant use in 'web'.

No. The generator run before we know if the model will be used on web or not.

nielsenko:
> Other than this, if we Realm cant conditionally build based on flutter target we are going to get lots of models with type issues...so web will not work...

We will add web support for `Decimal128` based on https://pub.dev/packages/decimal

packages/realm_dart/lib/src/handles/web/decimal128.dart
nielsenko: This will be addressed in #1713

nielsenko:
This is now released with [realm: ^3.0.0](https://pub.dev/packages/realm/versions/3.0.0)

PR #1710, opened 2024-06-05
Github composite action for setting up flutter on runner

nielsenko:
- Replace `subosito/flutter-action@v2` with a [puro](https://puro.dev/) based setup.
- Use a [composite github action](https://docs.github.com/en/actions/creating-actions/creating-a-composite-action) to simplify the setup of flutter (via puro) and melos in the workflow jobs.
- Add caching to the full flutter setup, including the pub-cache, flutter, vm-engine, the lots.
- Move pub-cache to `D:` drive (same as `$GITHUB_WORKSPACE`) for IO speed.

Currently caching is disabled on windows, but this PR is still faster wrt. to flutter setup on windows, than what we currently use. However, it is mostly for the simplifications we should consider this. Overall the setup steps in our jobs does n't contribute a lot to the full runtime.


.github/actions/setup-runner/action.yml
```diff
@@ -0,0 +1,84 @@
...
+
+    - name: Melos bootstrap
+      shell: bash
+      run: puro pub global run melos bootstrap
+
+    - name: Save cache
+      # Saving on feature branches will only benefit reruns, so we don't bother.
+      # Feature branches inherit the cache from the main branch anyway.
+      # This saves on cache storage.
+      if: github.ref != 'refs/heads/main' && steps.restore.outputs.cache-hit != 'true'
```
nirinchev, line 80: Should this be `github.ref == 'refs/heads/main'`? Also, we seem to be only be storing the cache if we don't have a cache hit, but then we use the cache primary key as input to the save action, which seems odd.

nielsenko: @nirinchev You are a bit early 😊 . Yes it should be `==`.. the `!=` is just for debugging.

You cannot update a cache in github actions, only write a new one, so if you have an exact hit, trying save on the same key would fail. The primary key is the key we tried to lookup, and where we will store the updated cache, if either none existed, or the one matched pertained to a stale set of `pubspec.yaml` files.

nielsenko:
@nirinchev I think I will accept that caching is simply disabled on windows for now. So I have opened the PR for review.

nirinchev (approved):

PR #1713, opened 2024-06-06
RDART-866: kn/decimal128 web support

nielsenko:
As part addressing #1374 in it was found that not allowing `Decimal128` fields on unmanaged realm models when used on the web platform was problematic for one of our users. Hence this PR adds support for that as well.

This is not a fully compliant IEEE 754-2008 Decimal128 implementation, as it is just based on the [decimal](https://pub.dev/packages/decimal) package. Which is based on the [rational](https://pub.dev/packages/rational) package, which is again based on the [BigInt](https://api.flutter.dev/flutter/dart-core/BigInt-class.html) class.

The issues are mostly in some of the odd corner cases of IEEE 754-2008 floating point values, such as:
* -0 < 0,
* NaN != NaN,
* 1 / 0 = Inf, etc.

Currently anything involving `Decimal128.nan`, `Decimal128.infinity`, and `Decimal128.negativeInfinity` will throw on web, including `Decimal128.isNaN`.

Also, be warned that this class is incredible slow compared to the native implementation, especially division.


packages/realm_dart/lib/src/handles/web/decimal128.dart
```diff
@@ -23,17 +43,94 @@ class Decimal128 implements intf.Decimal128 {
...
-  factory Decimal128.parse(String source) => webNotSupported();
+  factory Decimal128.parse(String source) => Decimal128._(Decimal.parse(source));
 
   /// Converts a `int` into a [Decimal128].
-  factory Decimal128.fromInt(int value) => webNotSupported();
+  factory Decimal128.fromInt(int value) => Decimal128._(Decimal.fromInt(value));
 
   /// Converts a `double` into a [Decimal128].
-  factory Decimal128.fromDouble(double value) => webNotSupported();
+  factory Decimal128.fromDouble(double value) {
```
nirinchev, line 55: Can we take a shortcut here and do a `Decimal.parse(double.toString())`?

packages/realm_dart/lib/src/handles/web/decimal128.dart
```diff
@@ -23,17 +43,94 @@ class Decimal128 implements intf.Decimal128 {
...
+
+  @override
+  Decimal128 abs() => Decimal128._(_value.abs());
+
+  @override
+  int compareTo(covariant Decimal128 other) {
+    final sign = _value.compareTo(other._value);
+    if (sign < 0) return -1;
+    if (sign > 0) return 1;
+    return 0;
```
nirinchev, line 119: Any reason not to return `sign` directly?

packages/realm_dart/lib/src/handles/web/decimal128.dart
```diff
@@ -1,9 +1,29 @@
 // Copyright 2024 MongoDB, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+import 'dart:typed_data';
+
+import 'package:decimal/decimal.dart';
+import 'package:rational/rational.dart';
+
+import 'package:realm_dart/src/handles/native/convert.dart';
```
nirinchev, line 9: Should this depend on something in native?

packages/realm_dart/lib/src/handles/web/decimal128.dart
nielsenko: Yes, I'm trying to make as many as the `decimal128_test.dart` tests pass as possible. The behaviour of `double` will not just return the sign of `a - b`, hence `native.Decimal128` doesn't either, and nor should `web.Decimal128`.

But you could argue that we are already deviating a lot from the semantics of IEEE 754-2008, and the semantics of `double` so perhaps it is a mood point.

packages/realm_dart/lib/src/handles/web/decimal128.dart
nielsenko: No. Fixed.

packages/realm_dart/lib/src/handles/web/decimal128.dart
nielsenko: Yes. Done.

nirinchev (approved):

Issue #1708, opened 2024-06-04
MultipleSyncAgents: Multiple sync agents attempted to join the same session


Navil:
### What happened?

I am seeing crashes for my users related to realm. I do not know how to reproduce it, but is is a crash that happens often enough to push my app above the Android Store threshold. This makes this topic relatively import for us.
<img width="1229" alt="image" src="https://github.com/realm/realm-dart/assets/3436399/93ca1a79-29b9-4085-964d-d53b2f8cfa33">

### Repro steps

-

### Version

Flutter - Channel stable, 3.22.1

### What Atlas Services are you using?

Both Atlas Device Sync and Atlas App Services

### What type of application is this?

Flutter Application

### Client OS and version

 e.g. Redmi gale (Redmi 13C) Android 13 (SDK 33)

### Stacktrace of the exception/crash you're getting

```shell
terminating due to uncaught exception of type realm::MultipleSyncAgents: Multiple sync agents attempted to join the same session

*** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
pid: 0, tid: 16329 >>> at.navil.aupair <<<

backtrace:
  #00  pc 0x000000000008ce34  /apex/com.android.runtime/lib64/bionic/libc.so (abort+164)
  #01  pc 0x0000000000be1364  /data/app/~~vukLyGJKm3ZahZDwjFA-IQ==/at.navil.aupair-2PjF-p25ZVY8Q72AyUDHvg==/split_config.arm64_v8a.apk!librealm_dart.so (BuildId: 3d1db96d07fa0c70b9d537bf8d1503b9fda596d8)
  #02  pc 0x0000000000be0c58  /data/app/~~vukLyGJKm3ZahZDwjFA-IQ==/at.navil.aupair-2PjF-p25ZVY8Q72AyUDHvg==/split_config.arm64_v8a.apk!librealm_dart.so (BuildId: 3d1db96d07fa0c70b9d537bf8d1503b9fda596d8)
  #03  pc 0x0000000000be0b14  /data/app/~~vukLyGJKm3ZahZDwjFA-IQ==/at.navil.aupair-2PjF-p25ZVY8Q72AyUDHvg==/split_config.arm64_v8a.apk!librealm_dart.so (BuildId: 3d1db96d07fa0c70b9d537bf8d1503b9fda596d8)
[...]

PR #1721, opened 2024-06-13
Try to reproduce

nielsenko:


nielsenko:
@nirinchev This is (/was) related to https://github.com/realm/realm-dart/issues/1708, which again relates to https://github.com/realm/realm-core/issues/7190. Perhaps add an update on those issues? @michael-wb worked on this, but the fix never landed before the deprecation.

BTW, there is a lot of missing info on the core issue, as most of the discussions happened on slack, but it is actually very easy to reproduce the issue on any supported platform, using hot-reload.

Issue #1735, opened 2024-06-27
EJSON - fromEJSON does not respect fields that have default value.


Rob-Biemans:
### What happened?

Because MongoDB/Atlas Search does not transfer information about fields that have their default values (nullable). The fromEJSON throws InvalidEJson because the fields are not set with NULL.

### Repro steps

**Create a model with nullable fields like:**

```dart
  class _AddressModel {
    late String? type;
    late String street;
    late String? suburb;
    late String city;
    late String? state;
    late String? postalCode;
    late String countryCode;
    late _LocationModel? location;
  }
```

**Try to convert Map (JSON) to Object**
_Note: type, suburb, state and postalCode have their default value (null) so nothing is returned from Atlas Search_

```dart
AddressModel testWithCrash = AddressModel.fromEJson<AddressModel>(
  {
    countryCode: NLD, 
    street: Malieveld, 
    city: Den Haag, 
    location: {
     coord: [{$numberDouble: 52.085413}, {$numberDouble: 4.317371}],
      type: Point}
  }
)

  static AddressModel _fromEJson(EJsonValue ejson) {
    return switch (ejson) {
      {
        'type': EJsonValue type, <--------- Throws because the json above does not contain that field.
        'street': EJsonValue street,
        'suburb': EJsonValue suburb,
        'city': EJsonValue city,
        'state': EJsonValue state,
        'postalCode': EJsonValue postalCode,
        'countryCode': EJsonValue countryCode,
        'location': EJsonValue location,
      } =>
        AddressModel(
          type: fromEJson(type),
          street: fromEJson(street),
          suburb: fromEJson(suburb),
          city: fromEJson(city),
          state: fromEJson(state),
          postalCode: fromEJson(postalCode),
          countryCode: fromEJson(countryCode),
          location: fromEJson(location),
        ),
      _ => raiseInvalidEJson(ejson),
    };
  }
```

### Version

Flutter 3.19.3 • channel stable - Dart SDK version: 3.3.1 (stable)

### What Atlas Services are you using?

Both Atlas Device Sync and Atlas App Services

### What type of application is this?

Flutter Application

### Client OS and version

realm: ^3.0.0

### Stacktrace of the exception/crash you're getting

```shell
Exception has occurred.
InvalidEJson (Invalid EJson for AddressModel: {countryCode: NLD, street: Malieveld, city: Den Haag, location: {coord: [{$numberDouble: 52.085413}, {$numberDouble: 4.317371}], type: Point}})
```

[...]

PR #1736, opened 2024-06-27
RDART-1062: Allow missing values (implicit null)

nielsenko:
Assuming:
```dart
@RealmModel()
class _Player {
  @PrimaryKey()
  late String name;
  _Game? game;
  final scoresByRound = <int?>[]; // null means player didn't finish
}
```
then the following now works:
```dart
Realm(Configuration.inMemory([Player.schema, Game.schema]));
expect(() => fromEJson<Player>({}), throwsA(isA<InvalidEJson>())); // illegal since mandatory name is missing
final p = fromEJson<Player>({'name': 'Ericsen'}); // legal, since game and scoresByRound are optional
expect(p.toEJson(), {'name': 'Ericsen', 'game': null, 'scoresByRound': <int?>[]});
```

Previously the call to `fromEJson<Player>(ejson)` would fail unless `ejson` was a map with shape `{ 'name': ..., 'game': ..., scoresByRound: ...}` despite `game` being nullable, and `scoresByRound` being optional.

This PR also rectifies and oversight regarding `RealmValue` and `Decimal128`. These types are now supported without explicit calls to `register`.
```dart
  test('RealmValue', () {
    final r = RealmValue.int(42);
    expect(r.toEJson(), {'\$numberInt': '42'});
    expect(fromEJson<RealmValue>({'\$numberInt': '42'}), r);
  });

  test('Decimal128', () {
    final d = Decimal128.fromInt(42);
    expect(toEJson(d), {'\$numberDecimal': '+42E+0'});
    expect(fromEJson<Decimal128>({'\$numberDecimal': '42'}), d);
  });
```

Also, sets are now supported:
```dart
  test('Set on RealmObject', () {
    final realm = Realm(Configuration.inMemory([AllCollections.schema]));
    final o = realm.write(() => realm.add(AllCollections(intSet: {1, 2, 3})));
    expect(o.intSet, {1, 2, 3});
    final deserialized = fromEJsonString<AllCollections>(toEJsonString(o));
    // deserialized is unmanaged, so will never compare equal, but we can test properties
    expect(deserialized.intSet, o.intSet);
  });
```

and `RealValue` can be deserialized from `DBRef`:
```dart
    test('RealmObject', () {
      final p = Player('Christian Eriksen');
      final rv = RealmValue.from(p);
      expect(rv.toEJson(), {'\$id': 'Christian Eriksen', '\$ref': 'Player'});
      expect(fromEJson<DBRef<String>>(rv.toEJson()), isA<DBRef<String>>().having((r) => r.id, '\$id', 'Christian Eriksen'));
      expect((fromEJson<RealmValue>(rv.toEJson()).value as Player).name, p.name);
    });
```

Fixes: #1735 
Fixes: #1737
Fixes: #1761 
Fixes: #1757


packages/CHANGELOG.ejson.md
```diff
@@ -1,3 +1,13 @@
+## 0.4.0
+
+- `fromEJson<T>` now accepts a `defaultValue` argument that is returned if  
+  `null` is passed as `ejson`.
+- `fromEJson<T>` now accepts a `allowCustom` argument that can be used to specify
+  the custom decoders are allowed to be used. Defaults to `null`, which specify that
```
nirinchev, line 6: This sentence reads awkwardly.

packages/ejson/lib/src/decoding.dart
```diff
@@ -32,43 +33,57 @@ const _commonDecoders = {
...
 }();
 
+/// Custom decoders for specific types. Use `register` to add a custom decoder.
+final customDecoders = <Type, Function>{};
+
+final _decoders = CombinedMapView([customDecoders, commonDecoders]);
+
 /// Converts [ejson] to type [T].
 ///
+/// [defaultValue] is returned if set, and [ejson] is `null`.
```
nirinchev, line 64: This doesn't appear to match the implementation - it looks like if the type is nullable, the default value is ignored even if it's set.

packages/ejson/lib/src/types.dart
```diff
@@ -31,6 +31,20 @@ enum EJsonType {
 /// and [MinKey](https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/#mongodb-bsontype-MinKey)
 enum BsonKey { min, max }
 
+final class DBRef<KeyT> {
```
nirinchev, line 34: Missing API docs

packages/realm_dart/test/serialization_test.dart
```diff
@@ -0,0 +1,106 @@
+// Copyright 2022 MongoDB, Inc.
```
nirinchev, line 1:
```suggestion
// Copyright 2024 MongoDB, Inc.
```

nirinchev (commented):
Need to go more thoroughly through the tests, but here's some initial comments

packages/ejson/lib/src/decoding.dart
```diff
@@ -32,43 +33,57 @@ const _commonDecoders = {
...
-  final type = T;
-  final nullable = type.isNullable;
-  final decoder = nullable ? _decodeNullable : _decoders[type.base];
-  if (decoder == null) {
-    throw MissingDecoder._(ejson, type);
-  }
-  final args = nullable ? [type.nonNull] : type.args;
-  if (args.isEmpty) {
-    return decoder(ejson) as T; // minor optimization
+T fromEJson<T>(EJsonValue ejson, {bool? allowCustom, T? defaultValue}) {
```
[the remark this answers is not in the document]
nielsenko: For `_allowCustom` think thread local storage, but since dart don't have shared heaps, we can just use a static variable. If the `allowCustom` parameter is set, it will use that for `_allowCustom` on any recursive calls to `fromEJson` such as would happen for list etc.

The problem I'm trying to solve is that you could `register` a type `T` with all optional values, meaning that `{}` would deserialize to `T` (and any other map for that matter, as redundant properties are already allowed). We don't try to search for best matching deserializer, so in the dynamic case, where the type to deserialize to is unknown (such as is the case for `RealmValue`), the custom deserializers gets in the way.

Hence this flag. I deliberately didn't document it, as it is very internal thing only really meant for `RealmValue`.

However, I will add, and elaborate the in-code comments.

packages/ejson/lib/src/decoding.dart
nielsenko: Yes.. the defaultValue is used for optional, but non-nullable fields. For nullable fields, the default value is always null. I'll elaborate the comment. Check the use in the generated code.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 ## vNext (TBD)
 
 ### Enhancements
-* None
+* Allow nullable fields to be absent in EJson, when deserializing realm objects. (Issue [#1735](https://github.com/realm/realm-dart/issues/1735))
```
nielsenko, line 4: Perhaps better to write:
```suggestion
* Allow nullable and other optional fields to be absent in EJson, when deserializing realm objects. (Issue [#1735](https://github.com/realm/realm-dart/issues/1735))
```

packages/ejson/lib/src/decoding.dart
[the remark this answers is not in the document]
nielsenko: The:
```dart
      {'a': 1, 'b': 2}: {
        // Even though Game could be deserialized from this (everything in Game
        // is optional, hence {} and by extension any map can deserialize to Game),
        // we always decode a map as a map inside a RealmValue.
        'a': {'\$numberInt': '1'},
        'b': {'\$numberInt': '2'},
      },
```
test case illustrate the issue.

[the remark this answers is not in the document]
nielsenko: And the following case in `_decodeAny`:
```dart
    Map<dynamic, dynamic> _ => _tryDecodeCustomIfAllowed(ejson) ?? _decodeDocument<String, dynamic>(ejson), // other maps goes last!!
```
Obviously any map can decode as a map, so we need to check custom decoders first, but in some cases that is not what you want either, hence the flag.

One could perhaps argue that `_decodeAny` should never even try the custom decoders, and always return a map for non BSOM types. Would yor prefer that?

nirinchev (approved):

packages/ejson/lib/src/decoding.dart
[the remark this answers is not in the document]
nielsenko: I have been thinking about it since yesterday. How do you feel about never trying the custom decoders in `_decodeAny`. Personally I think I prefer that for a couple of reasons:

1. Custom decoders will only be invoked when the type to deserialize to is known. Calling `fromEJson<dynamic>(..)`  (or implicit `fromEJson(..)` will never try the _custom_ decoders registered by the user. This removes ay ambiguity.
2. It is slightly more performant, as `_tryDecodeCustomIfAllowed` tries each custom decoder in turn, until the first match.
3. I can get rid of the `allowCustom`/`_allowCustom` shenanigans.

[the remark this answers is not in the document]
nirinchev: Sure, that also sounds viable

packages/ejson/lib/src/decoding.dart
nielsenko: Actually, there is no reason to restrict `defaultValue` to be used when `T` is non-nullable. Will adapt code to match comment.

packages/CHANGELOG.ejson.md
nielsenko: allowCustom argument removed again

Issue #1740, opened 2024-07-04
`RealmException: No such table exists. Error code: 3020.` during client reset


SPodjasek:
### What happened?

We've added three new collections to sync schema and during client reset following exceptions were thrown on internal testing pre-production devices:

Android: ClientResetError message: A fatal error occurred during client reset: 'User-provided callback failed', inner error: 'RealmException: No such table exists. Error code: 3020.'
iOS: ClientResetError message: A fatal error occurred during client reset: 'User-provided callback failed', inner error: 'RealmException: Error getting property count. No such table exists. Error code: 3020.'

Application hangs with 'gray-screen' and no information is displayed to the user, but after restart it starts without any issues.

### Repro steps

I'm still unable to reproduce this issue.
It never occurred during development, only on pre-production devices in internal testing & test flight.

### Version

3.1.0

### What Atlas Services are you using?

Atlas Device Sync

### What type of application is this?

Flutter Application

### Client OS and version

Android 14, iOS 17.5.1

### Stacktrace of the exception/crash you're getting

```shell
[Android]
          Non-fatal Exception: io.flutter.plugins.firebase.crashlytics.FlutterError: ClientResetError message: A fatal error occurred during client reset: 'User-provided callback failed', inner error: 'RealmException: No such table exists. Error code: 3020.'
       at RealmServices._initialize.<fn>(realm_services.dart:425)
       at ._syncErrorHandlerCallback(config_handle.dart:222)
       at ._FfiCallback_syncErrorHandlerCallback(dart:ffi)
       at RealmLibrary._realm_dart_delete_persistent_handle.#ffiClosure146(realm_bindings.dart)
       at RealmLibrary.realm_scheduler_perform_work(realm_bindings.dart:8712)
       at SchedulerHandle.invoke(scheduler_handle.dart:26)
       at Scheduler._handle(scheduler.dart:45)

[iOS]
          Non-fatal Exception: FlutterError
0  ???                            0x0 RealmServices._initialize.<fn> + 425 (realm_services.dart:425)
1  ???                            0x0 (null)._syncErrorHandlerCallback + 222 (config_handle.dart:222)
2  ???                            0x0 (null)._FfiCallback_syncErrorHandlerCallback (dart:ffi)
3  ???                            0x0 RealmLibrary._realm_dart_delete_persistent_handle.#ffiClosure146 (realm_bindings.dart)
[...]

PR #1748, opened 2024-07-15
RDART-1066: Make schema mode for synchronized realms explicit

nirinchev:
Change schema mode to additive explicit. When schema mode is .discovered, orphaned embedded objects are ignored, which causes a "no such table" exception to be called when we try to look up the table metadata.

Related to https://github.com/realm/realm-dart/issues/1740, but unlikely to have fixed it.

### DIFF
--- a/packages/realm_dart/lib/src/handles/native/config_handle.dart
+++ b/packages/realm_dart/lib/src/handles/native/config_handle.dart
@@ -86,7 +86,7 @@ class ConfigHandle extends HandleBase<realm_config> {
       } else if (config is InMemoryConfiguration) {
         realmLib.realm_config_set_in_memory(configHandle.pointer, true);
       } else if (config is FlexibleSyncConfiguration) {
-        realmLib.realm_config_set_schema_mode(configHandle.pointer, realm_schema_mode.RLM_SCHEMA_MODE_ADDITIVE_DISCOVERED);
+        realmLib.realm_config_set_schema_mode(configHandle.pointer, realm_schema_mode.RLM_SCHEMA_MODE_ADDITIVE_EXPLICIT);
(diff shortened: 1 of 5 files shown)


packages/realm_dart/lib/src/handles/native/config_handle.dart
```diff
@@ -86,7 +86,7 @@ class ConfigHandle extends HandleBase<realm_config> {
       } else if (config is InMemoryConfiguration) {
         realmLib.realm_config_set_in_memory(configHandle.pointer, true);
       } else if (config is FlexibleSyncConfiguration) {
-        realmLib.realm_config_set_schema_mode(configHandle.pointer, realm_schema_mode.RLM_SCHEMA_MODE_ADDITIVE_DISCOVERED);
+        realmLib.realm_config_set_schema_mode(configHandle.pointer, realm_schema_mode.RLM_SCHEMA_MODE_ADDITIVE_EXPLICIT);
```
nirinchev, line 89: This is the actual change, the rest is just cleanup of test code that I noticed while running the tests locally.

packages/realm_dart/test/realm_test.dart
```diff
@@ -1953,6 +1957,21 @@ void main() {
     expect(product.id, subscriptionId);
     expect(product.name, 'abc');
   });
+
+  test('Local realm can be opened with orphaned embedded objects', () {
+    final config = Configuration.local([Car.schema, AllTypesEmbedded.schema], path: generateRandomRealmPath());
+    expect(() => getRealm(config), returnsNormally);
+  });
+
+  baasTest('Sync realm with orphaned embedded objects, throws', (appConfig) async {
```
nirinchev, line 1966: These are the tests that test the new changes

PR #1762, opened 2024-07-29
Use qbs v2 and reenable schema migration tests

nirinchev:


packages/realm_dart/lib/src/cli/atlas_apps/baas_client.dart
```diff
@@ -538,6 +546,10 @@ class BaasClient {
           }''');
       }
 
+      print('Enable qbsv2 for ${app.name}');
```
nielsenko, line 549: Do we want this to print?

nielsenko (approved):

packages/realm_dart/lib/src/cli/atlas_apps/baas_client.dart
nirinchev: We do print a bunch of things when setting up the apps, so figured it doesn't hurt to add another print statement to make sure we know we're using qbsv2 (in case we ever want to have tests that run against qbsv1 app).

Issue #1778, opened 2024-08-13
realm_dart-3.4.0/lib/src/handles/native/init.dart:155:12: Error: Method not found: 'DBRef'.


Bilonik:
### What happened?

Cant run Realm 3.4.0

I think is missing the new ejson package.

### Repro steps

1. Download the new Release 3.4.0 and run.

### Version

3.24

### What Atlas Services are you using?

Atlas Device Sync

### What type of application is this?

Flutter Application

### Client OS and version

iOS 17

### Stacktrace of the exception/crash you're getting

```shell
Launching lib/main.dart on iPad Pro 13-inch (M4) in debug mode...
Running Xcode build...                                                  
Xcode build done.                                           17.2s
Failed to build iOS app
Error (Xcode): ../../../.pub-cache/hosted/pub.dev/realm_dart-3.4.0/lib/src/handles/native/init.dart:155:12: Error: Method not found: 'DBRef'.


Could not build the application for the simulator.
Error launching application on iPad Pro 13-inch (M4).
```

Bilonik:
Workaround:

dependency_overrides:
  ejson:
    git:
      url: https://github.com/realm/realm-dart
      ref: main
      path: packages/ejson

mniyaz:
> Workaround:
> 
> dependency_overrides: ejson: git: url: https://github.com/realm/realm-dart ref: main path: packages/ejson

can you help how to do this. I dont understand

SPodjasek:
@mniyaz just put lines shown above to your `pubspec.yaml`:

```yaml
dependency_overrides:
  ejson:
    git:
      url: https://github.com/realm/realm-dart
      ref: main
      path: packages/ejson
```

and run `flutter pub get` afterwards.

nielsenko:
@mniyaz You need to update your `pubspec.yaml` with the suggested _dependecy_override_
```yaml
dependency_overrides:
  ejson:
    git:
      url: https://github.com/realm/realm-dart
      ref: main
      path: packages/ejson
```

Alternatively you can pin `realm` to 3.3.0 until this mess is sorted out.

The reason this is happening is that `realm 3.4.0` was released without `ejson 0.4.0` being released. The `ejson` packages are still pre 1.0, and one of the things still outstanding is that the github action workflow for publishing (`publish-release.yml`) doesn't include them yet.

TimotejKovac:
Just pining realm to 3.3.0 using 

```
dependencies:
  realm: '3.3.0'
```

[...]

PR #1779, opened 2024-08-14
Bump ejson version

nirinchev:
Fixes https://github.com/realm/realm-dart/issues/1778

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,13 @@
 * None
 
 ### Fixed
-* None
+* Fixed ejson dependency to 0.4.0.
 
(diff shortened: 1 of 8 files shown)


nielsenko:
@nirinchev You need to push the ejson packages to pub dev manually as well. 
```sh
melos publish --scope='ejson' --yes --no-dry-run
```
as I recall

nirinchev:
I believe this should happen automatically when we do the release:

https://github.com/realm/realm-dart/blob/08e7ad57b1faa9d2a4260f9fde3287435358095f/.github/workflows/publish-release.yml#L174

Issue #1813, opened 2024-12-19
Incompatible with last stable version of flutter 3.27.1


John-Dormevil:
### What happened?

After upgrading flutter to the last stable version (3.27.1) the flutter application does not compile anymore.

### Repro steps

Upgrade to flutter 3.27.1 stable
try to run your project

### Version

3.27.1

### What Atlas Services are you using?

Local Database only

### What type of application is this?

Flutter Application

### Client OS and version

15

### Relevant log output

```shell
../../../.pub-cache/realm_generator-3.4.1/lib/src/pseudo_type.dart:16:7: Error: The non-abstract class 'PseudoType' is missing implementations for these members:
 - DartType.element3
Try to either
 - provide an implementation,
 - inherit an implementation from a superclass or mixin,
 - mark the class as abstract, or
 - provide a 'noSuchMethod' implementation.

class PseudoType extends TypeImpl {
      ^^^^^^^^^^
../../../.pub-cache/analyzer-6.11.0/lib/dart/element/type.dart:53:17: Context: 'DartType.element3' is defined here.
```

richard457:
None is fixing it from mongo unless someone from the community take it up, anyone?

nirinchev:
This should be an easy fix though - if anyone would like to submit a quick PR addressing the issue, I'd be happy to kick off a release. Otherwise it may need to wait until after the holidays.

haitr:
I have made a PR (https://github.com/realm/realm-dart/pull/1814) to temporarily fix the issue until they accept or modify the generator.

```
realm:
  git:
    url: https://github.com/haitr/realm-dart.git
    path: packages/realm
```

PR #1814, opened 2024-12-24
Fix: Resolve code gen in Flutter version 3.27.1

haitr:
Problem
It fails to start due to incompatibility issues with updated versions of the `analyzer` and `source_span` packages.

Solution
Pin the versions of `analyzer` and `source_span` packages.

Changes
Updated `pubspec.yaml` to fix package versions:
- `analyzer`: pinned to version 6.9.0
- `source_span`: pinned to version 1.8.0

### DIFF
--- a/packages/realm_generator/pubspec.yaml
+++ b/packages/realm_generator/pubspec.yaml
@@ -1,7 +1,7 @@
 name: realm_generator
 description: >-
-             Generates RealmObject classes from Realm data model classes.
-             This package is part of the official Realm Flutter and Realm Dart SDKs.
+  Generates RealmObject classes from Realm data model classes.
+  This package is part of the official Realm Flutter and Realm Dart SDKs.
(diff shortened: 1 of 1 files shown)


haitr:
@cla-bot check

victorcode1:
This needs to be resolved

nirinchev (changes requested):
I think this can only be merged as a short-term fix - pinning the analyzer package like that can cause incompatibilities with other generators that may require different versions.

As far as I understand, the failure reported in https://github.com/realm/realm-dart/issues/1813 is due to `element3` being added to `DartType` - could we instead of pinning the analyzer, provide a `noSuchMethod` implementation that returns `never`? Or does this break in a different way?

haitr:
@nirinchev Agreed. #1815 came up with the better implementation than mine.

PR #1837, opened 2025-03-26
closes #1824: doc implements

### LINKED ISSUE #1824: Cannot include a Realm class into another Realm class
### What happened?

I wonder how to include custom objects on Realm class which I could found nowhere on the doc.

### Repro steps

```dart
@RealmModel()
class _CustomField{
  // The tag if the custom field is a single string or a double string.
  late final String type;
  // The single field.
  late final String? single;
  // The double fields.
  late final String? our;
  late final String? their;
}
```

The codes above runs with no error. Then I tried

```dart
@RealmModel()
class _CustomField{
  // The tag if the custom field is a single string or a double string.
  late final String type;
  // The single field.
  late final String? single;
  // The double fields.
  late final String? our;
  late final String? their;
}

@RealmModel()
class _CustomFields{
  late List<CustomField> customFields;
}
```

Now it complains `List<CustomField> is not a realm model type`. Even without `List` the issue exists.

ChatGPT told me to add a label `@RealmEmbedded` but it doesn't seem to work. Dart language server complains `Undefined name 'RealmEmbedded' used as an annotation. Try defining the name or importing it from another library.` The script complains `Could not resolve annotation for 'class _CustomField'.`

What's more: What would it happen if I try to use an abstract class object as a data field of a Realm object? How to express things like Rust enums or C/C++ unions?

### Version

Dart 3.5.1 Flutter 3.24.1

Local Database only

Flutter Application

### Client OS and version

Windows 11 21H2

### Code snippets

_No response_

```shell

```

### Relevant log output

```shell

```

HeavySnowJakarta:
- Node the user to add `part` although the language server shows errors
- The usage of embedded objects

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ This repository holds the source code for the Realm SDK for Flutterâ„¢ and Dartâ
 
 ## Getting Started
 
-* Import Realm in a dart file `app.dart`
+* Import Realm in a dart file `app.dart`. Remember to add `part 'app<the file name>.realm.dart'` below the imports or Realm won't generate needed objects automatically, although the language server may complain the URI does not exist.
 
     ```dart
     import 'package:realm/realm.dart';  // import realm package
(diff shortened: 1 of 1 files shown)


HeavySnowJakarta:
@cla-bot check

HeavySnowJakarta:
ref: #1824

PR #1847, opened 2025-05-01
Add support for new versions of libs to realm_generator

YaroCabbage:
Add support for new versions of libs: analyzer v7, source_gen v2, dart_style v3


lokalathlet:
It would be fantastic if this change could be integrated into the code. Otherwise, we might find ourself limited to using other actual libraries such as 'freezed'.

nirinchev:
Superseded by https://github.com/realm/realm-dart/pull/1850

